PR #9, opened 2021-08-02
Standard Image Features

matfechner:
Signed-off-by: Mathias Fechner <mathias.fechner@plusserver.com>

### DIFF
--- a/Design-Docs/Image-Properties-Spec.md
+++ b/Design-Docs/Image-Properties-Spec.md
@@ -211,3 +211,25 @@ A boolean property that is not present is considered to be `false`.
   pricing (but it might be provided by a contracted 3rd party, e.g. the OS vendor).
 
 
+##  Standardized Image  
+
+Images which are "default" in Glance of any SCS-Provider and are certified by SCS, should have following Features:
+
+*   Optional: Latest LTS Distribution of:
+ -    Debian
+ -    Ubuntu, 
+ -    AlmaLinux and/or Rocky Linux
+ -    OpenSLES
+
+* Optional: A generic Kernel which is running on every Environment
+
+* Needed: Hardening: an enabled Mandatory Access Control System, either SELinux or Apparmor
+*   Optional: a definite User specification for the default Login User, not Distribution specified.
+* Needed: An only ssh-key enabled Authentication.
+* Needed: a Pipeline for building images with verification key 
+* Needed: a Pipeline pass through (install something, an test function)
+* Needed: a Mirror to hosted SCS-Images
+* Optional: default software tools
+    - tbd
+
+


fkr:
@garloff can you comment?

Design-Docs/Image-Properties-Spec.md
```diff
@@ -211,3 +211,25 @@ A boolean property that is not present is considered to be `false`.
...
 
+##  Standardized Image  
+
+Images which are "default" in Glance of any SCS-Provider and are certified by SCS, should have following Features:
+
+*   Optional: Latest LTS Distribution of:
+ -    Debian
+ -    Ubuntu, 
+ -    AlmaLinux and/or Rocky Linux
+ -    OpenSLES
```
garloff, line 222: This should be openSUSE Leap, right?

Design-Docs/Image-Properties-Spec.md
```diff
@@ -211,3 +211,25 @@ A boolean property that is not present is considered to be `false`.
...
 
 
+##  Standardized Image  
+
+Images which are "default" in Glance of any SCS-Provider and are certified by SCS, should have following Features:
+
+*   Optional: Latest LTS Distribution of:
+ -    Debian
+ -    Ubuntu, 
+ -    AlmaLinux and/or Rocky Linux
```
garloff, line 221: Hmmm, do we know already who will be the CentOS successor?

Design-Docs/Image-Properties-Spec.md
```diff
@@ -211,3 +211,25 @@ A boolean property that is not present is considered to be `false`.
...
+
+* Needed: Hardening: an enabled Mandatory Access Control System, either SELinux or Apparmor
+*   Optional: a definite User specification for the default Login User, not Distribution specified.
+* Needed: An only ssh-key enabled Authentication.
+* Needed: a Pipeline for building images with verification key 
+* Needed: a Pipeline pass through (install something, an test function)
+* Needed: a Mirror to hosted SCS-Images
+* Optional: default software tools
+    - tbd
+
```
garloff, line 234: The thoughts on building optimized images should -- IMVHO -- not be part of a specs file at this point.
Note that the Image-Properties-Spec.md file IS an official SCS standard -- thoughts on how to eveolve it should be tracked in a separate document.
While I think it's appropriate to suggest (or even mandate?) that the provider maintains a set of the most popular free Linux images as part of the spec, we should not discuss the early-stage ideas here in the same doc, IMVHO.

garloff (commented):
Please separate image building thoughts into a separate document.

fkr:
thanks for the feedback @garloff. 
@matfechner can you look at the feedback next feedback? Thanks!

Design-Docs/Image-Properties-Spec.md
```diff
@@ -211,3 +211,25 @@ A boolean property that is not present is considered to be `false`.
   pricing (but it might be provided by a contracted 3rd party, e.g. the OS vendor).
 
 
+##  Standardized Image  
+
+Images which are "default" in Glance of any SCS-Provider and are certified by SCS, should have following Features:
```
frosty-geek, line 216: I don't understand what you're trying to say @matfechner

Design-Docs/Image-Properties-Spec.md
```diff
@@ -211,3 +211,25 @@ A boolean property that is not present is considered to be `false`.
...
+
+*   Optional: Latest LTS Distribution of:
```
frosty-geek, line 218: So it's optional or mandatory to provide the latest LTS version? If it's optional why include this in the "standard"? In my mind there would be a mandatory section aswell if there is an optional section.

what happens if the next LTS version comes out, will the support be dropped for latest-1 version?

Design-Docs/Image-Properties-Spec.md
```diff
@@ -211,3 +211,25 @@ A boolean property that is not present is considered to be `false`.
...
+ -    Debian
+ -    Ubuntu, 
+ -    AlmaLinux and/or Rocky Linux
+ -    OpenSLES
+
```
frosty-geek, line 223: what about all the other linux distro's out there? is there a good way to not have a biased list? https://distrowatch.com/dwres.php?resource=ranking&sort=average100 didn't really help me besides remembering that there is a huge Arch Linux Community out there that I tend to forget...

fkr:
Discussion in Team IaaS on 2022-07-13

The question is: wether we're not in the position (or wether it is not our job) to provide a biased list for SCS environments?
List:

Ubuntu
Debian
Fedora
Rocky/Alma (centos stream)
openSUSE Leap
grml (rescue purposes)

where applicable LTS and regular. nothing with rolling releases.


@matfechner can you update the PR with this please?

Discussion on wether cluster api image should be listed here. The definition of the cluster image should come from the container layer, however every cluster api image needs to be part of a SCS environment.

fkr:
Am pulling this into the spring, since I want to have this finalized before R3. Let's finish this up in tomorrows edition of the SIG Standardization.

garloff:
> Discussion on wether cluster api image should be listed here. The definition of the cluster image should come from the container layer, however every cluster api image needs to be part of a SCS environment.

Our reference implementation for KaaS with Cluster-API does know where to get the node images from and registers them on the fly as needed. The images come from https://minio.services.osism.tech/openstack-k8s-capi-images/ and work for OpenStack CAPI environments (certainly for SCS IaaS). That said, it is better if the provider already has those images registered, saving 8mins or so image registration time and being more efficient for the disk and image storage as well.

I would expect SCS certified KaaS platforms to have node images for all popular k8s versions available.

garloff:
A view on standard images:
* Some providers prefer "optimized" images that have some cloud-specific pre-configuration; in the worst case, provider-specific images may even be required b/c special drivers are needed. (Full disclosure: OTC originally required special images, and continued build and providing omtimized images later, when the requirement was lifted.)
* Some customers appreciate the specific images; many don't, as they want something that is well-known and works across clouds with identical properties.

Considerations:
* Require unchanged upstream images to be available under specific names. (Downside: If these don't work due to specifics in the infrastructure, the provider can not offer them or get granted an exception.)
* The vendor_data mechanism can be used to inject configuration information, software, ... into freshly started VMs by default. (E.g. using local mirrors for package updates to provide a very valid example.)
* Additional, "optimized" images could be added under different names -- if we standardize the names, these would not have a cloud-specific name ... The source and build process would need to be made transparent if we somehow standardize this, though. Both post-processed upstream images and built-from-the-ground-up images could be done, though we might want to differentiate between those by namespace?

fkr:
As discussed (among others with @frosty-geek and @matfechner) this PR will be closed. The document has been updated by @garloff and this PR is now obsolete.

PR #12, opened 2021-09-10
Glossary first draft

MHammPS:
for discussing

### DIFF
--- a/Design-Docs/glossary.md
+++ b/Design-Docs/glossary.md
@@ -17,19 +17,18 @@ This is a draft version, which is free to be discussed
 
 ## Definition of a Region
 
-An OpenStack/SCS Region consists of at least one or more Availability Zones that share a Control Plane with their services. As a result, they share an API, as well as a CEPH cluster, i.e. a data store. Within the region, any Layer 2 networks are available to the user.
+An OpenStack/SCS region consists of at least one or more Availability Zones that share a Control Plane with their services. As a result, they share one API. Also a Control Plane can share one CEPH cluster over different fire departments or each Availbility Zone can have its own CEPH cluster. Within the region, any Layer 2 networks are available to the user. Availbility Zones which build a region are connected by redundant low-latency (<2ms (guess!!) ) high bandwidth (10s of Gbps) connections.
 
 Regions can be federated when the SCS code is ready.
 
-A region is located in one fire compartment.
 
-## Definition of a Avialbility Zone
+## Definition of a Availability Zone
 
 An Availability Zone is a (physical) group of multiple compute nodes, controlled by the region's control plane that provides the API and interface.
 
 An Availability Zone allows OpenStack compute hosts to be divided into logical groups and provides a form of physical isolation and redundancy from other Availability Zones, for example by using a separate power supply or network devices.
 
-When users provision resources, they can specify in which Availability Zone their instance should be created. In this way, cloud customers can ensure that their application resources are distributed across different computers to achieve high availability in the event of a hardware failure.
+When users provision resources, they can specify in which Availability Zone their instances should be created. In this way, customers can ensure that their application resources are distributed across different failure domains to achieve high availability in the event of a hardware failure.
 
 ## Definition of Host Aggregates
 
@@ -37,7 +36,7 @@ Host aggregates are a mechanism for partitioning compute nodes which is not expl
 Administrators assign flavors to host aggregates by specifying metadata on the host aggregate and customizing the extra specifications of the flavor. It is then up to the Nova scheduler to determine the best match for the user request. Compute nodes can also be in more than one host aggregate.
 
 
-Optionally, one can designate a host aggregate as an Availability Zone, e.g. for simplificatin reasons of the user selection of an availbility zone.  
+Optionally, one can designate a host aggregate as an Availability Zone, e.g. for simplification reasons of the user selection of an availbility zone.  
 Availability Zones differ from Host Aggregates in that they are shown to the user as a Nova boot option, so Compute VMs can be started on them.
 Compute Nodes, however, can only be in a single Availability Zone. We can configure a default Availability Zone where instances will be scheduled if the user does not specify an Availability Zone.
 
@@ -57,7 +56,7 @@ The services in the Cell Controllers can still call placement APIs, but cannot a
 
 ## Definition of a Control Plane
 
-In Openstack/SCS, a Control Plane consists of 5 hardware nodes, which together serve several Availability Zones and thus provide a common usable API for a region. The Control Plane also shares the network (Neutron), the Sheduler and the CEPH services.
+In Openstack/SCS, a Control Plane consists of at least 5 hardware nodes, which together serve several Availability Zones and thus provide a common usable API for a region. The Control Plane also shares the network (Neutron), the Scheduler and the CEPH services.
 
 It includes the Controller Nodes (Galera Cluster, RabbitMQ) and the Manager Nodes, Maas,...
 
@@ -89,7 +88,7 @@ The Rest API provides the core of openstack/SCS  and can be addressed for a whol
 
 ## Horizon
 
-Horizon is openstack's preferred GUI for the end user, but also for the administrator for a quick overview. It runs on the controller node. 
+Horizon is openstack's preferred GUI for the end user, but also for the administrator for a quick overview. It runs on the controller node. Other GUIs are possible, also GUIs which replace the horizon interface
 
 ## Message Queue
 
@@ -97,7 +96,8 @@ Most OpenStack services communicate with each other through the message queue. F
 
 ## Keystone
 
-The OpenStack Identity module called Keystone is used as an authentication and rights system between the OpenStack components. Keystone divides access to projects in the cloud into so-called "clients" ("tenant"). A tenant is a tenant of the cloud and has at least one associated user. It is possible to create multiple users per tenant with different rights. Keystone uses a token system for authorization and also supports connection to other authentication options such as LDAP. Furthermore, it is possible to abstract the authentication to the web server via an upstream web server and thus, for example, directly use one of the numerous existing authentication modules of the Apache web server for authorization.
+( The OpenStack Identity module called Keystone is used as an authentication and rights system between the OpenStack components. Keystone divides access to projects in the cloud into so-called "tenants". A tenant is a tenant of the cloud and has at least one assigned user. It is possible to create multiple users per tenant with different rights. Keystone uses a token system for authorization and also supports the connection to other authentication options such as LDAP. (wikipedia) )
+
 
 ## Glance
 
--- a/Design-Docs/glossary.md
+++ b/Design-Docs/glossary.md
@@ -1,0 +1,143 @@
+---
+title: SCS Definitions (Draft)
+version: 2021-09-09-001
+authors: Matthias Hamm
+state: v1.0 (for R0)
+---
+
+Glossary - Naming conventions for SCS (Draft) 
+
+# Motivation
+
+In order to have a consistent naming scheme in SCS environments, we need a definition of the different terms. 
+
+# Proposal
+
+This is a draft version, which is free to be discussed
+
+## Definition of a Region
+
+An OpenStack/SCS Region consists of at least one or more Availability Zones that share a Control Plane with their services. As a result, they share an API, as well as a CEPH cluster, i.e. a data store. Within the region, any Layer 2 networks are available to the user.
+
+Regions can be federated when the SCS code is ready.
+
+A region is located in one fire compartment.
+
+## Definition of a Avialbility Zone
+
+An Availability Zone is a (physical) group of multiple compute nodes, controlled by the region's control plane that provides the API and interface.
+
+An Availability Zone allows OpenStack compute hosts to be divided into logical groups and provides a form of physical isolation and redundancy from other Availability Zones, for example by using a separate power supply or network devices.
+
+When users provision resources, they can specify in which Availability Zone their instance should be created. In this way, cloud customers can ensure that their application resources are distributed across different computers to achieve high availability in the event of a hardware failure.
+
+## Definition of Host Aggregates
+
+Host aggregates are a mechanism for partitioning compute nodes which is not explicitly visible to users in an OpenStack/SCS cloud.  Host aggregates are based on arbitrary characteristics such as server type, processor type, GPU, disk speed, etc.
+Administrators assign flavors to host aggregates by specifying metadata on the host aggregate and customizing the extra specifications of the flavor. It is then up to the Nova scheduler to determine the best match for the user request. Compute nodes can also be in more than one host aggregate.
+
+
+Optionally, one can designate a host aggregate as an Availability Zone, e.g. for simplificatin reasons of the user selection of an availbility zone.  
+Availability Zones differ from Host Aggregates in that they are shown to the user as a Nova boot option, so Compute VMs can be started on them.
+Compute Nodes, however, can only be in a single Availability Zone. We can configure a default Availability Zone where instances will be scheduled if the user does not specify an Availability Zone.
+
+Info: A prerequisite for creating an Availability Zone is a host aggregate.
+
+## Definition of a Cell
+
+The Cells paradigm simplifies the handling of large Openstack deployments. 
+
+Cells is an OpenStack Nova feature that improves scalability for Nova in OpenStack Platform. Each Cell has a separate database and message queue, which increases performance when scaling. One can provision additional Cells to handle large deployments, and compared to Regions, this allows access to a large number of compute nodes through a single API.
+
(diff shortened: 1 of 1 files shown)


Design-Docs/glossary.md
```diff
@@ -0,0 +1,143 @@
...
+
+When users provision resources, they can specify in which Availability Zone their instance should be created. In this way, cloud customers can ensure that their application resources are distributed across different computers to achieve high availability in the event of a hardware failure.
+
+## Definition of Host Aggregates
+
+Host aggregates are a mechanism for partitioning compute nodes which is not explicitly visible to users in an OpenStack/SCS cloud.  Host aggregates are based on arbitrary characteristics such as server type, processor type, GPU, disk speed, etc.
+Administrators assign flavors to host aggregates by specifying metadata on the host aggregate and customizing the extra specifications of the flavor. It is then up to the Nova scheduler to determine the best match for the user request. Compute nodes can also be in more than one host aggregate.
+
+
+Optionally, one can designate a host aggregate as an Availability Zone, e.g. for simplificatin reasons of the user selection of an availbility zone.
```
horazont, line 40: "simplificatin"?

Design-Docs/glossary.md
```diff
@@ -0,0 +1,143 @@
...
+
+Each Cell has its own Cell controllers running the database server and RabbitMQ along with the Nova Conductor services.
+
+Nova Conductor services, called "Super Conductor", continue to run on the main controller nodes.
+
+The services in the Cell Controllers can still call placement APIs, but cannot access other API layer services via RPC, nor can they access global API databases on the control nodes.
+
+## Definition of a Control Plane
+
+In Openstack/SCS, a Control Plane consists of 5 hardware nodes, which together serve several Availability Zones and thus provide a common usable API for a region. The Control Plane also shares the network (Neutron), the Sheduler and the CEPH services.
```
horazont, line 60: s/Sheduler/Scheduler/

If the number of nodes is fixed, then their roles should be made clear.

Design-Docs/glossary.md
```diff
@@ -0,0 +1,143 @@
...
+
+This is a draft version, which is free to be discussed
+
+## Definition of a Region
+
+An OpenStack/SCS Region consists of at least one or more Availability Zones that share a Control Plane with their services. As a result, they share an API, as well as a CEPH cluster, i.e. a data store. Within the region, any Layer 2 networks are available to the user.
+
+Regions can be federated when the SCS code is ready.
+
+A region is located in one fire compartment.
```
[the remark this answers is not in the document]
garloff: I agree with Jonas and disagree with the proposal here.

For me, an "availability zone" is a failure domain. So you need to have a reasonable failure isolation between AZs of the same region, e.g. different rooms. My expectation of an AZ is that a broken switch, a fire, a power outage, a broken network uplink will very likely only take down one AZ (or none :-) ). There is some flexibility -- whether you have redundant internet connectivity shared between AZs or separate ones per AZ is up to you.

My definition would be:
A region consists of one or several availability zones, which are geographically located close to each other and are connected by low-latency (<2ms) high bandwidth (10s of Gbps) connections.

Design-Docs/glossary.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Definition of a Avialbility Zone
+
+An Availability Zone is a (physical) group of multiple compute nodes, controlled by the region's control plane that provides the API and interface.
+
+An Availability Zone allows OpenStack compute hosts to be divided into logical groups and provides a form of physical isolation and redundancy from other Availability Zones, for example by using a separate power supply or network devices.
+
+When users provision resources, they can specify in which Availability Zone their instance should be created. In this way, cloud customers can ensure that their application resources are distributed across different computers to achieve high availability in the event of a hardware failure.
```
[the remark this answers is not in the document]
garloff: Agree.

Design-Docs/glossary.md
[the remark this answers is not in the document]
garloff: Another question:
Don't we typically share identities across regions of one cloud provider? In OpenStack terms: the keystone is (logically) shared. (Whether we replicate the identities or have a central model with decentral caches is irrelevant.)
Federation is for me the next level, where I can authorize identities from several sources -- this does not seem to be required for a single provider multi region setup.
What do our IAM and OpenStack experts think here?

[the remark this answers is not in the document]
horazont: All you said about identity management is correct as far as I know. Regions still trust one another regarding authentication and authorization, which may not be the case for federated deployments.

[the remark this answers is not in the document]
horazont: >  A region consists of one or several availability zones, which are geographically located close to each other and are connected by low-latency (<2ms) high bandwidth (10s of Gbps) connections.

+unmetered

Also I guess we can bikeshed whether 2ms is really a good threshold. I am looking at a customer complaint at the moment where they have the issue that a latency of 1.5ms between database and app is noticeably slow in their setup.

Design-Docs/glossary.md
```diff
@@ -0,0 +1,143 @@
...
+In order to have a consistent naming scheme in SCS environments, we need a definition of the different terms. 
+
+# Proposal
+
+This is a draft version, which is free to be discussed
+
+## Definition of a Region
+
+An OpenStack/SCS Region consists of at least one or more Availability Zones that share a Control Plane with their services. As a result, they share an API, as well as a CEPH cluster, i.e. a data store. Within the region, any Layer 2 networks are available to the user.
+
```
evrardjp, line 21: Does it make sense to share one or more ceph clusters? What for?

In this concept, l2 network is stretched in the region, does that make sense?

Design-Docs/glossary.md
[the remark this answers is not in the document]
evrardjp: Low latency <2ms isn't necessarily possible in 2 DCs of the same region in case of fiber cut. You probably want redundant paths that aren't always optimal.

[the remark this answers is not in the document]
evrardjp: I agree on the definition of AZ = failure domain, and I think that SCS community members should clarify what their failure domain means for them. It's a differentiator between vendors.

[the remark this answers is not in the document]
evrardjp: To answer to @garloff , not all vendors implement federation or shared keystone. Some vendors prefer a shared nothing architecture... 

However, I believe in this case it would be wise to think about federation from day 1 for regions (even if it means k2k). AZs should not be a concern for keystone, as it should be a single keystone per region (with multiple load balanced endpoints). 
I am not personally fond of k2k but federation from day1 makes sense to me.

Design-Docs/glossary.md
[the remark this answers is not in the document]
evrardjp: Completely agree.

Design-Docs/glossary.md
```diff
@@ -0,0 +1,143 @@
...
+A region is located in one fire compartment.
+
+## Definition of a Avialbility Zone
+
+An Availability Zone is a (physical) group of multiple compute nodes, controlled by the region's control plane that provides the API and interface.
+
+An Availability Zone allows OpenStack compute hosts to be divided into logical groups and provides a form of physical isolation and redundancy from other Availability Zones, for example by using a separate power supply or network devices.
+
+When users provision resources, they can specify in which Availability Zone their instance should be created. In this way, cloud customers can ensure that their application resources are distributed across different computers to achieve high availability in the event of a hardware failure.
+
```
evrardjp, line 33: This doesn't map to my definition of AZ for openstack. You are focusing on compute, while AZ is a notion that nova, cinder, and neutron have.
Next to that, that definition of AZ brings questions regarding glance.

Design-Docs/glossary.md
evrardjp: Why 5 hardware nodes? In SCS the reference implementation should mention 5 hardware node, but the reference architecture shouldn't be limited to 5.

Design-Docs/glossary.md
```diff
@@ -0,0 +1,143 @@
...
+Each Cell has its own Cell controllers running the database server and RabbitMQ along with the Nova Conductor services.
+
+Nova Conductor services, called "Super Conductor", continue to run on the main controller nodes.
+
+The services in the Cell Controllers can still call placement APIs, but cannot access other API layer services via RPC, nor can they access global API databases on the control nodes.
+
+## Definition of a Control Plane
+
+In Openstack/SCS, a Control Plane consists of 5 hardware nodes, which together serve several Availability Zones and thus provide a common usable API for a region. The Control Plane also shares the network (Neutron), the Sheduler and the CEPH services.
+
```
evrardjp, line 61: That's quite optimistic to have all of that in 5 nodes btw.

Design-Docs/glossary.md
```diff
@@ -0,0 +1,143 @@
...
+The provider network is the network that is "in front", i.e. at the output points of the openstack/SCS. This is usually a public network, but can also be a private network in individual cases. IPs from the provider network can be assigned to instances within the SCS. The same applies to load balancers, of course. 
+
+## Definition of API
+
+The Rest API provides the core of openstack/SCS  and can be addressed for a whole region. It accepts and responds to end-user API calls.  The service supports the OpenStack Compute API, the Amazon EC2 API, and a special Admin API for privileged users to perform administrative actions. Policies are enforced and most orchestration actions can be started, such as launching an instance.
+
+## Horizon
+
+Horizon is openstack's preferred GUI for the end user, but also for the administrator for a quick overview. It runs on the controller node. 
+
```
evrardjp, line 93: Please define "preferred". I am not sure it's the case for all the vendors. (See also new projects around dashboards)

Design-Docs/glossary.md
```diff
@@ -0,0 +1,143 @@
...
+## Message Queue
+
+Most OpenStack services communicate with each other through the message queue. For example, Compute communicates with Block Storage services and Network services via the message queue. RabbitMQ, Qpid, and Zeromq are popular choices for a message queue service. When the message queue fails or becomes inaccessible, the cluster generally comes to a halt and ends up in a read-only state where the information is stuck at the point where the last message was sent. Therefore, this is clustered.  RabbitMQ has shown itself to be the most widespread and best supported variant in the OpenStack context, Qpid occurs occasionally, ZeroMQ lacks HA functionality to date
+
```
evrardjp, line 97: Is 0MQ still supported? Can't remember if that removed from oslo_messaging.

Design-Docs/glossary.md
```diff
@@ -0,0 +1,143 @@
...
+## Keystone
+
+The OpenStack Identity module called Keystone is used as an authentication and rights system between the OpenStack components. Keystone divides access to projects in the cloud into so-called "clients" ("tenant"). A tenant is a tenant of the cloud and has at least one associated user. It is possible to create multiple users per tenant with different rights. Keystone uses a token system for authorization and also supports connection to other authentication options such as LDAP. Furthermore, it is possible to abstract the authentication to the web server via an upstream web server and thus, for example, directly use one of the numerous existing authentication modules of the Apache web server for authorization.
+
```
evrardjp, line 101: We should probably not have the word tenant in here.

evrardjp: I think it might be good to link with the official keystone doc, which (IIRC) explains the external and other mechanisms for auth for keystone.

evrardjp: Do we really need to re-define those upstream project's definitions?
Shouldn't those be referring to upstream documentations instead?

evrardjp (commented):
Quick comments, sorry for my brievety.

Design-Docs/glossary.md
noonedeadpunk: I think OMQ uses amqp, which is supported?
What is misssing here is kafka, which is pretty popular option for messaging at least outside of the openstack.

Design-Docs/glossary.md
noonedeadpunk: Especially with skyline being announced and it's in active development now https://opendev.org/skyline/skyline-apiserver

Design-Docs/glossary.md
```diff
@@ -0,0 +1,143 @@
...
+In the OpenStack releases since Ussuri, the "OpenVirtualNetwork"(OVN) has established itself, it replaces many of the Neutron components, e.g. L3 and DHCP agent, so that Neutron only has to talk directly to OVN.
+
+## Cinder
+
+OpenStack Block Storage or Cinder provides virtual block storage in the form of virtualized storage media (hard disks, CDs, etc.). The block storage can be attached to virtual machines. An API interface allows Cinder to connect to Swift so that block storage media can communicate with object storage. Meanwhile, many other storage backends are also fully or partially supported. There is also the option of defining multiple backends and creating a volume type for each backend, so that when a new volume is created, it can be selected on which storage backend the volume is created.
+
+## Swift
+
+Swift is the so-called object storage that can be used by Nova. This is responsible for redundant data storage. Swift can also be used as a backend for Cinder or Glance. Objects are stored in so-called containers, which are primarily used to group objects and store metadata and in turn belong to individual accounts. Objects and containers are accessed via a REST API.
```
noonedeadpunk, line 136: I'm not sure Swift can be used as a backend to Cinder. It can for cinder-volume only?

Design-Docs/glossary.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Ceph OSD
+
+A Ceph OSD (Object Storage Daemon) logically represents a storage device in a Ceph cluster, which can logically be a hard disk, which is the ideal case. In other cases it can also be a raid, which however leads to considerable performance limitations due to caching or other raid optimization.
```
noonedeadpunk, line 140: I'd just say that it can be any block device, because ideal case would be nvme, but not hard disk:)
Also when we're talking about raid it would be important to underline that we mean disk behind raid card, which works in JBOD mode (or raid-0 from 1 drive for raid controllers that doesn't support JBOD), because having real raid for redundancy let's say, has no sense.

Design-Docs/glossary.md
```diff
@@ -0,0 +1,143 @@
...
+
+For the management of the data link layer, Neutron offers the possibility to use various already existing networking software such as Open vSwitch or the bridge functionality of the Linux kernel by means of plugins.
+
+In the OpenStack releases since Ussuri, the "OpenVirtualNetwork"(OVN) has established itself, it replaces many of the Neutron components, e.g. L3 and DHCP agent, so that Neutron only has to talk directly to OVN.
+
+## Cinder
+
+OpenStack Block Storage or Cinder provides virtual block storage in the form of virtualized storage media (hard disks, CDs, etc.). The block storage can be attached to virtual machines. An API interface allows Cinder to connect to Swift so that block storage media can communicate with object storage. Meanwhile, many other storage backends are also fully or partially supported. There is also the option of defining multiple backends and creating a volume type for each backend, so that when a new volume is created, it can be selected on which storage backend the volume is created.
```
noonedeadpunk, line 132: I'm not really sure about what API interface and what kind of connetion between Cinder and SWIFT we're talking here. I might miss smth ofc, but wasn't able neither to google not to find anything like that lurking among drivers code. Except we're talking about cinder-backup.

Also maybe worth mentioning, that it also supports encryption of specific volume types with LUKS, which can be done thanks to integration with Barbican. Volume encryption might be important bit to meet GDPR: https://www.thegdpr.net/2018/10/20/how-openstack-barbican-deployment-options-secure-your-cloud/

Design-Docs/glossary.md
```diff
@@ -0,0 +1,143 @@
...
+Optionally, the Controller Node runs parts of the Block Storage, Object Storage, Orchestration and Telemetry services.
+
+The Controller Node requires at least two network interfaces.
+
+## Definition of Compute Node
+
+A compute host runs the hypervisor part of compute that runs instances. By default, compute uses the KVM hypervisor. The compute host also runs a networking service agent that connects instances to virtual networks and provides firewall services to the instances through security groups.
+
+If you offer hyper-converged infrastructure, a compute host also serves the Ceph. This makes the storage dynamically scalable (horizontally and vertically). For the Ceph services, 1 CPU core and 4 GB of RAM are reserved per OSD to ensure appropriate performance. 
+
```
noonedeadpunk, line 77: Do we miss definition of Network Node?

Design-Docs/glossary.md
MHammPS: changed to "of at least 5 hardware nodes"

berendt:
- [ ] Migration to GLOSSARY.md

fkr:
@tibeer or @ra-rau will overtake

Design-Docs/glossary.md
tibeer: Mind game: all 5 CPs are in the same AZ, right? What happens to other AZ's if the the AZ with the CP burns down?

Design-Docs/glossary.md
tibeer: Horizon is still the "reference" implementation and actively developed by OpenStack. Skyline is the new approach which is currently still not the "preferred" method.

fkr:
superseeded by #82

PR #27, opened 2021-11-18
Add License marker - Docs are CC-BY-SA 4.0

fkr:


garloff:
Thanks for spotting!

PR #44, opened 2022-02-16
Disscussion about OVN Feature Matrix

matfechner:
Signed-off-by: Mathias Fechner <mathias.fechner@plusserver.com>

@osfrickler
@fkr 

Please take a  look

### DIFF
--- a/Operational-Docs/ovs-ovn-support-matrix.MD
+++ b/Operational-Docs/ovs-ovn-support-matrix.MD
@@ -5,26 +5,50 @@
 
 #### 1. OVN is since OpenStack Ussuri Release the preferential default in OpenStack Clouds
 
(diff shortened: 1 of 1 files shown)


fkr:
Discussed with @osfrickler. The matrix should have more details, be more verbose on certain topics (IPv6 was an example).
Best to be discussed between @matfechner and @osfrickler next week.

fkr:
Imho this can be merged and be worked on through further PRs if need be.

berendt:
We are working with the Github Reviews feature on our end. I think we should do that here as well. Then you can see directly in the dashboard where you should look at something.

But it looks like neither Mathias nor Jens are in the SCS Org. At least I can't assign them.

fkr (approved):
LGTM!

PR #47, opened 2022-02-18
Feat/cluster gitops

garloff:


garloff:
Link to the doc:
https://github.com/SovereignCloudStack/Docs/blob/feat/cluster-gitops/Design-Docs/Cluster-GitOps.md

Design-Docs/Cluster-GitOps.md
```diff
@@ -0,0 +1,78 @@
...
+modify and extend the settings and should allow an opt-in mechanism to
+do more detailed adjustments via overriding also the cluster-templates
+(kustomization style again).
+
+Cluster-admin credentials need to be provided to the owners of the cluster.
+The current thinking is to have a public key included in the git repo;
+the cluster-admin credentials for the created cluster would be encrypted
+with this public key and can then be published. (Would still use https
+to ensure that these are the true credentials.) Only the owner(s) of
+the private key can decrypt the credentials.
```
horazont, line 42: https://github.com/bitnami-labs/sealed-secrets may be worth looking at for this

Design-Docs/Cluster-GitOps.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Implementation thoughts
+
+The reconciliation loop would roughly look like this:
+1. Get the latest clusters from git (via a regular check or an event)
```
horazont, line 47: This should be done by something like fluxcd, shouldn't it?

Design-Docs/Cluster-GitOps.md
```diff
@@ -0,0 +1,78 @@
...
+1. Per cluster
```
horazont, line 48: I'm not sure what (and why) exactly is needed here beyond what already happens when one applies a cluster resource.

In my mind, the workflow would look like: "install fluxcd, register git repository, put cluster manifests in there, done", so this looks very complex. Am I missing anything?

joshmue:
I think this topic also is tightly related to what the general user experience should look like when provisioning cloud resources (be it VM instances, storage buckets or Kubernetes clusters).

From a user perspective, IMHO a single, consistent API for all resources would be best, like most big cloud providers have already.

Design-Docs/Cluster-GitOps.md
```diff
@@ -0,0 +1,82 @@
...
+   1. For all other optional services (nginx, flux, cert-manager, harbor, ...):
+      1. Deploy service if enabled, otherwise remove (if deployed before)
+   1. Sanity checks
+   1. optional CI tests
+
+## Open questions
+
+* Can this be integrated into capo or do we really need to create a loop around it?
+
+* On errors in this loop, we would move to the next cluster. However, how do we handle error reporting?
```
genofire, line 75: on fluxcd: it happen independent on every cluster on the same time (if you have not set an branch, tag (semver parsing included), or gitrepo for each cluster) ...

fluxcd has metrics, we use prometheus to get them and fire alerts, if something does not work.

mbuechse:

fkr:
> Is this still relevant? I will close this PR if nothing happens by July 31st.

i think, it is time to close it. Will do now.

PR #64, opened 2022-04-28
adjust meeting times

fkr:

### DIFF
--- a/Contributor-Docs/collaboration.md
+++ b/Contributor-Docs/collaboration.md
@@ -11,18 +11,18 @@ for more details.
 
 ### Project updates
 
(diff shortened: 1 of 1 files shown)


fkr:
@itrich if we manage to have a yaml that is used to feed the calendar it would be perfect (imho) - since then everyone can do PRs against the yaml and a login to our NC is not required in order to suggest a change to the times ;)

berendt:
Have a look at https://opendev.org/opendev/irc-meetings. We manage the OpenStack meetings since years this way and it works pretty good.

yaml2ical is the tool to generate the ical files based on the yaml files: https://opendev.org/opendev/yaml2ical

fkr:
@berendt thanks for the pointer!

PR #67, opened 2022-04-29
:book: Adding a proposal for standardized cluster infrastructure

batistein:
**Digital Sovereignty through standardized cluster stacks, provider-independent infrastructure, and clearly separated layers**

Signed-off-by: Sven Batista Steinbach [sven.batista-steinbach@syself.com](sven.batista-steinbach@syself.com)
Signed-off-by: Janis Kemper [janis.kemper@syself.com](janis.kemper@syself.com)

**Link:** https://github.com/syself/scs-docs/blob/syself/proposal/Design-Docs/standardizing-cluster-infrastructure.md

### DIFF
--- a/Design-Docs/standardizing-cluster-infrastructure.md
+++ b/Design-Docs/standardizing-cluster-infrastructure.md
@@ -1,0 +1,277 @@
+---
+title: Digital Sovereignty through standardized cluster stacks, provider-independent infrastructure, and clearly separated layers 
+version: 2022-04-29-001
+authors: Sven Batista Steinbach, Janis Kemper
+state: Draft
+---
+
+# Digital Sovereignty through standardized cluster stacks, provider-independent infrastructure, and clearly separated layers
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-goals](#non-goals)
+- [Proposal](#proposal)
+  - [Proposal 1: Pre-defined cluster stacks are necessary to ensure that clusters and applications work - a free configuration of cluster components is not possible](#proposal-1-pre-defined-cluster-stacks-are-necessary-to-ensure-that-clusters-and-applications-work---a-free-configuration-of-cluster-components-is-not-possible)
+  - [Proposal 2: Cluster operators should be able to define clusters independently of the provider](#proposal-2-cluster-operators-should-be-able-to-define-clusters-independently-of-the-provider)
+  - [Proposal 3: Users should be able to define applications independently of the cluster infrastructure](#proposal-3-users-should-be-able-to-define-applications-independently-of-the-cluster-infrastructure)
+- [User stories](#user-stories)
+  - [User story 1: DevOps expert configuring applications on SCS infrastructure](#user-story-1-devops-expert-configuring-applications-on-scs-infrastructure)
+  - [User story 2: Cloud provider implementing SCS stack](#user-story-2-cloud-provider-implementing-scs-stack)
+  - [User story 3: User wanting to implement a multi-cloud](#user-story-3-user-wanting-to-implement-a-multi-cloud)
+  - [User story 4: SCS operator wanting to have maintainable SCS infrastructure](#user-story-4-scs-operator-wanting-to-have-maintainable-scs-infrastructure)
+- [Risks and Mitigations](#risks-and-mitigations)
+- [Design details](#design-details)
+  - [Proposal 1: Using cluster stacks](#proposal-1-using-cluster-stacks)
+    - [What exactly is a cluster stack?](#what-exactly-is-a-cluster-stack)
+    - [What components does a cluster stack have? What has to be considered when defining a stack?](#what-components-does-a-cluster-stack-have-what-has-to-be-considered-when-defining-a-stack)
+      - [Layer 1: Pre-cluster state](#layer-1-pre-cluster-state)
+      - [Layer 2: Cluster state](#layer-2-cluster-state)
+      - [Layer 3: Kubernetes core](#layer-3-kubernetes-core)
+    - [Example of a cluster stack](#example-of-a-cluster-stack)
+  - [Proposal 2: Cluster definition is independent of the provider](#proposal-2-cluster-definition-is-independent-of-the-provider)
+  - [Proposal 3: Applications are configured independently of the cluster infrastructure](#proposal-3-applications-are-configured-independently-of-the-cluster-infrastructure)
+  - [Drawbacks](#drawbacks)
+  - [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+We want to define cluster infrastructure, which works on different cloud providers. The cluster infrastructure should have as much flexibility as possible but has to be well-defined and tested to ensure a good user experience while defining and creating clusters as well as launching applications.
+
+Infrastructure should be portable from one provider to the other and providers should be easily certifiable. The provider infrastructure should be transparent for everybody.
+
+Every cloud provider has to have a Cluster API provider integration and a cloud controller manager (CCM). It can be both an “actual” cloud provider with API and a bare-metal system (based on OpenStack).
+
+To facilitate the configuration of applications in these clusters, we propose to standardize CCM (load balancer) and naming of storage classes, so that applications can be configured independently of the cluster infrastructure.  
+
+## Motivation
+
+The main goal of SCS is to ensure the users’ and cluster operators’ sovereignty while using European cloud infrastructure. With this proposal, we want to ensure that the infrastructure, from the node image to core Kubernetes components, is transparent in all SCS providers. We want to have a concise structure and good documentation on what providers have to do to get certified by SCS and aim to keep the certification process itself lean.
+
+Furthermore, we consider the freedom of choosing a provider a key to the sovereignty of cluster operators. They should operate their clusters fully independently of the provider infrastructure.
+
+For users who launch and run applications on SCS-certified infrastructure, we want to ensure that they do not have to worry about security issues. SCS infrastructure should only get certified when all aspects of security are taken care of. The users should be able to configure their applications independently of the cluster infrastructure. Only this will lead to actual sovereignty, as they would otherwise have to worry about certain security issues, e.g. related to metrics-server with self-signed certificates.
+
+### Goals
+
+- Standardize infrastructure (e.g. definition of load balancers) of European cloud providers
+- Easy certification process for cloud providers
+- Concise documentation on how to fulfill SCS certification
+- Applications are configured independently of the cloud provider
+- Automated e2e tests for cluster and application configuration
+- Secure cluster infrastructure for users and cluster operators
+- Auditability & automated testing (unit and e2e tests) of SCS code
+
+### Non-goals
+
+- This proposal does not intend to define specifically all inputs, e.g. to define load balancers in SCS infrastructure. This is left for an open discussion in the SCS team and with cloud providers.
+- We don’t want to say that any specific components are must-haves for cluster infrastructure, but we will give suggestions.
+
+## Proposal
+
+This proposal does not specify any details about load balancer definitions, node images, or server types. It rather aims to define a way to handle standardization and certification in general.
+
+It consists of three parts:
+
+### Proposal 1: Pre-defined cluster stacks are necessary to ensure that clusters and applications work - a free configuration of cluster components is not possible
+
+A cluster stack consists of “big components”, e.g. node image and Cluster API provider; down to “small” ones, such as private/public network, IPv4/IPv6, container runtime (interface), and firewall. All of these components related to the image, networking, or cluster resources interact and there is no way of freely configuring such a cluster without heavy constraints.
+
+Why do we think that such cluster stacks are necessary?  
+
+- Some configurations are incompatible (e.g. choosing eBPF without a node image that supports it)
+- Some configurations might work but require further knowledge and additional work (Ubuntu and Cilium work together in general, but not with Cilium’s latest version because of a bug in older kernels)
+- Some configurations depend on the cloud providers. A user has to know these details to get a working system (MTU size, e.g. with Cilium).
+
+Assuming that a cluster consists of n components, then we would say that decisions need to be taken for all of these n components (i.e. Kubeadm as bootstrap controller or Ubuntu as OS for node image). The collection of these decisions defines a cluster stack.
+
+There can be multiple cluster stacks to satisfy different needs, but there should not be too many cluster stacks. This is necessary to keep the cluster stacks maintainable (defining, testing, updating) and to not overwhelm and confuse users with a large number of stacks that cannot be clearly differentiated.
+
+### Proposal 2: Cluster operators should be able to define clusters independently of the provider
+
+The cluster infrastructure should be independent of the providers because of a few reasons. It ensures true freedom of choice (a big aspect of digital sovereignty), allows easy migrations of infrastructure from one provider to another, and keeps the configuration of clusters easy for cluster operators.
+
+To achieve this, the Cluster API provider integrations have to be standardized in some way. Cluster operators should only configure objects that are independent of the provider. For example, instead of having to configure the `ProviderMachine` object of two different providers, the cluster operator should only configure an SCS machine object, which is then translated into the object of the relevant provider. This is not meant to change the provider integrations themselves, but rather to have an additional translation layer.
+
+### Proposal 3: Users should be able to define applications independently of the cluster infrastructure
+
+SCS should have a repository with application configurations. The configurations in this repository should not depend on the provider, as this would create a lot of additional maintenance work. Additionally, it would not be possible to migrate applications easily from one provider to another.
+
+To achieve independence of application configurations from the cluster infrastructure, there are mainly two tasks that have to be done: standardizing the CCM, which currently requires provider-specific configurations of service objects, and standardizing the naming of storage classes.  
+
+## User stories
+
+The persona we call “user” (using SCS clusters) is only supposed to work with applications and should not worry at all about the cluster infrastructure below. Cluster operators define and manage clusters based on pre-defined cluster stacks that are implemented by infrastructure providers.
+
+### User story 1: DevOps expert configuring applications on SCS infrastructure
+
+A DevOps expert wants to add a Traefik ingress to the SCS application stack. As the ingress needs a load balancer, he defines a service of type load balancer. In the SCS documentation, there is a section describing how to implement a load balancer and what kind of parameters can be set, such that the DevOps expert does not need a long time to finish the configuration. This configuration is now valid for all cloud providers and can be used for every cluster on different European cloud providers. The maintainability of SCS applications is ensured by this independence of the application and cluster layer.
+
+### User story 2: Cloud provider implementing SCS stack
+
+After the cloud provider implemented the SCS cluster stack, all applications are ready to be installed and work perfectly. Therefore, the cloud provider does not have to invest time in configuring applications but can rely on the pre-defined and well-maintained SCS stack. Its users can profit from the additional value generated by this feature of ready-to-install applications. E2E tests ensure that everything has been implemented correctly.
+
+### User story 3: User wanting to implement a multi-cloud
+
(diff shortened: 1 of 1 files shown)


itrich:
Wow, this is really awesome! 🤯
I'll take myself some time to go through your proposal, but this looks really great at first sight. Thank you so much and have a great weekend!

Design-Docs/standardizing-cluster-infrastructure.md
```diff
@@ -0,0 +1,277 @@
...
+- Some configurations depend on the cloud providers. A user has to know these details to get a working system (MTU size, e.g. with Cilium).
+
+Assuming that a cluster consists of n components, then we would say that decisions need to be taken for all of these n components (i.e. Kubeadm as bootstrap controller or Ubuntu as OS for node image). The collection of these decisions defines a cluster stack.
+
+There can be multiple cluster stacks to satisfy different needs, but there should not be too many cluster stacks. This is necessary to keep the cluster stacks maintainable (defining, testing, updating) and to not overwhelm and confuse users with a large number of stacks that cannot be clearly differentiated.
+
+### Proposal 2: Cluster operators should be able to define clusters independently of the provider
+
+The cluster infrastructure should be independent of the providers because of a few reasons. It ensures true freedom of choice (a big aspect of digital sovereignty), allows easy migrations of infrastructure from one provider to another, and keeps the configuration of clusters easy for cluster operators.
```
horazont, line 94: "allows easy migrations of infrastructure from one provider to another" is probably not true. At most, it simplifies migrations. But as soon as you want to reduce downtimes of your services, no migration between two clusters is ever "easy".

Design-Docs/standardizing-cluster-infrastructure.md
```diff
@@ -0,0 +1,277 @@
...
+- This proposal does not intend to define specifically all inputs, e.g. to define load balancers in SCS infrastructure. This is left for an open discussion in the SCS team and with cloud providers.
+- We don’t want to say that any specific components are must-haves for cluster infrastructure, but we will give suggestions.
+
+## Proposal
+
+This proposal does not specify any details about load balancer definitions, node images, or server types. It rather aims to define a way to handle standardization and certification in general.
+
+It consists of three parts:
+
+### Proposal 1: Pre-defined cluster stacks are necessary to ensure that clusters and applications work - a free configuration of cluster components is not possible
```
horazont, line 78: `s/Proposal/Part/`, I initially read these as mutually exclusive, but I don't think now that's how it was intended?

Design-Docs/standardizing-cluster-infrastructure.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Proposal 3: Users should be able to define applications independently of the cluster infrastructure
+
+SCS should have a repository with application configurations. The configurations in this repository should not depend on the provider, as this would create a lot of additional maintenance work. Additionally, it would not be possible to migrate applications easily from one provider to another.
+
+To achieve independence of application configurations from the cluster infrastructure, there are mainly two tasks that have to be done: standardizing the CCM, which currently requires provider-specific configurations of service objects, and standardizing the naming of storage classes.
```
horazont, line 102: What exactly is meant by "standardizing the CCM"?

Design-Docs/standardizing-cluster-infrastructure.md
```diff
@@ -0,0 +1,277 @@
...
+
+A DevOps expert wants to add a Traefik ingress to the SCS application stack. As the ingress needs a load balancer, he defines a service of type load balancer. In the SCS documentation, there is a section describing how to implement a load balancer and what kind of parameters can be set, such that the DevOps expert does not need a long time to finish the configuration. This configuration is now valid for all cloud providers and can be used for every cluster on different European cloud providers. The maintainability of SCS applications is ensured by this independence of the application and cluster layer.
+
+### User story 2: Cloud provider implementing SCS stack
+
+After the cloud provider implemented the SCS cluster stack, all applications are ready to be installed and work perfectly. Therefore, the cloud provider does not have to invest time in configuring applications but can rely on the pre-defined and well-maintained SCS stack. Its users can profit from the additional value generated by this feature of ready-to-install applications. E2E tests ensure that everything has been implemented correctly.
+
+### User story 3: User wanting to implement a multi-cloud
```
horazont, line 116: What is "a multi-cloud"?

Design-Docs/standardizing-cluster-infrastructure.md
```diff
@@ -0,0 +1,277 @@
...
+
+There are various components playing a role in defining cluster infrastructure. We consider the following three technical layers:
+
+1. Pre-cluster state - node image
+2. Cluster state - Cluster API and provider implementation
+3. Kubernetes core applications
+
+All of these layers depend on each other, such that it is not possible to freely select and exchange components. This is why we introduce the cluster stack which consists of a selection of sub-systems and configurations that follow a certain goal.
+
+For example, there can be a stack for Kubernetes version 1.22 and one for 1.23. There can also be stacks that use a node image based on Ubuntu and others with Fedora. These cluster stacks have to be implemented by the providers, as there are certain details, e.g. related to the node image, that have to be adapted to the respective provider.
```
horazont, line 144: Talking about the node image: Do we pin down the distribution which needs to be used?

I'm asking because RHEL/CentOS has some funny defaults which [*do* leak down into the container abstraction](https://gitlab.com/yaook/operator/-/issues/121#note_543076539) and hunting all of those down and pulling them straight sounds like a pretty daunting task. (And leading to fun hacks like [`realtime-hack.yaml`](https://gitlab.com/yaook/operator/-/blob/devel/ci/devel_integration_tests/deploy/realtime-hack.yaml#L20))

(FWIW, you could also read it the other way: Ubuntu has defaults which allow things which are (and possibly should not be) not allowed on RHEL and that affects portability.)

Also, SELinux and AppArmor also leak into k8s, keyword `securityContext`.

Design-Docs/standardizing-cluster-infrastructure.md
```diff
@@ -0,0 +1,277 @@
...
+
+#### What exactly is a cluster stack?
+
+First, SCS defines a stack and decides that it should use a certain Kubernetes version, that it has a certain node image, etc. Based on these decisions, the stack is defined in yaml files using some templating language (e.g. Helm/Helmfile or Jsonnet). These yaml files can be seen as a framework that is implemented by the providers. Some parts of the cluster stack configuration have to be adapted by the cloud providers. Good documentation from SCS ensures that implementing a cluster stack is straightforward. A templating language is used, so that cluster operators can specify certain configurations of the cluster stacks that have not been fixed. For example, the configuration of OpenID Connect.
+
+#### What components does a cluster stack have? What has to be considered when defining a stack?
+
+Based on the aforementioned layers, we want to make a list of all technical details, cluster components, and things to consider.
+
+##### Layer 1: Pre-cluster state
```
ajfriesen, line 156: Some of these things in the node look like highly specific implementation detail to me.

I think we should define rules on how to decide if something should be part of the node.
For example, if you are able to add a piece of software via helm or Kubernetes YAML files only there should be no need to implement this at the node image level.

There are also details like `cgroupfs` vs `systemd` which IIRC should not have any impact to the cluster operator.
At gridscale, we switched to `systemd` because we have better stability in high memory pressure situations with this config. However, this could potentially be different on other systems.

In general, we should define a goal here but not define the specific piece of software.

Design-Docs/standardizing-cluster-infrastructure.md
```diff
@@ -0,0 +1,277 @@
...
+
+##### Layer 2: Cluster state
+
+Cluster state consists of standardized Cluster API objects, such as `ClusterClass`, and other life cycle objects, e.g. `MachineDeployment` and `MachineHealthCheck`. Additionally, there are provider-specific objects that exist in every provider integration, like `ProviderCluster` and `ProviderMachineTemplate`, as well as other objects.
+
+SCS has to define the inputs for those objects in each Cluster Stack in such a way, that each provider can implement it.
+
+The most important standardization, but also the one that restricts the options to define clusters the most, is related to how the Kubernetes system is supposed to look like. Note that the Kubernetes system is heavily influenced by the components chosen for the pre-cluster state.
+
+The first decision that has to be taken is to select a bootstrap provider: Kubeadm or Talos. Talos would require a node image with Talos OS.
```
horazont, line 185: What about completely custom bootstrapping? What is the usecase for offering a switch between kubeadm and talos?

Design-Docs/standardizing-cluster-infrastructure.md
```diff
@@ -0,0 +1,277 @@
...
+
+A specific Kubernetes version has to be chosen as well, as other components depend on it.
+
+**What do we consider part of every cluster stack?**
+
+We recommend choosing these components for every stack to make sure that the cluster works well, is secure and that the maintainability of the cluster stacks is better. As all of the settings that we propose differ a lot between Talos and Kubeadm, we choose the latter one, as it is more common.
+
+In Kubeadm there are some settings that we would consider must-haves:
+
+- cloud-init
```
ajfriesen, line 197: This is potentially an implementation detail as well.

What do we want to achieve with cloud-init here?
I guess it is some bootstrap process at boot.

This could potentially also be achieved without cloud-init.
Allowing other solutions would allow providers to participate even if they do not have cloud-init.

horazont, line 197: What is cloud-init in the kubeadm context? I only know of [this cloudinit](https://cloudinit.readthedocs.io/), but that seems unrelated in this stage?

Design-Docs/standardizing-cluster-infrastructure.md
```diff
@@ -0,0 +1,277 @@
...
+
+Example:
+
+- OpenStack load balancer: [https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/openstack/loadbalancer.go#L67](https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/openstack/loadbalancer.go#L67)
+- Hetzner load balancer: [https://github.com/syself/hetzner-cloud-controller-manager/blob/master/internal/annotation/load_balancer.go](https://github.com/syself/hetzner-cloud-controller-manager/blob/master/internal/annotation/load_balancer.go)
+
+This can be avoided rather easily. We have to implement a new interface for load balancers, as a counterpart to the current one ([https://pkg.go.dev/k8s.io/cloud-provider#LoadBalancer](https://pkg.go.dev/k8s.io/cloud-provider#LoadBalancer)), which is a part of the cloud provider interface ([https://pkg.go.dev/k8s.io/cloud-provider#Interface](https://pkg.go.dev/k8s.io/cloud-provider#Interface)). SCS implements the current cloud provider interface and defines a new one, which SCS cloud providers have to implement. In there, we unify the naming of the annotations and give a pre-defined set of inputs to the function that, for example, creates a load balancer.
```
horazont, line 250: Wouldn't it be much easier to have a mutating admission hook which generates the corresponding annotations from a "standard" annotation?

THat way, we don't need intermediate objects or controllers or whatnot.

Design-Docs/standardizing-cluster-infrastructure.md
```diff
@@ -0,0 +1,277 @@
...
+
+Note that standardizing the inputs has the effect that cloud providers cannot have certain individual features, which require some unique inputs.
+
+The second task is to standardize the naming of storage classes.
+
+Different cloud providers have different storage classes and name them differently. As these names of the storage classes are used in persistent volume claims (i.e. parts of the configuration of our applications) ([https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reserving-a-persistentvolume](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reserving-a-persistentvolume)), we would need a separate configuration for each application and cloud provider. Therefore, there wouldn’t be any standardization. Instead, we propose to standardize the naming of storage classes, so that the PVCs in the configuration of applications can be defined independently of the provider.
+
+### Drawbacks
+
+Standardizing something always leads to less freedom of choice. Cloud providers would not be able to have individual inputs for creating load balancers, so some of their features might be obsolete.
```
horazont, line 265: This can be mitigated by using only a mutating hook instead of a custom object; then additional annotations could simply be passed through.

Design-Docs/standardizing-cluster-infrastructure.md
```diff
@@ -0,0 +1,277 @@
...
+We want to leave out the details of how this translation from SCS-defined properties into the respective provider integrations works. It can be discussed in future meetings.
+
+### Proposal 3: Applications are configured independently of the cluster infrastructure
+
+The first step of achieving independence of the layers, the cluster stacks have to be aligned in a certain way, e.g. that a CSR controller is a must-have for running metrics-server securely. This has been described in Proposal 1.
+
+The second step consists of two tasks:
+
+- Defining an adapted SCS cloud provider interface that unifies the inputs for creating load balancers
+- Standardizing the naming of storage classes
```
ajfriesen, line 241: Storage classes might have the same name but could have a big difference in performance.

Should a default storage class not be enough and the rest is provider-specific?

At gridscale, we currently have 3 performance classes for storage.
A fourth one will be coming.
Others could have just one or 2 it 4.

Also, providers could offer something that others don't, like local NVME storage.

horazont: That's really similar to the challenges of the [OpenStack flavor naming](https://github.com/SovereignCloudStack/Operational-Docs/blob/main/flavor-naming-draft.MD) and OpenStack volume type naming discussions (not sure if the latter has officially started yet).

Design-Docs/standardizing-cluster-infrastructure.md
horazont: The choice of cgroup handling is certainly visible, at least in some use-cases: https://gitlab.com/yaook/operator/-/issues/345

Might be that this specific one should be out of scope…

Design-Docs/standardizing-cluster-infrastructure.md
berendt: I think a distinction should be made here between the basic interchangeability of infrastructures and a migration. I think it's not necessarily the first goal to achieve migration between different infrastructures.

Design-Docs/standardizing-cluster-infrastructure.md
berendt: The same here. Why limit ourself to certain tools?

Design-Docs/standardizing-cluster-infrastructure.md
berendt: Good point. I think it would be good if we could define the SCS globally so that we have the same understanding of it everywhere.

Design-Docs/standardizing-cluster-infrastructure.md
horazont: The point I was trying to make is that attempting to specify what a distribution must deliver is going to be impossible to do completely. I never knew about that realtime scheduling parameter until the different default on CentOS broke stuff.

So if we care about portability, this is a challenge we must address in some way.

ajfriesen: There are also good reasons to derive from popular distributions (centos, ubuntu) to something build for light container runtimes like flatcar, coreos, and I bet there are many more.

For me, defining a distribution is again an implementation detail that should not be defined.

Design-Docs/standardizing-cluster-infrastructure.md
ajfriesen: We are using kubeadm without cloud-init since we miss it at this stage at gridscale.

I would be fine with missing something and therefore not be able to certify but this again looks like an implementation detail.

ajfriesen:
My general feeling is that many things are specific implementation details that we should not care about (at least in my opinion).

A more realistic example from my past:

When a fire department wants new trucks they need to do a public tender.
This tender is written in a way that only describes the functionality.
You can not say that the new truck needs to be from Mercedes.
Because Mercedes is an implementation detail. (Also this is illegal and distortion of competition)

My point is:
We should not define implementation detail. There are many ways to achieve the same goal.
It is totally possible to create a cluster without kubeadm, you can write your own tool if you desire. It does not really matter as long as you get a compliant k8s cluster.

Design-Docs/standardizing-cluster-infrastructure.md
janiskemper: @horazont @berendt we certainly meant inter-changeability, which is a necessary part of migrating clusters from one provider to another easily. However, there are many more things that play a role. Therefore, we mean "inter-changeability" with migrations in the text.

Design-Docs/standardizing-cluster-infrastructure.md
janiskemper: @horazont no, they are not exclusive. They are, more or less, independent and all necessary to achieve the independence of layers (such as applications independent of infrastructure provider)

Design-Docs/standardizing-cluster-infrastructure.md
janiskemper: We mean a multi-provider cloud by that.

Design-Docs/standardizing-cluster-infrastructure.md
janiskemper: @horazont I completely agree with you. This is why we would say we need to have a framework/guidelines on our side that follow best practices and include the knowledge of all participating companies. With the combined knowledge we can develop node images that are production-ready and follow all best practices. And have a good solution that provider-specific details can be implemented by everyone.

Design-Docs/standardizing-cluster-infrastructure.md
janiskemper: We just thought about those two because they are the two most popular ones. If you know others, then please add them!

Design-Docs/standardizing-cluster-infrastructure.md
janiskemper: @horazont the bootstrap provider kubeadm can generate output files for cloud-init and ignition. That's the relation between the two!

Design-Docs/standardizing-cluster-infrastructure.md
janiskemper: The proposed approach only requires implementing one function (CreateLoadBalancer) and creates a clear interface that cloud providers can implement. If you have an easier and better-defined approach, I'm of course open to it!

Without spending a lot of time thinking about the mutation hook approach, it sounds quite hacky and not transparent to me.

Design-Docs/standardizing-cluster-infrastructure.md
janiskemper: There are certainly different approaches here - @ajfriesen you pointed out one. We should probably also spend a close look at related discussions, as you say @horazont!

janiskemper:
@ajfriesen I fully agree with you and I hope that we followed your suggestions in many cases (and made it clear that it is only an opinion, as we wrote down an example cluster stack). 
I only think that we have a different understanding of the impact of certain decisions. The choice of Kubeadm or an equivalent does affect other (application) layers. 
To have more flexibility, we propose using (a few) different cluster stacks that take different choices of components. If one provider does not want to implement one cluster stack, they could go for the three others. That's at least the system that we had in mind!

fkr:
Discussed in Team Container Call on July 11th. This PR does not seem to reflect the latest results of the discussion.
@batistein can you rework this PR?

fkr:
Converted to Draft as long as this is WIP

Design-Docs/standardizing-cluster-infrastructure.md
garloff: cloud-init is used to bootstrap nodes IN OUR REFERENCE IMPLEMENTATION.
Anyone that wants to reuse some of our ref. implementation is well advised to ensure cloud-init works.
That said, it is not part of the user-facing container standard, as Andrej correctly observed.

@berendt: For compatibility at IaaS layer, we require cloud-init support, yes, but that's a different layer and thus a different discussion.

jschoone:
Hi @batistein and @janiskemper 
is this still WIP or should we close and restart?

janiskemper:
I guess we should close this and restart. We can still keep this as a reference, but I guess that we have now better possibilities to approach this than this proposal, thanks to VP05!

jschoone:
> I guess we should close this and restart. We can still keep this as a reference, but I guess that we have now better possibilities to approach this than this proposal, thanks to VP05!

Thanks for the reply, let's do it this way.

PR #70, opened 2022-05-30
Add glossary file

berendt:
Signed-off-by: Christian Berendt <berendt@osism.tech>

### DIFF
--- a/GLOSSARY.md
+++ b/GLOSSARY.md
@@ -1,0 +1,5 @@
+# SCS Glossary
+
+This file serves as the central glossary within SCS. It is intended to clearly
+define terms used within SCS where there may be differing understandings. The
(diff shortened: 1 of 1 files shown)


fkr:
As reminded by @itrich  in PB (where this PR came from): we should see about merging #12.

berendt:
We can merge this one and then rebase the other PR and adjust it a bit in terms of format? I think the glossary should not be included in the design docs but should be available independently for everything.

itrich:
> We can merge this one and then rebase the other PR and adjust it a bit in terms of format? I think the glossary should not be included in the design docs but should be available independently for everything.

Yes, let's proceed as suggested.

itrich (approved):

PR #81, opened 2022-07-25
Clarify reference implementation, testbed and where development happens

fkr:
This came up a while ago when we discussed where certain code lives.


itrich (approved):
These clarifications really help to understand the scope of SCS (standards vs. reference implementation). Thank you very much!

PR #102, opened 2022-08-31
Create status page comparison.md

master-caster:
Signed-off-by: Benjamin <2255421+master-caster@users.noreply.github.com>


master-caster (commented):
Initial commit to compare some status page apps

fkr:
very cool - thanks for your work @master-caster!
I would suggest that we don't place it in Contributor-Docs, since that folder is for documents regarding contributions.

Instead I suggest a new folder 'Decisions' in which we place a folder 'Status-Page'. There we can place your comparions (please no spaces in the filename).  Based on your document, we can then proceed with the decision making and document that in an ADR that references the comparison.

fkr (changes requested):
See comment about placing the file.

Issue #109, opened 2022-09-08
tools/flavor-name-check.py: -ib without prior CPU spec gets misinterpreted


garloff:
```
garloff@TuxKurt(wave-scs):/casa/src/SCS/Docs/Design-Docs [0]$ ./tools/flavor-name-check.py -v SCS-4C:16:50-ib
Flavor: SCS-4C:16:50-ib
 CPU:RAM: #vCPUs: 4, CPU type: Dedicated Core, ?Insec SMT: False, ##GiB RAM: 16.0, ?no ECC: False, ?RAM Over: False
 Disk: #:NrDisks: 1, #.GB Disk: 50, .Disk type: 
 No Hypervisor
 No NestedVirtualization
 CPUBrand: .CPU Vendor: Intel, #.CPU Gen: Pre-Skylake, Performance: Std Perf
 No GPU
 No Infiniband

ERROR: Could not parse: b
Traceback (most recent call last):
  File "/casa/src/SovereignCloudStack/Docs/Design-Docs/./tools/flavor-name-check.py", line 527, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/casa/src/SovereignCloudStack/Docs/Design-Docs/./tools/flavor-name-check.py", line 512, in main
    ret = parsename(name)
  File "/casa/src/SovereignCloudStack/Docs/Design-Docs/./tools/flavor-name-check.py", line 457, in parsename
    raise NameError("Error 60: Could not parse %s (extras?)" % n)
NameError: Error 60: Could not parse b (extras?)
```
`SCS-4C:16:50-ib` is a legal name, but our parser is not clever enough for it. It interprets the `-i` of `-ib` as indicator for an intel CPU and expects a number, or nothing and/or `h` to follow. Will need to special case this or improve the parser logic ...

PR #110, opened 2022-09-08
Fix/cpugen vs ib2

garloff:
This fixes issue #109:
To do so, allow for regexps to include an end anchor '$'
    and try to match against a string where trailing -XXXX gets cut off.


fkr (approved):
LGTM!

garloff:
Thanks, Felix!

Issue #96, opened 2021-06-18
Network interface speed


horazont:
See SovereignCloudStack/Monitoring#3 for my definition of Prio 1-4.

- Prio 3: As a Cloud Operator, I want to know when a network interface drops below the specified maximum speed on that link, as that indicates bad cabeling, a bad driver, or another (hardware) issue and may cause customer impact (reduced performance).
- Prio 2: As a Cloud Operator, I want to know when the network interfaces of a compute node are saturated, as that causes degration of performance.

PR #113, opened 2022-09-08
Feat/add basic docs for IAM federation (#96) (#160)

reqa:
* Design-Docs/IAM-federation/keystone-keycloak-federation.md plus two sketches (#96)

* ../Decisions/intra-SCS-federation-setup-description-for-osism-doc-operations.md as proposal for (#160)

Signed-off-by: Arvid Requate <requate@univention.de>


fkr (approved):
As discussed: while this holds items that should find their way into user stories, I'm merging the PR and will then see that the proper user stories are cut from this.

PR #116, opened 2022-09-15
We have a standard, add standard images.

garloff:
Standard images are not yet mandatory, but we share our intentions already according to the discussion from 2022-08-25:
https://github.com/SovereignCloudStack/minutes/blob/main/sig-standardization/20220825.md

### DIFF
--- a/Design-Docs/Image-Properties-Spec.md
+++ b/Design-Docs/Image-Properties-Spec.md
@@ -133,6 +133,12 @@ etc., up to 3 days later is not considered a violation of this policy. So the a 
 from an image with `monthly` update frequency might be 2021-04-14, 2021-05-14, 2021-06-15,
 2021-07-14, 2021-07-27 (hotfix), 2021-08-13 ...
 
+Promises to update the registered public images tpyically depend on upstream image providers
+(Linux distributors, OS vendors) keeping their promises to build and provide updated images.
+Failures from upstream are not a reason to claim the cloud provider to be in violation of his
+promises. However, if the provider observes massive upstream failures (which can e.g. cause
+increased security risks), we advise the provider to inform the users.
+
 We recommend updating images at least monthly.
 
 The `hotfix_hours` field indicates how providers deal with critical security issues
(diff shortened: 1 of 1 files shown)


fkr (commented):
I'd like to suggest that instead of hardcoding the versions, we use terms such as LATEST and STABLE and PREVIOUS STABLE and add a paragraph that gives examples based on the versions.

garloff:
You may want to read the rendered text here for reviewing:
https://github.com/SovereignCloudStack/Docs/blob/feat/image-standard-images/Design-Docs/Image-Properties-Spec.md

garloff:
> I'd like to suggest that instead of hardcoding the versions, we use terms such as LATEST and STABLE and PREVIOUS STABLE and add a paragraph that gives examples based on the versions.

Good point, so maybe I put `<LATEST>` there and put an explanation and an example at the time of writing in there.
The advantage is that it forces us to put more effort into explaining how this evolves as new versions come out.
And of course the advantage that we don't need to touch the spec just b/c new versions get released.
Disadvantage is that it needs a human reader to interpret the language with the dangers of misinterpretation ...

matfechner (approved):
LGTM

garloff:
Hi Felix, I have implemented your suggestions!
If you have further thoughts, please let us know.

frosty-geek:
I disagree with "image_description" in the Image Origin section as well as the section following it as these would only apply to images not provided by upstream (Ubuntu prebuilt cloud-images) IMHO but the specs don't say so explicitly. 
I would suggest that we add a note that this applies only to non-upstream images.

We should also add if upstream does not produce a refreshed image that there will be no updates even though the policy states it will be updated monthly.

garloff:
Hi Ralf,

thanks for your feedback, really appreciated!

> I disagree with "image_description" in the Image Origin section as well as the section following it as these would only apply to images not provided by upstream (Ubuntu prebuilt cloud-images) IMHO but the specs don't say so explicitly. I would suggest that we add a note that this applies only to non-upstream images.

For upstream images, we typically have a page where these can be downloaded from from the upstream vendor.
So I put almost the same URL in `image_source` and `image_description`. The `image_source` could actually be a direct URL to the (.qcow2) file and the `image_description` the page where it's listed and described by the upstream distributor. This may not be perfect, but we wanted to avoid the risk of having images where maybe a downloadable file is referenced but no description whatsoever , that's why we made this mandatory.
Would a sentence of clarifying language make you more agreeable or do you want us to chose a different approach?

> We should also add if upstream does not produce a refreshed image that there will be no updates even though the policy states it will be updated monthly.

When we make promises we typically depend on upstream keeping its promises as well.
So we need to research upstream policies and make own promises that are consistent with these. (If Ubuntu maintains 18.04 LTS until April 2023, we can not put a provided_util 2023-08-30 there (unless we have special arrangements with Canonical). But yes, if upstream breaks their promises, so will we.
We actually have language in the `hotfix_hours` that describes this; do we need to add an exception clause for the regular updates as well?

garloff:
Improvements included, as discussed.

garloff:
I take the thumbs-up as approving reviews and merge ...

PR #118, opened 2022-09-15
Feat/image metadata checker

garloff:
We have a tool now that checks compliance of the images with the image metadata specification.

### DIFF
--- a/Design-Docs/Image-Properties-Spec.md
+++ b/Design-Docs/Image-Properties-Spec.md
@@ -210,4 +210,9 @@ A boolean property that is not present is considered to be `false`.
   service provider does provide support for this image included in the image/flavor
   pricing (but it might be provided by a contracted 3rd party, e.g. the OS vendor).
 
+## Conformance test tool
 
+The `tools/` subdirectory has a testing tool `image-md-check.py` that retrieves the
+image list from a configured cloud and checks them for completeness (with respect
+to the intended future list of mandatory images) and checks each image for the
+completeness and consistency of mandatory properties.
--- a/Design-Docs/tools/image-md-check.py
+++ b/Design-Docs/tools/image-md-check.py
@@ -8,10 +8,11 @@
 # (c) Kurt Garloff <kurt@garloff.de>, 09/2022
 # SPDX-License-Identifier: CC-BY-SA-4.0
 
-import os, sys, getopt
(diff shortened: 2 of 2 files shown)


matfechner (approved):
LGTM

fkr:
For a starter this is OK, since we then have something, which is better than nothing.
For the metadata, I'd really like to not have it hardcoded in the script itself, that should live in a separate json (or similar structure), so that it can be consumed elsewhere too.

fkr (approved):

garloff:
> For a starter this is OK, since we then have something, which is better than nothing. For the metadata, I'd really like to not have it hardcoded in the script itself, that should live in a separate json (or similar structure), so that it can be consumed elsewhere too.

I once had a quick look at cuelang, but I was unable to invest enough time to become efficient in using it.
Maybe there are also better ways to design DSLs that describe the spec, I'm not an expert and definitely hope someone with more expertise can take over.
In some way, I have a DSL to describe the requirements:
```
# From the image metadata specification
# https://github.com/SovereignCloudStack/Docs/blob/main/Design-Docs/Image-Properties-Spec.md
os_props = (property("os_distro", True, ()),
            property("os_version", True, ()))
arch_props = (property("architecture", True, ("x86_64", "aarch64", "risc-v"), "CPU architecture"),
              property("hypervisor_type", True, ("kvm", "xen", "hyper-v", "esxi", None)))
hw_props = (property("hw_rng_model", False, ("virtio", None), "Random Number Generator"),
            property("hw_disk_bus", True, ("virtio", "scsi", None), "Disk bus: vda or sda"))
build_props = (property("image_build_date", True, ()),
               property("patchlevel", False, (),),
               property("image_original_user", True, (), "Default ssh username"),
               property("image_source", True, ()),
               property("image_description", True, ()))
maint_props = (property("update_frequency", True, ("yearly", "quarterly", "monthly", "weekly", "daily", "critical_bug", "never")),
               property("provided_until", True, ()),
               property("uuid_validity", True, ()),
               property("hotfix_hours", False, ()))
```
This could be expressed in a yaml/json doc, of course.
When you read the code and comments in `validate_imageMD()`, you can of course see that the DSL is not expressive enough yet and would also need a bit of documentation (e.g. what exactly does a value None in the list of allowed values mean?).

PR #119, opened 2022-09-16
Rename hw/nestvirt support from -vmx to -hwv.

garloff:
vmx is actually an intel specific name; it's called svm on AMD. (Good catch from Mathias, thanks!)

So chose a name that is neutral -- not only for x86-64, but even for arm and others: hwv is short for hardware virtualization.

### DIFF
--- a/Design-Docs/flavor-naming.md
+++ b/Design-Docs/flavor-naming.md
@@ -50,7 +50,7 @@ We believe the following characteristics are important in a flavour description:
 
 | Prefix | CPU | Suffix     | RAM[GiB] | optional: Disk[GB] | optional: Disk type | optional: extra features                           |
 |--------|-----|------------|----------|--------------------|---------------------|----------------------------------------------------|
-| SCS-   |  N  | L/V/T/C[i] | :N[u][o] | [:[Mx]N]           | [n/s/l/p]           | [-hyp][-vmx][-[arch[N][h][-[G/g]X[N][:M[h]]][-ib] |
+| SCS-   |  N  | L/V/T/C[i] | :N[u][o] | [:[Mx]N]           | [n/s/l/p]           | [-hyp][-hwv][-[arch[N][h][-[G/g]X[N][:M[h]]][-ib] |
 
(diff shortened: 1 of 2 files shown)


horazont (commented):
I'm missing a "design considerations" entry which explains the choice of `hwv` including the naming choice.

garloff:
> I'm missing a "design considerations" entry which explains the choice of `hwv`.

I would not want to include too much justification in a spec document.

Are you fine with a good squash-commit message that explains that:
- vmx was a poor choice b/c it's intel specific when most users would use vmx (intel) and svm (AMD) interchangably
- we decided against short forms for nested virtualization, as we wanted to expose a capability that allows for virtualization with reasonable performance through hardware support that can be delivered both by bare-metal instances and by nested virtualization.
- we in general try to keep the flags short to keep flavor name length managable
- we refrained from -hv to avoid confusion with hyper-v (-hyv) hypervisor designation

I have started a doc that justifies the existence of our flavor naming spec in PR #121.
We can collect more detailed considerations behind the names/abbreviations chosen there.

Let me know if that works for you!

garloff:
I want to finalize this - preferably with some feedback of yours, @horazont !

matfechner:
@garloff , @horazont  From my perspective we talk about the Linux hardware accelerated support with Kernel-based Virtual Machine, short  kvm 
it is the overlay name for this technology,  this include the both Hardware types Intel VT and AMD-V. https://www.linux-kvm.org

garloff:
@matfechner - thanks for the feedback!
Are you suggesting a different name for the flag? -kvm rather than -hwv?
This would not work -- we have an option to specify the hypervisor with -xen, -kvm, -hyv, -vmw already which would conflict.
Also the duality of vmx/svm is being used by xen just as by kvm, so I would not call out the technology that consumes the capability, but the capability itself (hardware acceleration for virtualization).
So I would go ahead and merge this with some reasoning in the commit log -- and there is always the discussion in github that could be researched (in case we get famous and someone does archeology on us :-) )

Issue #73, opened 2022-06-21
Flavor specs: Add rationale for encoding info into names


garloff:
While we encourage to use the short names in the SCS flavor naming, we do have the capability to encode a lot of details into the flavor names. This is useful, as we may have (large) clouds, where customers have the ability to chose between very finegrained flavor types for selecting capabilities and we need to do this by chosing the flavor names. (The mechanism to use extra_specs and match it with images is broken -- requiring custom images just to be able to select the right VM flavor is a concept broken by design IMVHO.)
We should clarify this in the docs.

garloff:
Addressed by #121.

PR #121, opened 2022-09-16
Add text discussing the approach to flavors.

garloff:
In short, the naming proposal serves two purposes. (1) Provide transparency on features/details/capabilities (2) Allow users to chose the right features

There is not really good alternative for (2) to create names. We have chosen a systematic approach to it, which allows to generate/parse this programmatically (and manually if you are familiar with the spec), but which can lead to rather complex names in extreme cases.

To only achieve (1), better ways exist by using metadata (extra_specs). So smaller providers that don't need users to chose between so many flavors don't need to encode all the details in the name, but could use short names (which is allowed by the spec) and make details discoverable via a to-be-written metadata spec. This is the suggested approach going forward.

This is an attempt to address #73.

### DIFF
--- a/Design-Docs/flavor-naming-strategy.md
+++ b/Design-Docs/flavor-naming-strategy.md
@@ -47,7 +47,7 @@ The spec allows for flavors to understate its own capabilities and does not forc
 expose all details; we don't mandate the
 very detailed names that encode Disk types, CPU vendor, generation, speed grade, hypervisor
 into the name. Actually we explicitly recommend for flavors to be available under the
-short `SCS-nV:m:l` names (with `n`, `m`, `l` being a numbers that denote CPUs, RAM in GiB
+short `SCS-nV:m:l` names (with `n`, `m`, `l` being numbers that denote #CPUs, RAM in GiB
 and disk size in GB resp.) and even mandate a small number of them to be available.
 
 For providers that have moderate size and low variance in their offering, we would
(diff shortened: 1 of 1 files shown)


garloff:
Rendered text can be reviewed here: https://github.com/SovereignCloudStack/Docs/blob/feat/flavor-naming-strategy/Design-Docs/flavor-naming-strategy.md

Design-Docs/flavor-naming-strategy.md
```diff
@@ -0,0 +1,153 @@
...
+ GPU: .Type: Pass-Through GPU, .Brand: nVidia, .Gen: Ampere, #.CU/EU/SM: 64, Performance: Std Perf
+ No Infiniband
+```
+and learn in addition that the provider promises not to have insecure
+SMT enabled or other CPU vulnerability protections turned off and
+that your memory is protected by ECC or better and not oversubscribed.
+
+You can also use the interactive mode `-i` to construct flavor names.
+
+If you dislike the hazzle of parsing the names, you may want to consider them
```
fkr, line 91: hassle, not hazzle

fkr (approved):
LGTM except for hazzle -> hassle.

I'm also a big fan that we merge these kind of PRs rather quickly and do subsequent PRs to let them become better instead of having long running PRs - and it is marked 'draft' anyways.

fkr:
Further note: Thanks @garloff for writing this up! :rocket:

PR #126, opened 2022-09-20
Avoid using CentOS 8.4 in example.

garloff:
It should not have the minor number as there is no way to get patches for older minor versions, so it's one stream, not several.


fkr (approved):
LGTM!

PR #134, opened 2022-09-21
removals are removals and deprecations are deprecations

fkr:


garloff (approved):
makes sense - thanks!
/lgtm

PR #140, opened 2022-09-23
Add 0001_machine_identity.md ADR draft

joshmue:
Signed-off-by: Joshua Mühlfort <muehlfort@gonicus.de>

PR #144, opened 2022-09-30
Add Draft "Choice of identity" Decision Record

joshmue:
While having not worked extensively with any of the mentioned options yet (and possibly misunderstanding some details), I tried to give a brief assessment of possible options for machine (and human) identities which could be used in a generic manner to authenticate against cloud services.

This document must not be merged before #140 is accepted as SCS consensus decision. Until then, its purpose is to share my thoughts about this topic, enabling future discussions.


mbuechse:

joshmue:
As commented in #140, it's still generally relevant.

But: The decision what type of identity (document) should be used, may be more tilted towards OIDC style tokens, currently.

fkr:
same as #140 - closed, since nobody is taking care of this.

PR #150, opened 2022-11-15
Feat/add ssd flavor adr

garloff:
The ADR for adding 2 mandatory flavors with local SSD/NVMe for supporting etcd (k8s control plane node) and databases well.
Once we approve this, we will need to adjust the flavor naming standard accordingly.


Decisions/scs-xxxx-v1-ssd-flavors.md
```diff
@@ -0,0 +1,249 @@
...
+standards.
+
+This Decision Record is about adding a few mandatory flavors on the IaaS level
+that include flavors with local SSD (or better) storage.
+
+# Motivation
+
+The [currently defined standard flavors](https://github.com/SovereignCloudStack/Docs/blob/main/Design-Docs/flavor-naming.md)
+(as of v1.1 from 2022-09-08) do not include
+flavors that use local storage. For certain workloads such as big data (Hadoop),
```
fkr, line 28: I'd refrain from naming explicit products, if not really needed.

Decisions/scs-xxxx-v1-ssd-flavors.md
```diff
@@ -0,0 +1,249 @@
...
+databases, local storage is highly desirable as replication may be handled at
+the application layer, making replication/redundancy in a networked storage solution
+(ceph in the SCS reference implementation) an unneeded and undesired property.
```
fkr, line 31: In this case ("ceph in the SCS reference implementation") it makes sense to explicitly name it. (see comment above)

Decisions/scs-xxxx-v1-ssd-flavors.md
```diff
@@ -0,0 +1,249 @@
...
+### Heartbeat slowdown
+
+To avoid causing too many fail-overs by occasional high latencies, the
+frequency of heartbeats can be lowered from the default 1/100ms.
+The reelection timeout should change along with it (typically set to
+10 beats).
+
+This will cause etcd to take a bit more time to notice the loss of a node,
+which is not typically critical if done within reasonable limits.
+This change however is not a silver bullet -- occasional write latencies
```
fkr, line 103: This change gives no guarantee -- occasional write latencies

(reads imho better than the silver bullet wording)

Decisions/scs-xxxx-v1-ssd-flavors.md
```diff
@@ -0,0 +1,249 @@
...
+above 100ms will still cause failed heartbeats, just less often.
+
+This change has been implemented in SCS's
+(k8s-cluster-api-provider)[https://etcd.io/docs/v3.5/op-guide/hardware/#example-hardware-configurations]
+reference implementation: The heartbeat has been changed from 1/100ms (10/s)
+to 1/250ms (4/s) and the reelection timeout form 1s to 2.5s.
+
+The etcd process also is afforded a higher CPU priority (lower niceness),
+resulting in a lower scheduling latency, as high-prio processes preempt lower-prio
+ones when they get woken up. The etcd process also gets it's IO priority
```
fkr, line 113: "The etcd process also gets its IO priority"

fkr (commented):
Left a few small comments.
I like it - the "Motivation" part is too long for my taste. A lot of the items explained there are (imho) the category "general education for system engineers" and if we do this for every of these documents, we repeat ourselves a lot. But I might be biased and it is really helpful to have that here. 
The reason why I'd prefer a less verbose style: I fear that only few will actually read it.

Decisions/scs-xxxx-v1-ssd-flavors.md
```diff
@@ -0,0 +1,250 @@
+title: SSD Flavors
+type: Decision Record
+status: Draft
+track: IaaS
+version: v0.1
+enhances: scs-xxxx-v1-flavor-naming.md
```
joshmue, line 6: There are dashes missing in this line and in the first line.

Adding them enables the Markdown renderer to understand them. Like in SCS-0001: https://raw.githubusercontent.com/SovereignCloudStack/Docs/main/Standards/scs-0001-v1-sovereign-cloud-standards.md

Decisions/scs-xxxx-v1-ssd-flavors.md
garloff: Ack. I had the had hdfs in mind where a networked disk is a particularly bad match. I change the wording to "big data filesystem."

Decisions/scs-xxxx-v1-ssd-flavors.md
garloff: Thanks for clarifying.

Decisions/scs-xxxx-v1-ssd-flavors.md
garloff: Fixed.

garloff:
> Left a few small comments. I like it, very well written! The "Motivation" part is a tad too long for my taste.
Obviously, I was trying to create a positive example how to write such docs.
Being more concise in the Motivation part is probably something we do want. I'll have to re-read the doc and see if I see obvious ways to do it.

Decisions/scs-xxxx-v1-ssd-flavors.md
garloff: Good catch, thanks. Hopefully it's correct now.

Decisions/scs-xxxx-v1-ssd-flavors.md
```diff
@@ -0,0 +1,250 @@
...
+# Introduction
+
+SCS defines an IaaS Flavor Naming standard that mandates a number of standard flavors
+to be available in each SCS-compliant IaaS offering. While offering or exposing
+IaaS is not a requirement for SCS-compliant infrastructure offerings -- SCS allows
+for platforms only exposing the container layer (plus S3 compatible object storage)
+for gen2 (container-based) cloud-native workloads --
+the SCS reference implementation does include a complete IaaS implementation that
+many providers want to expose as they have customers desiring access at this layer
+for gen1 (VM-based) cloud-native workloads or for the virtualization of more
```
joshmue, line 19: Personally, I'd avoid categorising use cases in this much detail, as does not add too much value to the introduction and may be disputable.

Especially `gen1 (VM-based) cloud-native` strongly implies that Kubernetes is "gen2" cloud native, which might be correct, but definitely is kind of opinionated.
For example, someone might opt to avoid Kubernetes and instead build a cloud native setup by fully embracing the Hashicorp stack which in many ways kind of competes with Kubernetes: Packer for building images, Terraform for VM/resource provisioning, Nomad for workload scheduling, Consul Connect for service meshing etc.
They might not agree that this setup is a "gen1" (hence kind of legacy) setup, it's just different.

Of course, in my impression, using Kubernetes (or Kubernetes with some Hashicorp tools integrated) is the more popular choice - and also the choice SCS settled on as primary use case. Nevertheless, it might be better to focus only on the technological terms, here.

Decisions/scs-xxxx-v1-ssd-flavors.md
```diff
@@ -0,0 +1,252 @@
...
+
+This Decision Record is about adding a few mandatory flavors on the IaaS level
+that include flavors with local SSD (or better) storage.
+
+# Motivation
+
+The [currently defined standard flavors](https://github.com/SovereignCloudStack/Docs/blob/main/Design-Docs/flavor-naming.md)
+(as of v1.1 from 2022-09-08) do not include
+flavors that use local storage. For certain workloads such as big data filesystems,
+databases, local storage is highly desirable as replication may be handled at
```
joshmue, line 31:
```suggestion
flavors that use local storage. For certain workloads such as big data filesystems
or databases, local storage is highly desirable as replication may be handled at
```

Decisions/scs-xxxx-v1-ssd-flavors.md
```diff
@@ -0,0 +1,252 @@
...
+in scenarios where the etcd backend storage is redundant and not affected by the
+single-node outage.
+
+Neither scenario fulfills typical requirements for production workloads.
+
+### RAM (tmpfs) etcd storage
+
+etcd could keep its database in volatile memory (e.g. on a tmpfs filesystem).
+For multi-node etcd clusters, this would actually work, as long as at least one
+cluster member stays alive. A loss of power affecting all nodes or a hardware
```
joshmue, line 83: I remember that in a team call, this was a topic of discussion.

Regarding the `this would actually work, as long as at least one cluster member stays alive` bit:

According to my own experiences, and most importantly, answers of contributors in the etcd issues (https://github.com/etcd-io/etcd/issues/7445), this is not really all that easy.

According to this information, this would not work unless shut-down etcd members are manually removed, which I do not think is the default.

Decisions/scs-xxxx-v1-ssd-flavors.md
```diff
@@ -0,0 +1,252 @@
...
+The reelection timeout should change along with it (typically set to
+10 beats).
+
+This will cause etcd to take a bit more time to notice the loss of a node,
+which is not typically critical if done within reasonable limits.
+This change however does not fully address the issue -- occasional write latencies
+above 100ms will still cause failed heartbeats, just less often.
+
+This change has been implemented in SCS's
+(k8s-cluster-api-provider)[https://etcd.io/docs/v3.5/op-guide/hardware/#example-hardware-configurations]
```
joshmue, line 110:
```suggestion
[k8s-cluster-api-provider](https://etcd.io/docs/v3.5/op-guide/hardware/#example-hardware-configurations)
```

I guess this should be a markdown link.

Decisions/scs-xxxx-v1-ssd-flavors.md
```diff
@@ -0,0 +1,252 @@
...
+it has not been found a necessity for etcd in our tests.
+Disk IO QoS is not part of this spec but may be considered in another one.
+
+Live-migration with local storage is significantly more difficult than with
+networked storage: The contents of the local disks also need to be replicated
+over to the new host. Live-migration for these VMs may thus take significantly
+longer or not be possible at all, depending the configuration from the provider.
+Not supporting live-migration is OK for flavors with local disks according
+to the flavor naming spec -- a capability to indicate whether or not
+live-migration is supported will subject to a flavor-metadata spec that
```
joshmue, line 222:
```suggestion
live-migration is supported will be subject to a flavor-metadata spec that
```

joshmue (dismissed):
I like the "design consideration" section in particular: It is detailed, but concise.

matfechner (approved):
LGTM with the reguested changes

Decisions/scs-xxxx-v1-ssd-flavors.md
```diff
@@ -0,0 +1,252 @@
+---
+title: SSD Flavors
+type: Decision Record
+status: Draft
+track: IaaS
+version: v0.1
```
itrich, line 6: According to [SCS-0001#versioning](https://github.com/SovereignCloudStack/Docs/blob/main/Standards/scs-0001-v1-sovereign-cloud-standards.md#versioning), I'd advise to drop the version number within the frontmatter of this document. Minor changes (e.g. rewording or typo fixing) should be done without touching the version number. Major changes can be indicating by increasing the version number in the filename.

itrich:
Please don't forget to give this file a reasonable filename according to [SCS-0001#document-format](https://github.com/SovereignCloudStack/Docs/blob/main/Standards/scs-0001-v1-sovereign-cloud-standards.md#document-format) before merging this PR 😇

Decisions/scs-xxxx-v1-ssd-flavors.md
```diff
@@ -0,0 +1,252 @@
+---
+title: SSD Flavors
+type: Decision Record
+status: Draft
```
itrich, line 4: To get used to our own standard [SCS-0001](https://github.com/SovereignCloudStack/Docs/blob/main/Standards/scs-0001-v1-sovereign-cloud-standards.md#document-format), this should be in status `Proposal`. Only the very last commit before merging should
- rename the file
- bump the status to `Draft`

neuroserve:
What about pruning volumes from local SSDs in order to make sure, that tenants cannot access previous users data (or parts thereof) from that local SSD.

Decisions/scs-xxxx-v1-ssd-flavors.md
garloff: Well, the discussion tends to be different: The k8s community considers k8s as the one and only "cloud-native" approach, which is untrue, IMVHO. Cloud-native existed before containers were widespread, but the k8s folks want to monopolize the wording. They even created a foundation with that claim.
So the wording acknowledges that a lot of current ("gen-2") cloud-native development targets containers, while other and older approaches exist ("gen-1") and continue to be valid and used.
There is no rating involved here, merely the observation that we have seen two big waves in the industry here ... the 2nd is larger and more modern.

Decisions/scs-xxxx-v1-ssd-flavors.md
garloff: I would probably reverse the order, to avoid that the "big data" qualifier that is applied to filesystems also appears to apply to databases, which would be unintentional.

Decisions/scs-xxxx-v1-ssd-flavors.md
garloff: Yes, thanks.

Decisions/scs-xxxx-v1-ssd-flavors.md
```diff
@@ -0,0 +1,252 @@
...
+
+This Decision Record is about adding a few mandatory flavors on the IaaS level
+that include flavors with local SSD (or better) storage.
+
+# Motivation
+
+The [currently defined standard flavors](https://github.com/SovereignCloudStack/Docs/blob/main/Design-Docs/flavor-naming.md)
+(as of v1.1 from 2022-09-08) do not include
+flavors that use local storage. For certain workloads such as big data filesystems
+or databases, local storage is highly desirable as replication may be handled at
```
joshmue, line 31:
```suggestion
flavors that use local storage. For certain workloads such as databases or big data
filesystems, local storage is highly desirable as replication may be handled at
```

Decisions/scs-xxxx-v1-ssd-flavors.md
joshmue: > I would probably reverse the order, to avoid that the "big data" qualifier that is applied to filesystems also appears to apply to databases, which would be unintentional.

Done here: https://github.com/SovereignCloudStack/Docs/pull/150#discussion_r1032434391

Decisions/scs-xxxx-v1-ssd-flavors.md
joshmue: Continuation of https://github.com/SovereignCloudStack/Docs/pull/150#discussion_r1023737025

Decisions/scs-xxxx-v1-ssd-flavors.md
joshmue: Ah, ok. Over-interpreted the "gen" as rating, then :). Still, even when "gen" only is meant to be understood as hint regarding temporal order or history, such high level categorization in waves/generations still opens a lot of room for potential misunderstandings/discussion.

For example, Nomad was released first one year *after* Kubernetes' first release and is still advertised as alternative. Should it be considered gen-1 cloud-native, even it is "younger" than gen-2? Or gen-3?
What does gen-1 mean exactly in contrast to classical VM workload?

So, I think referring only to well-established high level terms like IaaS/... as well as concrete technical terms like VM/container would eliminate any such potential for misunderstandings or discussions unrelated to the actual topic of the decision record (like the one I just started).

garloff:
> What about pruning volumes from local SSDs in order to make sure, that tenants cannot access previous users data (or parts thereof) from that local SSD? Shouldn't that be covered in a separate paragraph?

Good point, the volumes need to be zeroed out before being handed out to a user.

Decisions/scs-xxxx-v1-ssd-flavors.md
garloff: Adjusted

Decisions/scs-xxxx-v1-ssd-flavors.md
garloff: OK, ok, I used the words wave 1 and wave 2 instead.

Decisions/scs-xxxx-v1-ssd-flavors.md
garloff: Alright, I improved the wording, mentioning that this could be made work and that proper care needs to be taken to remove shut-down nodes.

Decisions/scs-xxxx-v1-ssd-flavors.md
garloff: Question is whether we want to allow for minor version updates, if we e.g. do editorial changes that clarifies the wording ... and do this in the version header without bumping the major version (which may carry new, in the worst case even incompatible changes). It's still good to reference the precise version number - which may be better to parse for humans that git hashes.

Decisions/scs-xxxx-v1-ssd-flavors.md
garloff: You never know which is the last commit before merging, no?
So as soon as the original author thinks that this is ready, can't she assign the status to Draft which becomes then effective as soon as the PR is merged?

Decisions/scs-xxxx-v1-ssd-flavors.md
garloff: Changed the order of words to databases or big data filesystems, avoiding "big data" also potentially being interpreted as qualifier to databases.

garloff:
> Good point, the volumes need to be zeroed out before being handed out to a user.

Done:

Like with networked storage, the provider must ensure that data
from previous users is not accessible (e.g. by securely erasing it
or by using a different encryption key) when local storage gets
allocated to a new VM.

garloff:
Alright, I wanted to resolve a conversation (the one about purging volumes before reuse).
And accidentially closed the PR. Reopened now.
We need fresh approvals for being able to merge.

joshmue (approved):

Decisions/scs-xxxx-v1-ssd-flavors.md
fkr: @garloff *if* we use a version number, we should use semver imho.

garloff: Fair comment. So major.minor.patch with patches probably being clarifications, while minor versions might introduce new backwards compatible requirements (i.e. someone fulfilling v1.8.x automatically fulfills also v1.y.z with y<=7) and major bumps even possibly having requirements that conflict with earlier versions?
If we do so, the filename should probably encode major.minor, no?

garloff:
Any chance to get an approving review, @itrich , @fkr, @matfechner, @berendt, @joshmue so we can merge this as a *draft* standard?

joshmue (approved):

PR #154, opened 2022-11-21
Stabilize SCS-0001

itrich:
Signed-off-by: Eduard Itrich <eduard@itrich.net>

### DIFF
--- a/Standards/scs-0001-v1-sovereign-cloud-standards.md
+++ b/Standards/scs-0001-v1-sovereign-cloud-standards.md
@@ -1,7 +1,8 @@
 ---
 type: Procedural
-status: Draft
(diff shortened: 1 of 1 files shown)


itrich:
Hmm, I think that we should improve this process for the future. As far as I can recall from last Product Board meeting, we decided that we reach out to the community and wait _one_ week for further feedback before we officially stabilize this standard. Neither did the peer group completely review this standard draft.

I'm aware that this may be over-engineered for this rather well-discussed standard, but we should be absolutely clear, consistent and traceable about how we conclude standards.  Let's improve this for the future.

garloff:
You're correct, I was too fast on this one.

PR #156, opened 2022-11-24
correct link to flavor naming check script

jklippel:
Small correction to the link of the flavor name checking script.


tibeer (approved):
Thanks for that!

PR #158, opened 2022-11-27
Add a "Related Documents" optional section to consider.

garloff:
We should often have standards / decision records that relate to other SCS or upstream standards. So suggest to consider it.


fkr (approved):
good idea!

garloff:
Maybe also optionally linking to existing conformance tests? (Or is that another separate optional section?)

PR #163, opened 2022-12-15
Feat/flavor checker direct

garloff:
Rather than feeding the output from `openstack flavor list -f value -c Name` to `flavor-name-check.py`, we can create a tool that uses the openstacksdk to retrieve the flavor list and directly check.
This has one big advantage: We get some data (like vCPUs and RAM and root disk size) from OpenStack and can compare it with what's advertized in a name.
And indeed, found one warning (overprovided resources) on gx-scs and one error (underprovided resources) each on plus-pco and wavestack.

This gets us close to a complete conformance test for flavors.

Short of actually booting VMs and looking inside and doing long-term performance testing ...

### DIFF
--- a/Design-Docs/tools/flavor-names-openstack.py
+++ b/Design-Docs/tools/flavor-names-openstack.py
@@ -12,11 +12,12 @@
 # provided by openstack, such as the number of vCPUs and memory.
 # 
 # (c) Kurt Garloff <garloff@osb-alliance.com>, 12/2022
-# License: CC-BY-SA 4.0
+# SPDX-License-Identifier: CC-BY-SA 4.0
 
 import os, sys
 import openstack
 import importlib
+import yaml
 fnmck = importlib.import_module("flavor-name-check")
 
 def usage():
@@ -43,9 +44,10 @@ def main(argv):
         	print("You need to have OS_CLOUD set or pass --os-cloud=CLOUD.", file=sys.stderr)
 	conn = openstack.connect(cloud=cloud, timeout=32)
 	flavors = conn.compute.flavors()
-	nonSCSFlv = []
+	MSCSFlv = []
 	SCSFlv = []
 	wrongFlv = []
+	nonSCSFlv = []
 	warnFlv = []
 	errors = 0
 	for flv in flavors:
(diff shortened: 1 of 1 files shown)


tibeer:
Why was this pull request created? I thought there is a task that is passed to another person working on exactly this topic.

frosty-geek:
My guess is that "somebody' was bored on the sofa and wanted to do a little finger exercise ☺️

garloff:
> Why was this pull request created? I thought there is a task that is passed to another person working on exactly this topic.

Because we wanted to have
(a) the code that creates compliant flavors (flavor-manager) 
and
(b) the conformance check tool
created independently from each other.

This PR is completing the pre-existing work on (b).

I understand from your comment that you assumed working on (b) as well which would turn some of the work into duplicated efforts. 

Not ideal. Communication breakdown? 

I consider the tool as the reference for judging flavor conformity. If that has been superceded by some other tool, I have missed that news and the rationale I guess.

berendt:
The problem is that we have https://github.com/SovereignCloudStack/issues/issues/236 and the work on it is to be bundled by Tim in the new openstack-flavor-manager (https://github.com/osism/openstack-flavor-manager). If we continue to work here in parallel, this will lead to redundancies etc.

IMO we should stop working here and proceed in openstack-flavor-manager. The first code is already there and Tim is currently working on it.

garloff:
So the flavor-manager will do the SCS flavor spec conformance checking in the future?
Is that what we want? 

If so, we need to ensure it can be invoked on non-SCS OpenStack envs and be ran in a mode that does just that: SCS Conformance Checking.

I would then remove both pieces of the current test tool here to avoid confusion and to avoid the frustration of duplicated work.

The only piece remaining here would be a README, a YAML with the list of mandatory flavors plus - if we have a good way to express the grammar - a file that describes the naming grammar. (I fear that regexps are not the best way to denote grammars although I made heavy use of them in my parser.) 

(Probably the current code will be reused for the gx-self-description-generator, that was actually the cause for me looking at it again.)

fkr:
My suggestion would be as follows: 

- As a reference, I'd go ahead and merge this into the tree, since it does not harm to have more than implementation of the validation - especially _if_ they're authored individually.
- We should make sure the scope of the current work on the osism flavor manager is clear and easy to understand. I'll make sure to further outline the stuff that @tibeer and I discussed (which in tl;dr is part of the osism issue that @berendt referenced)

If need be, we can schedule a call to outline this together next week (I'm on PTO, but can submerge from that in order to not block the work from @tibeer on this).

garloff:
Discussed this in the IaaS team meeting:
* Conformance tests (and other artifacts used to validate standards compliance) live next to the standards in the standards repository (currently still: Docs/Design-Docs repo).
* The reference implementation can use/reference code from the standards repo to do its job as needed or alternatively do it's own independent implementation (possibly still consuming some artifacts from the standards repo) -- the canonical place for standards compliance though will always be the standards repo
* If the most mature conformance check implementation for a standard currently is somewhere else (e.g. in image-manager), we should split it out and move it to the standards repo
* The conformance checks in the standards repo should be linked with a framework into a testsuite that can be ran and integrated into a continuous conformance validation process

tibeer (commented):
Tested against PlusCloud Open:
```sh
columbia@columbia tools % ./flavor-names-openstack.py 
scs-community-beermann:
  FlavorReport:
    GoodSCSFlavors:
    - SCS-16V:64
    - SCS-16V:64:100
    - SCS-1V:0.5
    - SCS-1V:0.5:20
    - SCS-1V:1
    - SCS-1V:1:10
    - SCS-1V:1:20
    - SCS-2V:2
    - SCS-2V:2:20
    - SCS-2V:4:100
    - SCS-2V:4:20
    - SCS-2V:4:50
    - SCS-2V:8:100
    - SCS-4V:16:100
    - SCS-4V:32:50
    - SCS-4V:8:100
    - SCS-4V:8:50
    - SCS-8V:16:100
    - SCS-8V:32:50
    - SCS-8V:8
    - SCS-8V:8:100
    MandatoryFlavors:
    - SCS-16V:32
    - SCS-16V:32:100
    - SCS-1L:1
    - SCS-1L:1:5
    - SCS-1V:2
    - SCS-1V:2:5
    - SCS-1V:4
    - SCS-1V:4:10
    - SCS-1V:8
    - SCS-1V:8:20
    - SCS-2V:16
    - SCS-2V:16:50
    - SCS-2V:4
    - SCS-2V:4:10
    - SCS-2V:8
    - SCS-2V:8:20
    - SCS-4V:16
    - SCS-4V:16:50
    - SCS-4V:32
    - SCS-4V:32:100
    - SCS-4V:8
    - SCS-4V:8:20
    - SCS-8V:16
    - SCS-8V:16:50
    - SCS-8V:32
    - SCS-8V:32:100
    MissingFlavors: []
    WarnSCSFlavors: []
    WrongSCSFlavors: []
    nonSCSFlavors:
    - octavia
  FlavorSummary:
    GoodSCSFlavors: 21
    MandatoryFlavors: 26
    MissingFlavors: 0
    WrongSCSFlavors: 0
    errors: 0
    nonSCSFlavors: 1
    warnings: 0

columbia@columbia tools % 
```
I would assume, that the tool should report five missing flavors, since 26 are mandatory and only 21 are good.
Or do I misunderstand the output?

garloff:
You misinterpreted the output:
The tool found *all* 26 mandatory flavors.
It found an *additional* good 21 flavors, i.e. 21 flavors that comply with the SCS naming spec.
It also found one non SCS flavor, which is fine, as non-SCS flavors are allowed.
Thus all good, exit code zero.

What would not have been allowed:
- Missing flavors
- Flavors that start with "SCS-" but which do not comply with the spec rules (or which report parameters from the OpenStack glance API which are not aligned with what the names claim)

So from my perspective all good. Do we need better wording in the report? Or better documentation?

tibeer:
I would then change the output to something like this
```
  SCSFlavorSummary:
    TotalAmount: 47
    MandatoryFlavorsPresent: 26
    MandatoryFlavorsMissing: 0
    OptionalFlavorsValid: 21
    OptionalFlavorsWrong: 0
  OtherFlavorSummary:
    TotalAmount: 1
  TotalSummary:
    Errors: 0
    Warnings: 0
 ```
 What do you think?

garloff:
> I would then change the output to something like this
> 
> [...]
> 
> What do you think?

I like it and I implemented it. Please check!

tibeer (approved):
Great work! Ready to be merged.

PR #165, opened 2022-12-16
Fix counting SCS flavors.

garloff:


tibeer (changes requested):
Please add comments to point out, what the ´scs´ variable is used for. Moving variables to the top of a function is always good. But what I don't get is why `scs` is set to a fixed value of ´1´ when the first argument passed to the script is ´-i´.  ´-h´, ´-d´, ´-v´ and ´-c´ are self explanatory because of the variables used below (help, debug, verbose, check). ´-i´ is unclear for me.

garloff:
> Please add comments to point out, what the ´scs´ variable is used for. Moving variables to the top of a function is always good. But what I don't get is why `scs` is set to a fixed value of ´1´ when the first argument passed to the script is ´-i´. ´-h´, ´-d´, ´-v´ and ´-c´ are self explanatory because of the variables used below (help, debug, verbose, check). ´-i´ is unclear for me.

I added a comment. `-i` is explained in the `usage()` function, when `-h` (help) is used.
The `scs` variable counts the good SCS flavors. When one is input interactively (with `-i`), we have at least one, that's why we set `scs` to one there. Only used for reporting.

I hope the code readability is better now.

tibeer:
Seems like some flavors are not accepted. We created a tool that generates a huge amount of possible flavor names according to the ruleset. The tool can be found here: https://github.com/osism/openstack-flavor-manager/tree/main/scs_flavor_generator

Since the generated file is too large (>4GB), I needed to apply some trickery to validate every line:
```sh
while read line; do if ! python3 flavor-name-check.py $line; then echo $line; fi; done </Users/columbia/Documents/osism/openstack-flavor-manager/scs_flavor_generator/flavors1.txt
```

The validator marks flavors like `SCS-1L:0.5-z-GI1hhh` as invalid, even though they are valid according to the rules.

tibeer (changes requested):
Seems like some flavors are not accepted. We created a tool that generates a huge amount of possible flavor names according to the ruleset. The tool can be found here: https://github.com/osism/openstack-flavor-manager/tree/main/scs_flavor_generator

Since the generated file is too large (>4GB), I needed to apply some trickery to validate every line:
```sh
while read line; do if ! python3 flavor-name-check.py $line; then echo $line; fi; done </Users/columbia/Documents/osism/openstack-flavor-manager/scs_flavor_generator/flavors1.txt
```

The validator marks flavors like `SCS-1L:0.5-z-GI1hhh` as invalid, even though they are valid according to the rules.

frosty-geek:
May I suggest that you only generate a list of mandatory flavors then the list should be only like 30ish flavors long.

Then I would query the list of flavors a CSP provides and check those against the standard.

Even with our extensive flavors we've offered in the past (we've reduced the flavor list around 10/2022) we never had more than 300 flavors, thus I don't understand the point of a 4GB long list.

Ralf

garloff:
> Seems like some flavors are not accepted. We created a tool that generates a huge amount of possible flavor names according to the ruleset. The tool can be found here: https://github.com/osism/openstack-flavor-manager/tree/main/scs_flavor_generator
> 
> Since the generated file is too large (>4GB), I needed to apply some trickery to validate every line:
> 
> ```shell
> while read line; do if ! python3 flavor-name-check.py $line; then echo $line; fi; done </path/to/huge/list/of/flavors.txt
> ```
> 
> The validator marks flavors like `SCS-1L:0.5-z-GI1hhh` as invalid, even though they are valid according to the rules.

An easier test-case is `SCS-1L:0.5-GI1h`, which is rejected.
According to the spec it should be rejected (!), but then the tool does allow to generate such a name with `-i`.
We can become stricter in what we generate (approach A) or allow this (and adjust the tool and spec -- approach B).
See discussion in the first commit message here (going for approach B):
https://github.com/SovereignCloudStack/Docs/compare/main...fix/gpu-highperf

Thanks for generating so much test data, I will look at it beyond this issue.

garloff:
The other possible resolution can be seen here:
https://github.com/SovereignCloudStack/Docs/compare/main...fix/gpu-highperf2
We continue to disallow `-GI1h` and now ensure that `flavor-name-check.py -i` does not produce it.

tibeer (approved):

PR #168, opened 2022-12-19
Add decision record template

joshmue:
Providing a simple template to get started with writing a decision record should be helpful when creating decision records every now and then.


joshmue:
I almost created a second PR with a template with almost identical contents in the `Standards` directory.

Maybe dropping the `Decisions` directory and putting _all_ documents ("decision records", "standards" and "procedural") into the `Standards` directory would make more sense?

garloff (approved):
I like it!

garloff:
> I almost created a second PR with a template with almost identical contents in the `Standards` directory.
> 
> Maybe dropping the `Decisions` directory and putting _all_ documents ("decision records", "standards" and "procedural") into the `Standards` directory would make more sense?

There are significant differences between a standard (rules how an implementation needs to behave) and decisions (which *could* touch on implementation details of the reference implementation).
But, putting the user/consumer hat one, one might not know ahead of time, whether SCS may have an ADR that is implementation specific or maybe a standard to a certain topic and having to search at two places could be very confusing. SO you'd have my +1 on this proposal.

PR #169, opened 2022-12-19
Add draft for KaaS LB standard

joshmue:
Addresses https://github.com/SovereignCloudStack/issues/issues/227


joshmue:
Per discussions in the latest container team meeting, for this decision record, we'll de-scope client IP preservation completely.

To sum up:

- enabling the health monitor at octavia by default will help ensuring that setting `externalTrafficPolicy: Local` will not result in frequent connectivity issues, which is the major point for this decision record anyway.
- `externalTrafficPolicy: Local` will *not* have the effect of that the actual client IP is visible to the application server. It will only cause Kubernetes to stop doing SNAT, which will only expose the direct TCP client IP - which is the Octavia HAProxy.

So currently, client IP preservation - at least on IP level - does not seem to be possible.

Comparing with other clouds, *as far as I can tell*:
Google and Azure implement Load Balancers at at lower (**router**) level (aka DSR?), making preserving the actual source IP super easy. (Without understanding specifics, [AWS also somehow implements client IP preservation](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-target-groups.html#client-ip-preservation)).
OpenStack Octavia makes use of higher (**proxy**) level, [requiring an extra implementation at this high level](https://kubernetes.io/docs/tutorials/services/source-ip/#cross-platform-support).

TODO: Even though this will not trivial to implement, create a backlog item for this, as feature parity with major clouds is desired.

joshmue:
I pretty much rewrote the record.

I know that the result of the container team call was to actively de-scope client IP preservation from this document, but there is too much pointing towards it to be able to just omit it.

I added a few considerations and three more detailed ways to go.

Standards/scs-XXXX-vN-kaas_load_balancers.md
```diff
@@ -0,0 +1,126 @@
...
+* [Improve performance](#performance)
+  * Significant improvements are possible, but are not validated yet
+
+### Conclusion (**TBD**)
+
+| Option 1 | Option 2 | Option 3 |
+|----|----|----|
+| DO require SCS compliant clouds to work with `externalTrafficPolicy: Local` (in the reference implementation: enabling the health check mechanism to avoid constant connectivity problems); Do NOT require them to preserve the client IP | DO NOT require SCS compliant cloud providers to support `externalTrafficPolicy: Local` (in the reference implementation: enabling the health check mechanism to avoid constant connectivity problems if users set it anyway) | DO NOT require SCS compliant cloud providers to support `externalTrafficPolicy: Local` (Do NOT configure health checks in the reference implementation in order to not derive from upstream defaults) |
+
```
garloff, line 54: There is another option (maybe option 0).
Require SCS compliant cloud managers to support TrafficPolicy: Local (e.g. by enabling health monitors) AND require them to preserve the original requester's IP address (e.g. by enabling some PROXY protocol setup in a way that it's transparent to the L7 nginx.

Standards/scs-XXXX-vN-kaas_load_balancers.md
```diff
@@ -0,0 +1,126 @@
...
+# Decision
+
+* A Kubernetes `Service` of `type=LoadBalancer` with all non-madatory fields being unset: Must work as expected, out of the box
```
garloff, line 57: s/madatory/mandatory/ (spelling)

Standards/scs-XXXX-vN-kaas_load_balancers.md
```diff
@@ -0,0 +1,126 @@
...
+* A Kubernetes `Service` of `type=LoadBalancer` with `externalTrafficPolicy: Local` set and all other non-madatory fields being unset: **TBD**
```
garloff, line 58: ditto

Standards/scs-XXXX-vN-kaas_load_balancers.md
```diff
@@ -0,0 +1,126 @@
...
+
+# Conformance Tests
+
+With default `externalTrafficPolicy: Cluster`:
```
garloff, line 62: I guess I would add the conformance test here to the Design-Docs/tools repository. (This will become the standards/tools repository in the future), rather than including it here and then reference it. This would have the advantage that the script can be called from the scs-conformance-check.py (see PR #182).

garloff (changes requested):
Very well written ADR, thanks!
I do however have a few suggestions, which I'd like us to consider before we finalize this.

Note: We do have the option to defer the IP preservation to a secondary ADR, but I do guess we need another discussion: Supporting TrafficPolicy: Local without supporting IP preservation might be a confusing outcome. On the other hand currently (withou health monitor) things appear to work and then break by only having every third request (in a cluster with 3 worker nodes and one nginx replica) reaching nginx. Much worse breakage, more easily detectable though.

Standards/scs-XXXX-vN-kaas_load_balancers.md
joshmue: Alright! Removed it here, can add another test which fits to the ultimate decision later.

Standards/scs-XXXX-vN-kaas_load_balancers.md
joshmue: I added a bunch of options and wanted them to "build upon" each other; This one would be number 5 right now.

cah-hbaum:
> This PR only requires the bare minimum (a working `Service` implementation of `type=LoadBalancer`), so I do not see any blocker for this PR.
> 
> Of course, this also means that the standard does not provide too much value other than showing off alternatives, being the basis for other standards, or kick starting implementation of e2e tests. So technically, it would not really matter whether this is merged or abandoned. I'd opt to do either, though.
> 
> @mbuechse @garloff @jschoone How to proceed?

I'm neither of the mentioned people, but IMO this is a good document. But I agree with your assessment that the overall value isn't that big. While reading it, I had the impression, that this is more of a Decision Record. So personally, I would just change this to a DR and take it as a base for a future standard.

joshmue:
> I'm neither of the mentioned people

Your feedback is welcome, of course. Thanks! Just wanted to limit to mentioning as much as possible.

> But I agree with your assessment that the overall value isn't that big. While reading it, I had the impression, that this is more of a Decision Record. So personally, I would just change this to a DR and take it as a base for a future standard.

Indeed, I structured it a little bit DR-like. I'm still a bit unsure whether wording should be adjusted to be more standard-like or making it a DR completely. Mostly because this would have been most likely a standard if the result had been different. Should that make a difference? Also, it's been more that a year now, so the decision may be reevaluated. :man_shrugging:

jschoone:
> > I'm neither of the mentioned people
> 
> Your feedback is welcome, of course. Thanks! Just wanted to limit to mentioning as much as possible.
> 
> > But I agree with your assessment that the overall value isn't that big. While reading it, I had the impression, that this is more of a Decision Record. So personally, I would just change this to a DR and take it as a base for a future standard.
> 
> Indeed, I structured it a little bit DR-like. I'm still a bit unsure whether wording should be adjusted to be more standard-like or making it a DR completely. Mostly because this would have been most likely a standard if the result had been different. Should that make a difference? Also, it's been more that a year now, so the decision may be reevaluated. 🤷‍♂️

Hi @cah-hbaum and @joshmue,
sorry for the very late response, I was pretty sure I answered :facepalm: 
Now I forgot what it was.
I also think there's no blocker and I like the idea from Hannes to just make a DR out of it

mbuechse:

joshmue:
Seems like I cannot change the source branch to some branch in this repo (not directly anyway).

@cah-hbaum would you prefer to

1) close this and create a new PR from a new branch in this repo OR
2) merge this PR into a new branch in this repo, then create another PR from the newly created branch to main?

cah-hbaum:
@josmue Please merge, we will address other comments in the following PR.

cah-hbaum (approved):
Approved here, will review in the upcoming PR!

PR #170, opened 2022-12-21
Add draft for "KaaS default StorageClass" standard

joshmue:
Addresses https://github.com/SovereignCloudStack/issues/issues/213.


garloff (commented):
Very well written document, thanks!
This reflects what we had agreed upon previously.
We have deviated  from that agreement a bit last week by discussing in detail how further storage classes would differ in performance (IOPS ...) and in wanting to delay this standard a bit until we have a clear picture of those.

Let's see whether we can get away with one sentence that
- The application can not expect performance or latency guarantees for volumes of this storage class. More than hundred IOPS and the protection against single-disk / single-host failure at a good price (in the order of 0.1€ per GB per month as of 2023)

joshmue:
@garloff Addressed the performance/pricing bit - in a very general manner. If you want, we could also add the fixed numbers you mentioned as orientation.

garloff:
> @garloff Addressed the performance/pricing bit - in a very general manner. If you want, we could also add the fixed numbers you mentioned as orientation.

Looks good to me!

horazont (dismissed):

garloff:
One more thing: (OK, call me Columbo if you like!)
- We decided we would like to require encryption at rest for our default storage class (protecting customer data against improperly disposed storage media)
- This was conditional on the IaaS' team evaluating the performance impact and always providing encrypted storage in our reference implementation and requiring providers to have it in our standards

So do we want to wait for this?

garloff (approved):
Adding a number should not have dismissed the positive review from @horazont.
And I also agree with this and we have discussed today that we want to advance this one into the merged Draft state.

PR #166, opened 2022-12-16
Feat/read mand flavors

garloff:
This implements https://github.com/SovereignCloudStack/issues/issues/235

PR #177, opened 2023-01-11
Add linters for

tibeer:
relates #166

- python
- markdown

Reason: Improve code quality

Signed-off-by: Tim Beermann <beermann@osism.tech>


tibeer:
@garloff please review the errors reported by the linter. If they are not valid for you, we need to adjust the linter rules (_.flake8_)

.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
+ignore = E501
```
garloff, line 2: If we could configure the check to limit the line length to 120 chars, I would actually leave it on.
But without that, disabling it completely indeed avoids a lot of annoyance. Screens are larger than 80x25 in this decade (and the previous one).

.github/workflows/lint-markdown.yml
```diff
@@ -0,0 +1,18 @@
...
+    paths:
+      - '**.md'
+      - .github/workflows/lint-markdown.yml
+
+jobs:
+  lint-markdown-syntax:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: reviewdog/action-markdownlint@v0
```
garloff, line 15: This is some github builtin markdown linter?
I have no idea what it does ...
But yes, in general, getting some consistency in what markdown we put in there is definitely something I appreciate.

.github/workflows/lint-python.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  lint-python-syntax:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.x'
+      - run: pip3 install flake8
+      - run: flake8
```
garloff, line 19: flake8 is a good baseline. Nothing to disagree with (once E501 is disabled or configured differently) -- except maybe the two empty lines required b/f every function, but this is minor.

I tend to use pylint which goes much further in its advice. But obviously we can not require 10/10 points. Maybe 9/10?

Or maybe we leave flake8 as CI in there and recommend coders to use pylint in addition?

garloff (dismissed):
Should we submit cleanup patches to all .py and .md files to pass prior to merging this?

.github/workflows/lint-python.yml
garloff: Well actually, I would vote for disabling E226 as well (enforcing spaces around operators).
While I'm not against being forced to write ` a = b + 1` instead of `a = b+1`, I might find `a = 2*b +1` better readable than `a = 2 * b + 1`. (And I would certainly want to enforce the spaces around the assignment operator.) More importantly `substr = stg[idx+1:]` is significantly more readable than `substr = stg[idx + 1:]`, which E226 complains about unfortunately.
Obviously, I know this is a question of taste ... which is hard to argue about.

garloff: Going though more software, I ended up disabling:
```
    E251 unexpected spaces around keyword / parameter equals
    E226 missing whitespace around arithmetic operator
    E221 multiple spaces before operator
    E501 line too long (XXX > 79 characters)
```
E251: Readability can be improved by spaces around keyword arguments (when you have long lists with multiple lines
E221: Aligning = statements in lines below each other can improve readability

garloff: Maybe I should explain by example:
E251
`
retval = function(kwd1 = value1+1,
                  kwd2 = function2(value2, kwdarg=value3),
                  kwd3 = astring[-4:])
`
is less readable without the spaces around the kwd[123] = assignments, IMVHO.

garloff: E221:
```
attr1 = obj.attr1
at2   = obj.at2
att3  = obj.att3
```
or
```
val1 = expr1 + 1*val
val2 = ex2  + 2*val
val3 = exp3  + 3*val
```
visually helps to avoid mistakes.

garloff (approved):
I suggest to merge these improvements.
If we can agree to become stricter in the future and fix up code such that it complies to more rules, we can always still do so. (Not however that in my personal taste, these rules would make some things uglier.)

.github/workflows/lint-markdown.yml
tibeer: @garloff these are repositories in general. So something like "reviewdog/action-markdownlint" translates to github.com/reviewdog/action-markdownlint

PR #182, opened 2023-01-14
Feat/compliance tool

garloff:


garloff:
Please review, so we can move forward!
Please see header comments (docstring) in scs-compliance-check.py and run against your cloud!
@cedegi, @frosty-geek, @berendt.

Standards/scs-xxxx-v1-sovereign-cloud-standards-yaml.md
```diff
@@ -101,7 +109,8 @@ Every list of standards consists of several standards that – altogether – de
 |-----	|------	|-------------	|---------- |
 | `{layer}.versions.standards.name` | String | Full name of the particular standard | _Flavor naming_ |
 | `{layer}.versions.standards.url` | String |  Valid URL to the latest raw version of the particular standard  | _https://raw.githubusercontent.com/SovereignCloudStack/Docs/main/Standards/SCS-0003-v1-flavor-naming.md_ |
-| `{layer}.versions.standards.check_tool` | String | Valid URL to the latest check tool that verifies compliance with the particular standard | _https://raw.githubusercontent.com/SovereignCloudStack/Docs/main/Design-Docs/tools/flavor-name-check.py_ |
+| `{layer}.versions.standards.check_tool` | String | Valid local filename or URL to the latest check tool that verifies compliance with the particular standard | _https://raw.githubusercontent.com/SovereignCloudStack/Docs/main/Design-Docs/tools/flavor-name-check.py_ |
```
itrich, line 112: Should we refactor this to an array of `{layer}.versions.{version}.standards.check_tool.url` and `{layer}.versions.{version}.standards.check_tool.args`?

If we ever have to – for whatever reason – add multiple check tooks for a standard, we could simply append further tools to the overall YAML key {layer}.versions.{version}.standards.check_tool`.

Design-Docs/tools/scs-compatible.yaml
```diff
@@ -0,0 +1,35 @@
+name: SCS Compatible
+url: https://raw.githubusercontent.com/SovereignCloudStack/Docs/main/Design-Docs/tools/scs-compatible.yaml
+iaas:
+  versions:
```
tibeer, line 4: Do we really need another sub-key? I would prefer to directly start the array.

Design-Docs/tools/scs-compatible.yaml
```diff
@@ -0,0 +1,35 @@
...
+    - version: v2
+      stabilized_at: 2023-10-03
+      standards:
+        - name: Flavor naming
+          url: https://raw.githubusercontent.com/SovereignCloudStack/Docs/main/Standards/SCS-0003-v2-flavor-naming.md
+          check_tool: ./flavor-names-openstack.py
```
tibeer, line 10: ditch the leading "./". URLs can be easily identified by a leading "http" string.
By removing the leading "./" you'll also add the ability to expand the tool to auto-execute the programs in the future even when they reside in a different path.

tibeer: same for the following check_tool parameters

garloff:
I would like us to merge this rather sooner than later.
Any concerns left or questions that need answering?

Design-Docs/tools/scs-compliance-check.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+def run_check_tool(executable, args, verbose=False, quiet=False):
+    "Run executable and return exit code"
+    exe = [executable, ]
+    if args:
+        exe.extend(args.split(" "))
+    # print(f"{exe}")
+    # compl = subprocess.run(exe, capture_output=True, text=True, check=False)
+    compl = subprocess.run(exe, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
```
tibeer, line 65: Just an idea which might lead to a separate issue: Instead of using subprocess, import the files as libraries dynamically? E.g. like this http://web.archive.org/web/20120315061111/http://www.diveintopython.net/functional_programming/dynamic_import.html

tibeer (commented):
See comments. All in all i would prefer more comments :)

tibeer:
Relates SovereignCloudStack/issues#236

Design-Docs/tools/scs-compatible.yaml
itrich: We used to need the sub-key as there were (in an older version of the YAML standard) further keys on that level (e.g. `upcoming`, `lates`, ...). We could agree on ditching the extra key. What do you think @garloff?

Design-Docs/tools/scs-compliance-check.py
itrich: What if a test (in the future) won't be a Python script but rather some binary executable?

garloff: I would keep the design generic and have the ability here to run any executable, e.g. also shell scripts.

Design-Docs/tools/scs-compatible.yaml
garloff: Well, `./subdirectory/tool` would have worked without having to change the code.

To beautify this somewhat, I added to the spec that all check_tool file names are relative to the scs-compliance-tool.py path, which simplifies calling it from somewhere else. (Assuming that the main check tool and the individual checkers remain in the same git repo, this will work well.)
Implemented in commit adc8fc3f985997dc051747b93da643b4b17aab64.

Design-Docs/tools/scs-compatible.yaml
garloff: You mean like:
```yaml
iaas:
  - version: v2
    stabilized_at: yyyy-mm-dd
    [...]
  - version: v1
    stabilized_at: jjjj-mm-tt
```
Works for me.

Standards/scs-xxxx-v1-sovereign-cloud-standards-yaml.md
garloff: Having the ability to list multiple tools has the advantage potentially saving us a wrapper that calls multiple checking tools.
```yaml
iaas:
  - version: v2
    stabilized_at: 2525-02-25
    standards:
      - name: Flavor naming
        url: https://...
        check_tool:
          - url: filename_or_url
            args: -v
          - url: filename2
```
I have no strong opinion on this.

Design-Docs/tools/scs-compatible.yaml
garloff: > same for the following check_tool parameters

I don't understand this comment, sorry.
The standards[].check_tool_args: takes a string that is passed to the tool (split by whitespace)

Standards/scs-xxxx-v1-sovereign-cloud-standards-yaml.md
garloff: In case it's not obvious: The logic of course would be that all tools need to pass, i.e. we just add the errors.

Design-Docs/tools/scs-compatible.yaml
garloff: See branch feat/compliance-tool-drop-versions

Standards/scs-xxxx-v1-sovereign-cloud-standards-yaml.md
garloff: See branch feat/compliance-tool-drop-cersions-add-tool-array.
Note that I renamed check_tool to check_tools to indicate it's an array.

Remaining question: Is the condition flags per check tool or per standard?
If for example a standard gets enhanced with additional requirements, I could image that the new properties are being tested in a new tool. In v1, we might still call this new tool with the optional condition, generating a report that allows us to understand how far we have to still go. In v2, it would then switch to mandatory ...

Design-Docs/tools/scs-compatible.yaml
tibeer: Fine for me. The second command related to line 14, line 25, line 29 as they all contain the "check_tool" key.

tibeer (approved):
LGTM!

itrich (approved):
With all the changes merged recently, this looks very good to me. We should merge this and propose to stabilize the ADR on cert-spec.yaml (in an additional PR).

itrich (approved):
LGTM!

garloff:
Sorry, my editorial change dismissed your review, I need to ask for a re-review, @itrich @tibeer

tibeer (approved):

PR #184, opened 2023-01-16
Add Standard document for version policy regarding k8s updates

ajfriesen:
Here is my first draft regarding the version policy.


joshmue:
Nice!

Right now, the record primarily concerns the speed of picking up new Kubernetes releases.

There are also a few other relevant topics, which may end up (a) not standardized, (b) in this record or (c) as a series of records:

## Support of non-current Kubernetes versions / pace of EOL-ing

What older releases is the provider required to support? All non-current "Minor" releases that are still supported upstream? Maybe more or less? Are patch releases treated differently?

## Responsibilities provider vs. customer

Does the provider prepare version upgrades, but the customer triggers the upgrade on their schedule? Perhaps use another model for patch releases?

garloff (commented):
Document looks really good to me.
A few minor suggestions:
1.) Please add the DCO (Signed-Off-By) to your commit message, otherwise we can not merge the PR. `git --amend -s` and then force push (`--force-with-lease`) to the branch.
2.) A small section that says that there how long minor versions stay supported is subject to another Standard document. (The comment from Joshua is a good one, of course!)

ajfriesen:
We discussed that we want to create different ADR for the following topics:
- How long do scs compliant providers need to support versions
- Who does the upgrade of the cluster (user of that cluster or the provider or some mixed mode?)

garloff (dismissed):
/lgtm

garloff (approved):
Looks still good.

PR #187, opened 2023-01-20
Write result of compliance check to given path

itrich:
This change adds the feature the output the result of the compliance check to a given path with option `-o` / `--output`, e.g. 
```
./scs-compliance-check.py scs-compatible.yaml iaas -o report.yaml
```

There is still an open TODO that I couldn't solve yet. I want to add the overall result to {layer}.versions.{version}.result. Any idea?

### DIFF
--- a/Design-Docs/tools/scs-compliance-check.py
+++ b/Design-Docs/tools/scs-compliance-check.py
@@ -40,7 +40,7 @@ def usage():
     print(" -d/--date YYYY-MM-DD: Check standards valid on specified date instead of today")
     print(" -V/--version VERS: Force version VERS of the standard (instead of deriving from date)")
     print(" -c/--os-cloud CLOUD: Use specified cloud env (instead of OS_CLOUD env var)")
-
+    print(" -o/--output path: Generate report of compliance check under given path")
 
 def is_valid_standard(now, stable, obsolete):
     "Check if now is after stable and not after obsolete"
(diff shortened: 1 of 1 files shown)


Design-Docs/tools/scs-compliance-check.py
```diff
@@ -180,14 +186,21 @@ def main(argv):
...
+                    standard.update({"result" : error})
             if not optional:
                 errors += error
             if not quiet and "check_tool" in standard:
                 print(f"... returned {error}")
+            if output:
+                report.update({layer : {"versions": [bestversion]}})
             for kwd in standard:
-                if kwd not in ('check_tool', 'check_tool_args', 'url', 'name', 'condition'):
+                if kwd not in ('check_tool', 'check_tool_args', 'url', 'name', 'condition', 'result'):
```
garloff, line 198: result is still not a valid keyword in the standard specification, no?
So we should continue to complain if we find it.

itrich: You're right! I removed the key and changed how the result is collected. It now complains that `standard.result` is not a valid keyword, but I'm actually only manipulating the (copied) dictionary `report`. I'll have to continue investigating.

garloff: copy.deepcopy() creates a real copy, but makes the index() function fail.
We don't need it though, it's always 0.

PR #199, opened 2023-01-30
Stabilize the SSD flavor decision.

garloff:
Note: This means that our plan to implement these and require these at a later point in time in our list of mandatory flavors in the flavor-naming-spec.md. The precise time when these will become required is to be determined in a separate discussion. (My current assumption is shortly after R4, i.e. April 2023.)

### DIFF
--- a/Decisions/scs-0110-v1-ssd-flavors.md
+++ b/Decisions/scs-0110-v1-ssd-flavors.md
@@ -1,7 +1,8 @@
 ---
 title: SSD Flavors
 type: Decision Record
-status: Draft
+status: Stable
+stabilized_at: 2023-02-08
 track: IaaS
 enhances: scs-xxxx-v1-flavor-naming.md
 ---


garloff:
Mark as Work in Progress, do not merge prior to 2023-02-08.

itrich:
/remind-me to stabilize this ADR on 2023-02-08

fkr (changes requested):
Before this is stablized, the primary Flavor Naming should be adjusted to reflect the change in regard to `:`, since this ADR builds on top of that and introduces flavors that (as of now) carry the `:`.

garloff:
The merged PR #226 took care of updating this ADR to the new syntax for flavor names.
Please re-review, so we can stabilize this one.
Next steps IMHO:
(1) Understand whether or not we have these new flavors implemented in the SCS clouds soon (aka R4) or later
(2) Based on the understanding, decide whether the addition of these two new *mandatory* flavors should be (a) merged in the v2 flavor spec now or (b) whether we should hold off and add those two late.

frosty-geek:
> Before this is stablized, the primary Flavor Naming should be adjusted to reflect the change in regard to `:`, since this ADR builds on top of that and introduces flavors that (as of now) carry the `:`.

+1 to me this is hard blocked before a final decision is made on the future of flavor naming v2

garloff:
> > Before this is stablized, the primary Flavor Naming should be adjusted to reflect the change in regard to `:`, since this ADR builds on top of that and introduces flavors that (as of now) carry the `:`.
> 
> +1 to me this is hard blocked until a final decision is made on the future of flavor naming v2

There are a few decisions that are independent from each other:

(a) Do we want (in the future) to mandate the availability of certain flavors with certain capabilities at all?
  (a1) If so, should the two new SSD flavors be part of the list and when?
(b) Do we want to mandate a standardized naming scheme for the flavors in the future?

So, even if we turn the naming scheme into a recommendation in the future (so we drop (b)), we might still say yes to (a) and (a1).

fkr:
@fkr to push this forward (alongside @garloff, @berendt and one stakeholder from the CSP community to finalize in CW 22.

mbuechse:
Superseded by https://github.com/SovereignCloudStack/standards/pull/304.

PR #202, opened 2023-01-31
Move feature-rich scs-compatible.yaml to -test.yaml.

garloff:
Implement a minimal scs-compatible.yaml that matches what we consider the mandatory IaaS standard currently.


tibeer (approved):
LGTM!

PR #206, opened 2023-02-07
Restructure standards repository according to scs-0002-v1

itrich:
This PR restructures this repository according to [scs-0002-v1-standards-docs-org](https://github.com/SovereignCloudStack/standards/blob/main/Decisions/scs-0002-v1-standards-docs-org.md).

- `tools` has been moved to `Tests` with a subdirectory for each layer and test
- Certification YAML spec have been adapted accordingly
- `Design-Docs` has been renamed to `Drafts`

We still need to discuss where to move `Operational-Docs` to. As far as I remember, @maxwolfs is already working on this, so this directory remains unchanged for now.


garloff:
The flavor-name checking tools search for `SCS-Spec.MandatoryFlavors.yaml` in the current working directory and in the directory where the checking tool itself resides. We could obviously add the parent directory to of the script directory if we wanted.
You have left the SCS-Spec.MandatoryFlavors.yaml` in the top level directory `Tests`.
I don't think it belongs there.
We could argue that it belongs to `Tests/iaas/` or to `Tests/iaas/flavor-naming/` -- I can see good reasons for both.

garloff:
> You have left the `SCS-Spec.MandatoryFlavors.yaml` in the top level directory `Tests/`. I don't think it belongs there. We could argue that it belongs to `Tests/iaas/` or to `Tests/iaas/flavor-naming/` -- I can see good reasons for both.

See feat/comply-with-scs-0002-move-flavorlist branch for an implementation of the `SCS-Spec.MandatoryFlavors.yaml` moved to `Tests/iaas/`.

itrich:
You're right! I somehow have forgotten to move the `SCS-Spec.MandatoryFlavors.yaml` to the `iaas` directory. I've seen that you have done this in a dedicated branch. However, I'd rather would advocate to keep the YAML file alongside the script within the `Tests/iaas/flavor-naming/` directory. What do you think?

garloff:
No strong opinion.
The Spec file is intended to be consumable by external tools, such as e.g. the flavor-manager. That's why I don't want to bury it too deep down and assigned a right to live on it's own, independently from the script that consumes it.
Thus the `Tests/iaas/` thought.
But you can as well consider it to be part of the checking script and move it there then. We shoudl then just refrain from using it around much in the future, to avoid breakage ...

itrich:
> No strong opinion. The Spec file is intended to be consumable by external tools, such as e.g. the flavor-manager. That's why I don't want to bury it too deep down and assigned a right to live on it's own, independently from the script that consumes it. Thus the `Tests/iaas/` thought. But you can as well consider it to be part of the checking script and move it there then. We should then just refrain from moving it around much in the future, to avoid breakage ...

I neither have a strong opinion on this. Keeping the specification YAML alongside the script feels more "clean", but this is just an arbitrary and opinionated feeling. I get your point and I'm happy with both places.

What do you think @tibeer?

tibeer (approved):
As far as I can tell, this looks good to me.

Issue #192, opened 2023-01-26
Flavor naming adjustment considerations


garloff:
In issue #190 we learn that OpenShift is using IaaS flavor names as k8s annotations (taints or tolerations?).
This means that in order for OpenShift to run well on SCS IaaS, adjustments need to be made.
At this point, OpenShift is the only tool known to us that imposes such restrictions on IaaS choices, but we need to watch out for other restrictions that may come down from tools in the k8s world.

Assuming that we need to adjust our own [flavor spec](https://github.com/SovereignCloudStack/standards/blob/main/Design-Docs/flavor-naming.md), thus following option 3 in https://github.com/SovereignCloudStack/standards/issues/190#issuecomment-1403346954, we have a few options:

3a) We make an easy transformation: e.g. `SCS-2V:8:20s-GNa:64` to `SCS-2V_8_20s_GNa_64` (`s/:/_/g`) or `SCS-2V-8-20s_GNa-64` (`sed -e 's/-/_/g' -e 's/:/-/g' -e 's/^SCS_/SCS-/'`)   (2vCPU, 8Gib RAM, 20GB SSD, passthrough GPU nVidia Ampere with 64 SMs)
3b) We reevaluate our approach to systematic flavor naming, resulting in a new schema that can not be easily machine-translated.

The transition will be painful for existing providers and customers. In any case a *transition strategy* will need to be implemented, with transition periods, where the old flavor names will still be accepted (e.g. by being translated or by offering flavors under both new and old name).

The advantage of 3a is that we can easily machine-translate, so user settings/automation can be easily adjusted. We could even translate at the API interface to make things smoother.

garloff:
Notes: Just translating `:` with `-` would result in `SCS-2V-8-20s-GNa-64` which is harder to parse, as the 64SMs do belong to the GPU described before, but the syntax no longer reflects this. I would want to avoid this.

garloff:
I prefer the `sed -e 's/-/_/g' -e 's/:/-/g' -e 's/^SCS_/SCS-/'` appproach:
- Machine translatable and no need to restart the discussion on our (naming strategy)[]
- The simple flavors (which we will see in 95%+ of the cases) don't have underscores, which are somewhat ugly to my eyes.

Comments welcome!

garloff:
Decision taken by merging #226.

Issue #190, opened 2023-01-24
Flavor Naming Scheme: Use of colon character as delimiter


wwentland:
# Summary

The current [flavor naming standard][scs-flavor-naming] uses colon characters (i.e. `:`) as delimiters. This is an unfortunate choice, as it is incompatible with requirements stipulated by downstream applications such as Kubernetes.

# Details

Kubernetes uses [labels][k8s-labels] and [annotations][k8s-annotations] for a variety of use cases. Most importantly in the context of this discussion will be their use for [taints and tolerations][k8s-taints-tolerations] and the standard [well-known labels, annotations and taints][k8s-well-known-labels-annotations-taints].

The requirements for valid [label][k8s-label-charset] values are:

- must be 63 characters or less (can be empty),
- unless empty, must begin and end with an alphanumeric character ([a-z0-9A-Z]),
- could contain dashes (-), underscores (_), dots (.), and alphanumerics between.

That is, it must adhere to the regular expression `(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')`.

The usage of colon characters is not allowed and their usage in flavor names causes issues when, for example, the kubelet wants to set the well-known `node.kubernetes.io/instance-type` label:

```
Jan 01 08:09:10 foo-bar-box hyperkube[12345]: I0103 08:09:10.23456 14821 kubelet_node_status.go:426] "Adding node label from cloud provider" labelKey="node.kubernetes.io/instance-type" labelValue="SCS-4V:16:50"
[...]
Jan 01 08:09:10 foo-bar-box hyperkube[12345]: E0103 08:09:10.23456 14821 kubelet_node_status.go:94] "Unable to register node with API server" err="Node \"foo-bar-node\" is invalid: metadata.labels: Invalid value: \"SCS-4V:16:50\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue', or 'my_value', or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')" node="foo-bar-node"
```

Given the significance of Kubernetes in the SCS project, it would be advisable to reconsider the naming scheme and adopt one that is compatible with the data model defined there.

Thank you!

[scs-flavor-naming]: https://github.com/SovereignCloudStack/standards/blob/main/Design-Docs/flavor-naming.md
[...]

PR #226, opened 2023-02-17
Feat/new flavor scheme

garloff:
This is a first version how the new flavor naming v2 could look like to address issue #190, reflecting the considerations from #192.
A document that describes how the transition can be done smoothly still needs to be written.
The check tools which has been updated along with the standard can be run in a mode to accept both old v1 and new v2 flavor names, which hints how a transition might be implemented.
To check for v1 compliance, the tool has an option `-1` now, so we don't need to carry an old version of the test tool.


Tests/iaas/flavor-naming/flavor-name-check.py
```diff
@@ -27,11 +27,14 @@
 import sys
 import os
 import re
+# import getopt
```
fkr, line 30: unused, commented out import

Tests/iaas/flavor-naming/flavor-name-check.py
```diff
@@ -217,9 +251,10 @@ def out(self):
                     i += 3
                     lst.append(att)
                 par += 1
-            elif self.outstr[i+1] == ":":
+            elif self.outstr[i+1] == ":":   # change?
                 if att:
-                    ostr += ":%"
+                    # ostr += ":%"
```
fkr, line 256: imho remove, if left in commented out, imho leave comment why it is left there, so it clear why it was left there.

fkr (approved):
LGTM! left two very minor comments

Tests/iaas/flavor-naming/flavor-name-check.py
garloff: Fair comment, this has changed from v1 to v2. But git history is good enough if we ever want to do archeology.

Tests/iaas/flavor-naming/flavor-name-check.py
garloff: Removed.
Will readd in a subsequent patch, as the hand-written option parsing needs to go.

tibeer (approved):
LGTM! Just one hint: I have the feeling that the `flavor-name-check.py` script might become too complex to be easily maintained. That being said, I do not have any good idea on how to solve this.

wwentland (approved):
LGTM, thank you for addressing my concerns. Much appreciated!

fkr (approved):
LGTM

itrich (approved):
LGTM

PR #232, opened 2023-02-23
Add a GitHub Action to auto-generate verbose flavors spec

### LINKED ISSUE #216: Flavor standards file schema changed
@garloff I noticed during "finishing" the flavor manager, that the original agreed file structure diverged a lot. The initial file structure required values for the names to make the flavor manager as generic as possible: https://github.com/osism/openstack-flavor-manager/blob/main/flavors.yaml. Why was this changed to "only" the names?

itrich:
Fixes #216
Blocked by https://github.com/SovereignCloudStack/github-manager/pull/98

### DIFF
--- a/.github/workflows/create-flavors-spec.yml
+++ b/.github/workflows/create-flavors-spec.yml
@@ -1,0 +1,44 @@
+name: Generate Spec.MandatoryFlavors.verbose.yaml
+
+on:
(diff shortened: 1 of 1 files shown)


tibeer (approved):
Looks good! Let's test if it works

Issue #233, opened 2023-02-23
Re-evaluate types of existing records: "Decision Record" or "Standard"?


joshmue:
There are some "Decision Records" which looked like "Standards" to me.

Verify and open up a cleanup PR, if necessary.

joshmue:
#240 fixes this document's location only.
#241 fixes (IMHO) this document's type and location.
#242 adds a template in `Standards/` like it already exists for `Decisions`.

Having another look at things, I found that AFAIU, the current structure actually violates [SCS-0001](https://github.com/SovereignCloudStack/standards/blob/13b0df8f452ab565628a1abf677d76e91282e497/Standards/scs-0001-v1-sovereign-cloud-standards.md), which says:

> The pull request MUST add exactly one SCS document, in the Standards folder.

So, bringing up the open question of [SCS-0002](https://github.com/SovereignCloudStack/standards/blob/13b0df8f452ab565628a1abf677d76e91282e497/Decisions/scs-0002-v1-standards-docs-org.md) again...

> We don't know yet whether there is always a clear distinction between Standards and Decisions (ADRs) -- if this turns out too hard to decide, we may unify those two directories. (The decision will still be needed and be reflected in the header, but finding the docs will be easier.)

Question: Follow SCS-0001 by putting all files into a single directory or change SCS-0001 to reflect the de-facto presence of multiple directories?
If doing the former, all linked PR's I just created (except partly #241) can be just closed.

@garloff @horazont @itrich Opinions?

garloff:
Having them all in the Standards/ folder looks good to me.

joshmue:
Only `status-page-comparison` remains to be moved, but that is not scope of this issue. Hence, I'm closing it.

PR #241, opened 2023-02-28
Move scs-0110-v1-ssd-flavors.md to Standards

joshmue:
Partially addresses #233.

### DIFF
--- a/Standards/scs-0110-v1-ssd-flavors.md
+++ b/Standards/scs-0110-v1-ssd-flavors.md
@@ -1,6 +1,6 @@
 ---
 title: SSD Flavors
-type: Decision Record
+type: Standard
 status: Draft
 track: IaaS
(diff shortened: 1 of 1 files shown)


Standards/scs-0110-v1-ssd-flavors.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: SSD Flavors
-type: Decision Record
+type: Standard
 status: Draft
 track: IaaS
 enhances: scs-0100-v2-flavor-naming.md
```
fkr, line 6: Shouldn't this actually be 0101? (since we started with -0001- as well....

```suggestion
enhances: scs-0101-v2-flavor-naming.md
```

fkr (approved):
Your change looks good. I actually think the number should be adjusted....(but that has nothing to do with your change)

fkr (approved):

PR #243, opened 2023-03-01
Feat/document numbering scheme

fkr:


itrich:
I disagree with introducing a numbering scheme. We will break our own rules sooner or later, so I see no point in standardizing the document's numbering.

fkr:
yes, i think the same. however people already apply rules and use patterns- which i find very confusing as ***@***.*** - +49-172-1717311GPG/PGP: 824CE0F0 /  2082 651E 5104 F989 4D18  BB2E 0B26On 1. Mar 2023, at 18:34, Eduard Itrich ***@***.***> wrote:﻿
I disagree with introducing a numbering scheme. We will break our own rules sooner or later, so I see no point in standardizing the document's numbering.

fkr:
not needed anymore.

Standards/scs-0001-v2-sovereign-cloud-standards.md
```diff
@@ -76,6 +76,15 @@ The file name of an SCS document is formed using the following pattern:
...
 the resulting file name would be `scs-0390-v2-flavor-naming.md`.
+The second digit in `XXXX` describes the track where the document belongs:
+
+| Track | Number |
+| --- | --- |
+| IaaS | 1 |
+| KaaS | 2 |
+| Ops | 3 |
+| IAM | 4 |
+| Global | 5 |
```
garloff, line 87: Global is 0.

garloff (commented):
Do we want to assign a new number to SCS-0001?
I feel like we clarify something that we have been doing before, so it looks like an editorial change to me that does not necessarily require a new version of the standard.
(If we had minor versions, I would probably use a version 1.1 for this ...).

No strong opinion on this though.

PR #191, opened 2023-01-26
Feat/add status page design decision

fkr:

PR #245, opened 2023-03-02
this document is not needed anymore since it is in the status page adr

fkr:
see https://github.com/SovereignCloudStack/standards/pull/191


itrich (approved):
LGTM.

fkr:
What I actually just wondered: rm'ing this file (because it became part of the ADR) will cause links from issues to point to nowhere. Since those are issues that are closed (and as such, of no relevance any longer), I'm in the 'this is not a problem and I much rather have clean repos without cruft'-camp - but since we just had this discussion, I want to throw this up in the air.

@itrich @maxwolfs @garloff

joshmue:
> If i am understanding your issue correctly you can simply adjust the links within those issues referencing the file from the specific commit at that time for correct history. Just choose the commit and browse to the file for copying the specific url.

Which of course can also be done at the time of writing an issue/comment. Not using permalinks to code/markdown files IMHO always risks to have broken links after some time (ignoring possibility of repository deletion etc.).

+1 for merging this PR, "this is not a problem and I much rather have clean repos without cruft" at least in this case.

fkr:
paging @joshmue :) please review

joshmue (approved):
LGTM

PR #248, opened 2023-03-02
move to standards directory - this way all standards and decision doc…

fkr:
…uments are in one folder

discussed with @maxwolfs and @itrich during the hackathon


itrich (approved):
LGTM!

PR #205, opened 2023-02-04
Add ADR proposal for "requirements for SSO identity federation"

reqa:
WIP, first draft, some iterations expected for this MR.
Purpose of MR: collecting feedback for the ADR proposal.

Signed-off-by: Arvid Requate <requate@univention.de>

PR #249, opened 2023-03-02
rename to just have one incrementing number

fkr:
discussed with @joshmue, @itrich and @maxwolfs during the hackathon.


garloff:
I preferred the other number scheme, but that is not a strong opinion at all, so I'd not at all fight for it if your preference is clear.

The only trouble is that a rename tends to breaks lots of links, which is really bad. This includes our own real estate (web page, scs-compliant.yaml), but also slide decks (like the one used at the Alsaca talk) and external web pages. We have had quite some disturbance with the rename of the old Docs repo and the internal reorg (scs-0002). So this would be yet another one. **We should have a clear vision how we stop breaking links left and right.**

As there is no problem in IT that can't be resolved by another level of indirection:
What is our plan to provide stable links in the future? Do we need an indirection service and whenever renames happen do update the remapping tables? (This could also provide some side benefits like e.g. providing links to scs-XXXX-vlatest.)

Topic for SIG standardization? Or more generically our information architect (SIG documentation) and community manager (SIG community)? Thus product board? Any insight on best practices from others? @horazont, @jschoone?

itrich:
I'd propose to have a static link service. We can either host our own open-source solution or realize this with Jekyll as we already did for release notes (e.g. <https://scs.community/release-notes-r0/>).

fkr:
@itrich indeed, that would also be my suggestion. From my perspective we should try to keep things simple in order to be faster and things like a simple numbering scheme, ability to work with static links and yet still be able to adjust locations and filename is crucial since otherwise we spend an enormous amount of time trying to get things perfect at first shot.

fkr:
ping @itrich @joshmue @garloff @maxwolfs 
this affects subsequent PRs (like https://github.com/SovereignCloudStack/standards/pull/205) so it would be good to get this either in **or** rejected so we know how to deal with the numbering.

joshmue:
The other scheme (as attempted to document in #243) has IMHO the following drawbacks:
- It is not intuitively clear, that certain digits in an apparent integer mean different things
- It is not intuitively clear, what digits stand for
    - Why the second digit for the track?
    - Why `4` for `IAM`?
    - The first digit is unused, if it would be used in the future, standards would probably need to be renamed
- There are potential scalability issues: No more than 99 can be created for each track

Alternatively, the track could also be encoded as a number of letters like `iaas`, `kaas`, `ops`, `iam`, `global` in the file name (or e. g. `iaas`, `kaas`, `oper`, `iden`, `glob` to have a fixed length). But this information is encoded in the record anyway, so this should not be necessary.

I think treating the number as just a simple number is much more easy to understand.

joshmue (approved):

garloff:
Please provide stable links, so we only have to do all the link fixing once.

fkr:
We should then consider re-opening https://github.com/SovereignCloudStack/standards/pull/243 so that the scheme is documented..

PR #252, opened 2023-03-15
Add: IaaS ADR on providing virtio-rng entropy source

matfechner:
from this Disscussion

https://github.com/SovereignCloudStack/docs/pull/24

Closes SovereignCloudStack/issues/issue#234

### DIFF
--- a/Standards/scs-0100-v1-entropy.md
+++ b/Standards/scs-0100-v1-entropy.md
@@ -1,0 +1,86 @@
+---
+title: ADR SCS Entropy
+type: Standard
+status: Draft
+track: IaaS
+enhances: scs-0100-v1-entropy.md
+---
+
+## 1. Terminology of Entropy
+
+### 1.1. Entropy in information Technology
+
+Entropy is a concept that is widely used in the scope of information
+technology. It is a measurement of the amount of disorder or randomness in
+a system. Entropy is used to measure the amount of information in a
+self contained systems, as well as the amount of incertitude that exists
+in this system. For cryptographic procedures and operations good entropy
+is a must have!
+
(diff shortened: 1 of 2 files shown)


frosty-geek:
Is this a standard or an ADR? I'm confused by the title...
Further more if this is a standard all CSPs have to follow, do we have to change the mandatory fields for all images?

garloff:
> Is this a standard or an ADR? I'm confused by the title...

This will become a compliance requirement, so it's a standard.

> Further more if this is a standard all CSPs have to follow,

By merging this, it becomes a merged draft standard which then goes into stabilization.
As soon as it is stabilized, it becomes *eligible* to be chosen as a certification requirement for IaaS SCS-compatible certification. If we elect to do so, we would incorporate it into `Tests/scs-compatible.yaml`.

There is no automatism, but yes: we certainly intend to make it a mandatory standard.
 
> do we have to change the mandatory fields for all images?

Not if you already comply to the mandatory image standard:
https://github.com/SovereignCloudStack/standards/blob/main/Drafts/Image-Properties-Spec.md#technical-requirements-and-features
It already requires the `hw_rng_model` image property. (No, this is not a coïncidence.)

The only new requirement here is the flavor extra specs fields.
The `hw_rng:Allowed=true`  setting should already be active in your flavors.
I would definitely recommend that we mandate it, as we want to guarantee good entropy for VMs on SCS clouds.

@matfechner: Are there technical reasons to require `hw_rmg:rate_bytes` and `rate_period`, or are there reasonable defaults, so we could make these optional? I have not seen these two much ...

garloff (approved):
Works for me now.

PR #254, opened 2023-03-21
Add markdown linter and link validator

maxwolfs:
In order to implement the standards and their lifecycle state to <https://docs.scs.community>, it's necessary that all markdown files are linted correctly and in the same way. 

This PR:

* adds the same link validator action and markdownlint with ruleset as in the doccs repository
* removes old markdownlint action

Additionally to linting all markdown files I have added the missing frontmatter prop `title` to some of the standards.


Standards/scs-0100-v2-flavor-naming.md
itrich, line 1: It's kind of ironic and funny, that the linter is starting to mess around with our flavor namings :rofl:

PR #259, opened 2023-03-29
(see #168) add draft of a document that will describe what image will be auto crawled in which way

master-caster:
Hi *,

as of issue [#168](https://github.com/SovereignCloudStack/issues/issues/168)
we would like to provide some basic images. This is a first draft that describes
the necessity and provides a first list of links that contain metadata that
may be used to keep the images up to date


fkr (commented):
I think, we can discard this as it has been superseeded by the other work in this area.

PR #260, opened 2023-03-29
Fix/accept v1 mandatory flavors

garloff:


Tests/iaas/flavor-naming/flavor-name-check.py
```diff
@@ -700,9 +713,14 @@ def main(argv):
...
         namecheck = outname(*ret)
-        if completecheck and name in scsMandatory:
-            scsMandatory.remove(name)
-
+        if completecheck:
+            if name in scsMandatory:
+                scsMandatory.remove(name)
+            elif accept_old_mand:
+                newnm = old_to_new(name)
+                # print(f"DEBUG: {name} -> {newnm} in list: {newnm in scsMandatory}")
```
fkr, line 721: not really a big fan of debug code that is left - even with commented out. but that is personal taste.

fkr (approved):
LGTM. The only comment has nothing to do with the functional additions. thanks @garloff!

Tests/scs-compatible.yaml
```diff
@@ -27,6 +27,7 @@ iaas:
         url: https://raw.githubusercontent.com/SovereignCloudStack/standards/main/Standards/scs-0100-v2-flavor-naming.md
         check_tools:
           - executable: ./iaas/flavor-naming/flavor-names-openstack.py
+            args: "--accept-old-mand"
```
garloff, line 30: We need to revert this line as we have chosen another approach: v2 compliance does mean that all mandatory v2 flavor names need to be present and we have instead chosen to expose both v1 and v2 testing results. As this is a (minor) incompatibility b/w v1 and v2 standards, better make this very visible.

PR #262, opened 2023-03-31
Adjust table to test public clouds against v1 and v2

itrich:
The final table can be seen here: https://github.com/SovereignCloudStack/standards/tree/feat/adjust-compliance-table#scs-compatible-clouds

The badges have been created with [shields.io](https://shields.io/category/build), which allows nicer and more flexible badges.

2023-03-31 20_21_49-SovereignCloudStack_standards at feat_adjust-compliance-table


maxwolfs (approved):
lgtm

PR #287, opened 2023-05-17
add ISSUE templates

fkr:
suggested by @mbuechse

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,10 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: "[BUG]"
+labels: bug
+assignees: ''
+
+---
+
+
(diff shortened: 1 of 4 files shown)


mbuechse:
Not sure if this is the right place to discuss the following: @fkr explained to me his scheme of adding labels to issues (one for the outcome, one for the track, yet others still that I currently don't recall, and lastly maybe "needs refinement"). I like this scheme, and I guess there might even be documentation that explains this already, but maybe we can even add a line about this to the templates, so whoever opens an issue is confronted with this information directly and can't miss it?

fkr:
@mbuechse good suggestion. Am wondering wether really adding it to the templates helps or wether it makes sense, if we publish a blogpost and add some documentation on "how we work". 🤔

mbuechse (approved):
:100:

fkr:
I opened an issue to follow up on this: https://github.com/SovereignCloudStack/issues/issues/332

PR #291, opened 2023-05-25
Image metadata is an approved std since long. Move.

garloff:
The standard existed before the standardization process itself was described and the directory structure was fixed. This resulted in it living in the Drafts directory, which was misleading. Move it to standards and adjust the header fields.


fkr (approved):
LGTM.
If this is merged, this PR needs to be merged as well: https://github.com/SovereignCloudStack/website/pull/661

PR #292, opened 2023-05-25
Refactor the flavor naming document to make it more readable.

garloff:
Main change is to move the complex and rarely used extensions to the bottom of the document and add a section that explains why these exist at all and how they may or may not be used.

The change does not change the specification in any way, thus no new version number is assigned.

This change is in preparation of two decsisions and PRs that will be created:
(1) Reduce the mandated flavors by making the flavors with disks
    only recommended (as discussed on May 8).
(2) Discuss an alternative approach to flavor naming from PS.

### DIFF
--- a/Standards/scs-0100-v2-flavor-naming.md
+++ b/Standards/scs-0100-v2-flavor-naming.md
@@ -68,10 +68,9 @@ encoding all details) as well as very detailed longer names.
 
 ## Complete Proposal for systematic flavor naming
 
-| Prefix | CPUs & Suffix    | RAM[GiB]        | opt: Disk[GB]&type      | opt: extensions |
-| ------ | ---------------- | --------------- | ----------------------- | ----------------|
-| `SCS-` | N`L/V/T/C`[`i`]` | `-`N[`u`][`o`]` | [`-`[M`x`]N[`n/s/l/p`]] | [`_`EXT]        |
-
+| Prefix | CPUs & Suffix      | RAM[GiB]            | opt: Disk[GB]&type            | opt: extensions |
+| ------ | ------------------ | ------------------- | ----------------------------- | ----------------|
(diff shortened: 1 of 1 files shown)


Standards/scs-0100-v2-flavor-naming.md
```diff
@@ -68,20 +68,28 @@ encoding all details) as well as very detailed longer names.
 
 ## Complete Proposal for systematic flavor naming
 
-| Prefix | CPU | Suffix       | RAM[GiB]   | optional: Disk[GB] | optional: Disk type | optional: extra features                            |
-| ------ | --- | ------------ | ---------- | ------------------ | ------------------- | --------------------------------------------------- |
-| `SCS-` | `N` | `L/V/T/C[i]` | `-N[u][o]` | `[-[Mx]N]`         | `[n/s/l/p]`         | `[_hyp][_hwv][_[arch[N][h][_[G/g]X[N][-M[h]]][_ib]` |
+| Prefix | CPUs & Suffix    | RAM[GiB]        | opt: Disk[GB]&type      | opt: extensions |
+| ------ | ---------------- | --------------- | ----------------------- | ----------------|
+| `SCS-` | N`L/V/T/C`[`i`]` | `-`N[`u`][`o`]` | [`-`[M`x`]N[`n/s/l/p`]] | [`_`EXT]        |
```
tibeer, line 73: The backticks are slightly broken, is this intended?

tibeer (approved):
See comment, rest looks good!

Standards/scs-0100-v2-flavor-naming.md
garloff: Actually, this was intentional. The things that get into the string literally are supposed to be within backticks (i.e. the `SCS-` prefix or the letters indicating the vCPU type), but no longer the brackets [] that indicate something is optional nor the N or M letters that are placeholders for numbers.

PR #293, opened 2023-05-26
Call doc v2.1. Backticks were still wrong at one place.

garloff:
Also improve wording at one place.


garloff:
Note to reviewers:
* Tim: You were right: I did not have all the backticks correct yet, sigh.
* Matthias: In case you wonder: We don't reflect the minor version number in the document file name.
  This is OK, as the minor version changes can only improve the wording but are not allowed to introduce any more requirements (nor drop any).

mbuechse (approved):

Issue #271, opened 2023-04-27
SCS IaaS Standard flavors and Flavor naming discussion


garloff:
The SCS community had agreed on a standard flavor naming scheme and standard flavors back in 2021. The discussion was somewhat lengthy. In the process a document explaining the [logic](https://github.com/SovereignCloudStack/standards/blob/main/Drafts/flavor-naming-strategy.md) and a [v1 flavor naming specification](https://github.com/SovereignCloudStack/standards/blob/main/Drafts/flavor-naming.md) were agreed upon.
This was considered to be a mandatory SCS standard on the IaaS layer since then.
The software engineering rule that something that is not tested is (almost) always broken also hit us here: In summer 2022, we observed SCS cloud providers not implementing the standard fully correctly. Since then a [checker tool](https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/flavor-naming/flavor-name-check.py) has been created that can validate a flavor or generate a flavor name from parameters.
A tool to [check all flavors](https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/flavor-naming/flavor-names-openstack.py) has been added and it is called from the [compliance check test suite](https://github.com/SovereignCloudStack/standards/blob/main/Tests/scs-compliance-check.py) whose results are displayed on [github](https://github.com/SovereignCloudStack/standards). Our CSP partners were successful in becoming fully compliant in fall 2022.

Unfortunately, the original v1 proposal contains `:` in their names and certain k8s tools (OpenShift) use the infra flavor names as labels in Kubernetes, where `:` is not allowed. This [causes trouble](https://github.com/SovereignCloudStack/standards/issues/190) running OpenShift on top of SCS IaaS with v1 flavor names. A [v2 scheme](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0100-v2-flavor-naming.md) was [discussed](https://github.com/SovereignCloudStack/standards/issues/192) and created.

Unfortunately, the migration causes some [pain](https://github.com/SovereignCloudStack/standards/issues/228) for providers. In the Nuremberg Hackathon, we found out that not all SCS IaaS providers have a clear commitment to implement the v2 flavors and work with us to minimize the pain. This would be unfortunate, stripping SCS from one of the benefits where flavor related pain points for migrations and/or simultaneous multi-cloud deployments are addressed.

[...]

PR #294, opened 2023-05-26
Feat/flavor v3 make disks recommended

garloff:
Notes on this PR:
* It consists of several steps (commits) that can be reviewed individually
  - Copy old v2 standard, add distinction b/w mandatory and recommended in text and in test tools (`--v3` aka `-3` option), add SSD flavors from 0110-v1 as mandatory.
* The last step could in theory also add the SSD flavors as recommended only
* In `--v3` mode, missing recommended flavors will create a warning, but no compliance error
* The test tools currently work with just one version of the required flavor file. This is currently broken by not dropping the SSD flavors from v2 and v1 tests -- this may need a different approach. (Maybe I would continue to use one version of the file with the required flavors, but annotate the new SSD ones, or create two versions or just hardcode dropping them without `--v3`. That latter is a bit hacky, admittedly.)
* wavestack and regio.digital both currently pass the v3 tests already (SSD flavors exist), ps fails by not having the new v2 names yet nor the SSD flavors.
* we would still have to decide when this standard becomes valid and when v1 and v2 are phased out.


garloff:
This implements Ralfs suggestions from https://github.com/SovereignCloudStack/standards/issues/271#issuecomment-1538371315

Tests/iaas/flavor-naming/flavor-name-check.py
```diff
@@ -629,11 +629,14 @@ def readmandflavors(fnm):
                 break
     with open(fnm, "r", encoding="UTF-8)") as fobj:
         yamldict = yaml.safe_load(fobj)
-    ydict = yamldict["SCS-Spec"]["MandatoryFlavors"]
+    man_ydict = yamldict["SCS-Spec"]["MandatoryFlavors"]
```
mbuechse, line 632: The `yamldict` is changed in place, so we could as well proceed as follows (modulo typos):

```
if prefer_old:
    for name_list in yamldict["SCS-Spec"].values():
        for i, name in enumerate(name_list):
            name_list[i] = new_to_old(name)
return yamldict["SCS-Spec"]["MandatoryFlavors"], yamldict["SCS-Spec"]["RecommendedFlavors"]
```

Then it's more obvious that the change is done in place, we get rid of the duplicated code, and the return statement is more telling (in my opinion).

Tests/iaas/flavor-naming/flavor-names-openstack.py
```diff
@@ -82,7 +87,18 @@ def main(argv):
...
+    scsMandatory, scsRecommended = fnmck.readmandflavors(scsMandFile)
+    if not v3mode:
+        scsMandatory = [*scsMandatory, *scsRecommended]
+        scsRecommended = []
+        # HACK: Drop SSD flavors from list
+        scsV2Mand = []
+        for flv in scsMandatory:
+            ret = fnmck.parsename(flv)
+            disk = ret[1]
+            if not disk or "disktype" not in disk.__dict__ or disk.disktype != "s":
```
mbuechse, line 99: This is `__dict__` check is quite hard to swallow for my taste. Could we instead use `not disk.parsed`? Would that do the trick as well?

mbuechse:
```suggestion
            if not disk or not disk.parsed or disk.disktype != "s":
```

mbuechse: On second thought, I think the list `scsV2Mand` should be constructed via list comprehension, and then the variable could be dropped for good. Something like so:

```python
def is_ssd_flavor(name):
    disk = fnmck.parsename(flv)[1]
    return disk and disk.parsed and disk.disktype == "s"

scsMandatory = [name for name in scsMandatory if not is_ssd_flavor(name)]
```

Tests/iaas/flavor-naming/flavor-names-openstack.py
```diff
@@ -187,28 +211,42 @@ def main(argv):
     # Produce dicts for YAML reporting
     flvSCSList = {
         "MandatoryFlavorsPresent": MSCSFlv,
-        "MandatoryFlavorsMissing": scsMandatory,
+        "MandatoryFlavorsMissing": scsMandatory
```
mbuechse, line 214: I would strongly advocate for trailing commas here and below!

Tests/scs-compatible.yaml
```diff
@@ -36,3 +36,21 @@ iaas:
         url: https://opendev.org/openinfra/interop/src/branch/master/guidelines/2022.11.json
         condition: mandatory
         # Unfortunately, no wrapper to run refstack yet, needs to be added
+  - version: v3
```
mbuechse, line 39: I'm still puzzled by there being two yaml files. This PR only changes one of them. Why is that (I just don't know)?

mbuechse (commented):
I have added a few remarks, nothing critical as far as I can tell.

Tests/iaas/flavor-naming/flavor-manager-input.py
```diff
@@ -102,22 +105,38 @@ def main(argv):
 
     outspec = SpecSyntax.spec_dict()
     if list_mode:
-        scs_mand_flavors = args
-        flavor_type = "optional"
+        scs_mand_flavors = []
+        scs_rec_flavors = args
+        # flavor_type = "optional"
```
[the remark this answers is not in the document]
fkr: I agree with @mbuechse and would suggest to remove it.

Tests/iaas/flavor-naming/flavor-manager-input.py
```diff
@@ -102,22 +105,38 @@ def main(argv):
...
     else:
-        scs_mand_flavors = fnmck.readmandflavors(scs_mand_file)
-        flavor_type = "mandatory"
+        scs_mand_flavors, scs_rec_flavors = fnmck.readmandflavors(scs_mand_file)
+        # flavor_type = "mandatory"
```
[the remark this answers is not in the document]
fkr: I agree with @mbuechse and would suggest to remove it.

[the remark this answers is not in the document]
garloff: Done.

Tests/iaas/flavor-naming/flavor-name-check.py
```diff
@@ -616,7 +616,7 @@ def old_to_new(nm):
 
 
 def readmandflavors(fnm):
-    "Read mandatory flavors from passed YAML file, search in a few paths"
+    "Read mandatory (and recommended) flavors from passed YAML file, search in a few paths"
```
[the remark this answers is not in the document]
garloff: Addressed in https://github.com/SovereignCloudStack/standards/pull/294/commits/5f0167b00dedc5b2df94c87784e0bbc863c71cd4.

Tests/iaas/flavor-naming/flavor-manager-input.py
```diff
@@ -102,22 +105,38 @@ def main(argv):
...
-    outspec[flavor_type] = olist
+    rec_list = []
+    for name in scs_rec_flavors:
+        try:
+            ret = fnmck.parsename(name)
+            assert ret
+            rec_list.append(SpecSyntax.mand_dict(name, ret))
+        except NameError as exc:
+            print(f"{exc}", file=sys.stderr)
+            errors += 1
```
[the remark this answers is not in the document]
garloff: Addressed in https://github.com/SovereignCloudStack/standards/pull/294/commits/07bd711b0dc36a9841e2545f64a06c3783b6a330 with helper function.

Tests/iaas/flavor-naming/flavor-names-openstack.py
garloff: List comprehensions are definitely more readable, thanks.
Implemented in https://github.com/SovereignCloudStack/standards/pull/294/commits/642aa69165fc2712e0e807efc2e7d8f770454027

Tests/iaas/flavor-naming/flavor-name-check.py
garloff: Implemented in https://github.com/SovereignCloudStack/standards/pull/294/commits/02044a8b183a92de04df03d30dfee12ab7fa4cb7

Tests/iaas/flavor-naming/flavor-names-openstack.py
garloff: No strong opinion. I tend to think it's a good habit to have these commas, so I just did it.
Implemented in https://github.com/SovereignCloudStack/standards/pull/294/commits/84fde967cbb3119ab87667a7bb5569b39b0d8269

garloff:
I intend to redo the distinction between v3 flavors by changing the format of  `SCS-Spec.MandatoryFlavors.yaml`
to have sections `SCS-Spec.MandatoryFlavors`, `SCS-Spec.MandatoryFlavorsV3`, `SCS-Spec.RecommendedFlavors` and `SCS-Spec.RecommendedFlavorsV3` sections and handle the filtering in the loader, which get's a version number passed.
After that, I intend to remove the "work in progress" label and ask for more review.
Thanks for all the good input, @mbuechse.

Tests/scs-compatible.yaml
garloff: Moved into a testing directory with https://github.com/SovereignCloudStack/standards/pull/294/commits/e5c0d811b07d76b2941da650267dc0f4944312d6.

mbuechse:
> After that, I intend to remove the "work in progress" label and ask for more review.

Oh, I totally missed that label.

> I intend to redo the distinction between v3 flavors by changing the format of SCS-Spec.MandatoryFlavors.yaml

Now of course I'm wondering whether there is a more general (if not to say elegant) solution... You know, I did criticize some code duplication, but I'm not entirely against duplication if it serves a purpose. In that case, I can imagine using one file per version, even if that means some duplication, but then the code stays quite simple, and everything is very explicit.

mbuechse:
Addendum: It would actually nicely mirror the situation with the standard texts (the markdown files in this repo), where we have one file per major version, regardless of duplication.

garloff:
> > I intend to redo the distinction between v3 flavors by changing the format of SCS-Spec.MandatoryFlavors.yaml
> 
> Now of course I'm wondering whether there is a more general (if not to say elegant) solution... You know, I did criticize some code duplication, but I'm not entirely against duplication if it serves a purpose. In that case, I can imagine using one file per version, even if that means some duplication, but then the code stays quite simple, and everything is very explicit.

No, duplicating complex code is a bad idea. And the name parser is not trivial ...
One set of tools that can handle all versions (as long as we stick to a scheme that resembles what we currently have in v1/v2/v3).
We can discuss whether the `SCS-Spec.MandatoryFlavors.yaml` should be versioned. That would be harmless duplication.
I have decided against it for now (and instead have introduced `MandatoryFlavorsV3` list), but that's a discussion we can have and I'm open to be convinced differently. But no, the parser should be able to handle all versions IMVHO.

garloff:
This is ready for review IMVHO.
I'd like to get this merged as Draft quickly.
So we have a solid proposal that is in line with current flavor name handling in SCS (v1, v2).

We may or may not stabilize it, depending on how the discussion goes w.r.t to the alternative proposal from @PSwatchmen at https://github.com/SovereignCloudStack/standards/issues/271#issuecomment-1536315678.

garloff:
Sidenote: The MD link checker fails, as the linked file will only exist once the merge has been done.
So this failure can be ignored in judging the maturity of the PR, IMVHO.
Another sidenote: I have manually set DCO to pass -- the reason is that one commit from myself missed the needed signoff. As the commit history will be squashed and I know that I contributed this willingly and with all needed permissions, I can take care of it when the squashing happens. Thus no blocker either.

mbuechse:
> No, duplicating complex code is a bad idea. And the name parser is not trivial ...

Just for the record (for posterity): We are in agreement here. My aim was to duplicate the yaml, not the Python file. Unfortunately, I didn't write sufficiently clearly.

Tests/iaas/flavor-naming/flavor-name-check.py
```diff
@@ -627,13 +639,38 @@ def readmandflavors(fnm):
...
+            for i, name in enumerate(name_type):
+                name_type[i] = new_to_old(name)
+    mand = yamldict["SCS-Spec"]["MandatoryFlavors"]
+    recd = yamldict["SCS-Spec"]["RecommendedFlavors"]
+    if v3:
+        try:
+            mandv3 = yamldict["SCS-Spec"]["MandatoryFlavorsV3"]
+            mand.extend(mandv3)
+        except KeyError:
+            pass
```
mbuechse, line 665:
```suggestion
        mand.extend(yamldict["SCS-Spec"].get("MandatoryFlavorsV3", ()))
```

Tests/iaas/flavor-naming/flavor-name-check.py
```diff
@@ -627,13 +639,38 @@ def readmandflavors(fnm):
...
+        try:
+            recdv3 = yamldict["SCS-Spec"]["RecommendedFlavorsV3"]
+            recd.extend(recdv3)
+        except KeyError:
+            pass
```
mbuechse, line 670:
```suggestion
        recd.extend(yamldict["SCS-Spec"].get("RecommendedFlavorsV3", ()))
```

Tests/iaas/flavor-naming/flavor-name-check.py
```diff
@@ -665,26 +704,31 @@ def main(argv):
...
+        scsMandatory, scsRecommended = readflavors(argv[1], v3_flv, prefer_old)
         scsMandNum = len(scsMandatory)
+        scsRecNum = len(scsRecommended)
         if debug:
             print(f"Check for completeness ({scsMandNum}): {scsMandatory}")
         argv = argv[2:]
 
     # Interactive input of flavor
-    if argv[0] == "-i":
+    if len(argv) and argv[0] == "-i":
```
mbuechse, line 731: Just a bit more "pythonic":
```suggestion
    if argv and argv[0] == "-i":
```

tibeer (approved):
LGTM

tibeer:
@garloff openstack flavor manager has been altered: https://github.com/osism/openstack-flavor-manager/pull/11

garloff:
Great, thanks Tim

PR #299, opened 2023-06-02
Remove duplicate status page comparison

joshmue:
The comparison is present in the `Standards/scs-0400-v1-status-page-create-decision.md` file already, no need to keep a duplicate version at a deprecated path.


joshmue:
Duplicate of #245 .

PR #302, opened 2023-06-07
Propose to stabilize scs-0300 (SSO IdP criteria)

reqa:
* Propose to stabilize scs-0300 by 2023-06-21.


fkr (approved):
imho fine to stabilize

PR #304, opened 2023-06-07
Stabilize the SSD flavor decision

mbuechse:
As per the team IaaS agreement from the [meeting on 2023-06-07](https://github.com/SovereignCloudStack/minutes/blob/main/iaas/20230607.md#stabilize-the-ssd-flavor-decision-199).

The new flavors have already been added to v3 (draft) of the flavor naming standard: https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0100-v3-flavor-naming.md

Makes https://github.com/SovereignCloudStack/standards/pull/199 obsolete.


mbuechse:
Not sure what to do about the failing markdown-lint check, as the file it chokes on has nothing to do with this PR.

frosty-geek (commented):
Is this a decision record or a standard?

mbuechse:
@frosty-geek It's a decision record that documents the decision to change the standard [scs-0100-flavor-naming](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0100-v3-flavor-naming.md), introducing a new version, namely v3.

Decisions/scs-0110-v1-ssd-flavors.md
```diff
@@ -1,7 +1,8 @@
 ---
 title: SSD Flavors
-type: Standard
-status: Draft
+type: Decision Record
```
garloff: Correct, this is not a full-fledged standard on its own right.
It is a decision, documenting reasoning and outcomes that then informs the flavor-naming standard.
However, since the last consolidation, both Standards and Decision Records live in the Standards directory.
The reason is that these are not always so clearly distinguishable and decisions tend to inform standards -- so keeping them together (and also have unique naming/numbering) makes it easier to keep an overview.

I have addressed this with commit https://github.com/SovereignCloudStack/standards/pull/304/commits/217d07641774f9af7af68f8b68edc967a82ee1ea

garloff (commented):
With the change I made this looks OK to me.
Will approve this by 2023-06-14, but want to give others the chance to review and speak up.

Note: The stabilization of this makes it possible to be consistent when stabilizing scs-0100-v3-flavor-naming, which references this decision or the requirements of local SSDs (more than 1000 sequential IOPS and power-loss protection).

garloff (approved):
Approve as per decision in team IAM 2023-06-14

PR #305, opened 2023-06-11
Renumber metering 0110/0111 to 0410/0411.

garloff:
This put it into the Ops track -- the metering approach is not restricted to the IaaS layer, even if we implement it there first.

This also avoids the conflict with the previously existent scs-0110-v1-ssd-flavors.
(Note that we don't want to have double numbers, not even between decisions and standards.)


fkr:
I don't comprehend the numbering.
According to the table in https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0001-v1-sovereign-cloud-standards.md this change would move it to the OPS track...

horazont (approved):

PR #306, opened 2023-06-11
Feat/stabilize flavors v3

garloff:
Proposal to stabilize v3 flavor naming and immediately start using it for compliance testing with a new v3 compliance standard, basically replacing v2 on the spot.
Reasoning:
- CSPs fulfilling v2 also fulfill v3 (by having the SSD flavors)
- The CSP not yet fulfilling v2 will likely prefer jumping to v3 directly (as it reduces the mandatory flavors)
- v2 flavor-naming was referenced in IaaS v2 tests, yet we had not formally stabilized it. Ouch - this should not have happened.

Note: We have conducted a longer discussion whether the v1 -> v2/3 switch should be used to change our approach to flavor naming and we had an alternative proposal, see https://github.com/SovereignCloudStack/standards/issues/271.

In the end, only one of our partners favored the alternative proposal. Several pointed out that they prefer not to have significant change, as it would weaken our standards. Some said that they really like that CPUs and RAM (and root disk if present) can be trivially seen from the name. So we decided that we go with the v3 names.

Note2: After long discussions, I'd like to close this quickly, thus the short timeline. If there is a need for more discussions, please do speak up, we can certainly extend this by a week IF there is a need.

Note 3: As IaaS v3 will effectively become the successor of IaaS v1 compliance (hiding the intermediate v2 somewhat), this means that -- if we don't change the v1 obsoletion, which is currently set for 2023-10-31, this means that by the end of Oct, we need all providers to implement the new v2/v3 flavor names AND have the two flavors with local SSDs to be compliant with SCS IaaS compliance. PLEASE speak up if this is a problem. (We could address this with different timelines or -- if really needed - by making the SSD flavors only recommended. I'd hate the latter, but I hate to lose a CSP partner even more.)


garloff:
One additional note:
I expect PR #304 to land first, which is good.

tibeer (approved):

fkr (approved):

PR #314, opened 2023-06-21
Standardize image names and source URLs (resolves #326)

### LINKED ISSUE #326: Standardize image names and source URLs
We want users to be able to upload their own images to Glance, so we mandate that this must be possible (with some quota).

We want to standardize image names and source URLs with the intention that the CSP can provide any subset of these images at will; AND if an image with a name from that list is provided, it has to be from the given source URL. The list should be given by a yaml file, whose schema could be a subset of the one for the image-manager yaml files.

The only mandatory image should be the kubernetes-capi images. These could be listed in the yaml file as well and marked mandatory.

Work on this issue can be done inside https://github.com/SovereignCloudStack/standards/pull/314 -- at least the applicable parts from that PR should be reused.

Definition of done:
- [x] A new standard is written that covers image upload and standard names/URLs (given via a yaml file that is regarded as an annex).
- [x] The yaml file lists the mandatory kubernetes-capi images.
- [x] From the existing image metadata standard the section "Standard images" is removed.
- [x] A test script has been written for the new standard, and it has been added to <https://github.com/SovereignCloudStack/standards/blob/issue/267/Tests/testing/scs-compatible-test.yaml>.
- [x] The new standard has been added to https://github.com/SovereignCloudStack/standards/issues/317

mbuechse:
Resolves #326


Standards/scs-0104-v1-standard-images.md
```diff
@@ -0,0 +1,45 @@
...
+
+The following images are mandatory:
+
+- `Ubuntu <LATESTLTS>`, `Ubuntu <PREVLTS>`, `Debian <STABLE>`
+- Note that `<LATESTLTS>` refers to the latest LTS version, which at this point is `22.04`.
+  The `<PREVLTS>` is the previous LTS version, at the time of this writing (9/2022) it's `20.04`.
+  We don't carry the `.x` patch numbers in the standard image names. We switch to requiring the
+  newest Ubuntu LTS version when the `.1` version comes out (around July/August). At this point
+  the old `<PREVLTS>` version becomes optional ...
+- For Debian, we use the latest STABLE version, which is `11` at the time of this writing.
```
fkr, line 18:
```suggestion
- For Debian, we use the latest STABLE version, which is `12` at the time of this writing.
```

Standards/scs-0104-v1-standard-images.md
```diff
@@ -0,0 +1,45 @@
...
+track: IaaS
+---
+
+## Standard images
+
+The following images are mandatory:
+
+- `Ubuntu <LATESTLTS>`, `Ubuntu <PREVLTS>`, `Debian <STABLE>`
+- Note that `<LATESTLTS>` refers to the latest LTS version, which at this point is `22.04`.
+  The `<PREVLTS>` is the previous LTS version, at the time of this writing (9/2022) it's `20.04`.
```
fkr, line 14:
```suggestion
  The `<PREVLTS>` is the previous LTS version, at the time of this writing (7/2023) it's `20.04`.
```

Standards/scs-0104-v1-standard-images.md
```diff
@@ -0,0 +1,45 @@
...
+  We don't carry the `.x` patch numbers in the standard image names. We switch to requiring the
+  newest Ubuntu LTS version when the `.1` version comes out (around July/August). At this point
+  the old `<PREVLTS>` version becomes optional ...
+- For Debian, we use the latest STABLE version, which is `11` at the time of this writing.
+  Similar to Ubuntu, we would do the switch and require the latest STABLE to be made available within
+  ~3 months after release.
+- When a CentOS successor emerges, we would have one in the mandatory list.
```
fkr, line 21:
```suggestion
```

Standards/scs-0104-v1-standard-images.md
```diff
@@ -0,0 +1,45 @@
...
+
+The following images are recommended:
+
+- `Alma Linux 8`, `Alma Linux 9` (whichever versions are being supported upstream)
```
fkr, line 25:
```suggestion
```

Standards/scs-0104-v1-standard-images.md
```diff
@@ -0,0 +1,45 @@
...
+- `SLES 15SP4`
+- `RHEL 9`, `RHEL 8`
+- `Windows Server 2022`, `Windows Server 2019`
+
+We are also looking into standard suggestions for
+
+- `openSUSE Leap 15.4`
+- `Cirros 0.5.2`
+- `Alpine`
+- `Arch`
```
fkr, line 40: Imho we should reduce this to just CIrros and remove the other three, in order to keep it as short as possible.

fkr (approved):
Left a few comments, nothing blocking. thanks for the work!

fkr:
@frosty-geek @Nils98Ar @jnull  please comment through the eyes of a CSP.

jnull:
@wwentland please verify this proposal with our naming conventions

Standards/scs-0104-v1-standard-images.md
frosty-geek: +1

frosty-geek (approved):
LGTM

Standards/scs-0104-v1-standard-images.md
mbuechse: If we go ahead with this, should we maybe move Cirros to the recommended list? I mean, the term "looking into standard suggestions" doesn't mean much, and for some reason the test script really does contain a list called `sugg_images` as though these images are suggested (which they aren't if you read closely).

Nils98Ar:
Looks good!

Is it planned to provide all recommended and suggested images via the image manager? Even SLES, RHEL or Windows Server to use it with BYOL?

Standards/scs-0104-v1-standard-images.md
osfrickler: Cirros 0.5.2 is pretty outdated. I also would recommend to not include cirros at all, it should never be deployed on a public cloud where access from the outside is possible.

Standards/scs-0104-v1-standard-images.md
```diff
@@ -0,0 +1,44 @@
...
+  the old `<PREVLTS>` version becomes optional ...
+- For Debian, we use the latest STABLE version, which is `12` at the time of this writing.
+  Similar to Ubuntu, we would do the switch and require the latest STABLE to be made available within
+  ~3 months after release.
+
+The following images are recommended:
+
+- `Alma Linux 8`, `Alma Linux 9` (whichever versions are being supported upstream)
+- `Debian <PREVSTABLE>` (the one pre-latest `<STABLE>`, `10` at the time of writing)
+- `Fedora <LATEST>` (`36` currently, this will get replaced quickly as the next Fedora comes out)
```
osfrickler, line 26: Debian oldstable is 11 now, Fedora latest is 38.

Standards/scs-0104-v1-standard-images.md
```diff
@@ -0,0 +1,44 @@
...
+- `SLES 15SP4`
+- `RHEL 9`, `RHEL 8`
+- `Windows Server 2022`, `Windows Server 2019`
+
+We are also looking into standard suggestions for
+
+- `openSUSE Leap 15.4`
+- `Cirros 0.5.2`
+- `Alpine`
+- `Arch`
```
fkr, line 39:
```suggestion
```

fkr: As per comment by @osfrickler just remove it all.

Standards/scs-0102-v2-image-metadata.md
```diff
@@ -0,0 +1,240 @@
...
+  to the image via the injected SSH key or provided password. (This can be set to `none` if no default
+  user name exists for the operating system.)
+- Optional: `patchlevel` can be set to an operating specific patch level that describes the
+  patch status — typically we would expect the `image_build_date` to be sufficient.
+
+- Recommended: `os_hash_algo` and `os_hash_value`: The sha256 or sha512 hash
+  for the image file. (This references the image file in the format it is stored in, we
+  recommend raw over qcow2 for systems that use ceph.) Note that these values are
+  typically generated automatically upon image registration.
+- Recommended: Digital image signature according to the [glance image
```
josephineSei, line 189: Note: this is not possible for Public images and thus should not be stated here.
I have an example to show here: https://input.scs.community/RID5cVHpTaG_AEDYsB9gVQ

mbuechse: Probably a very good point, but I suppose this PR is the wrong place, because it only splits the standard into two separate documents.

mbuechse: @josephineSei To be even more clear: I didn't write any of that. I only copied the text into a new file. However, the new file will have draft status (as is required by scs-0001), so we can still discuss this topic and change the document once this PR is through. By any means, raise your concerns, but preferably not with this PR (but rather in the team IaaS meetings), because the problem exists in V1 of the standard, which is currently in effect, and V2 will be in draft state afterwards and can be changed. This course of action had actually been agreed upon with Markus.

fkr:
> Is it planned to provide all recommended and suggested images via the image manager? Even SLES, RHEL or Windows Server to use it with BYOL?

No, the images must be downloaded by the entity that uploads them to the deployment.
The [OpenStack Image Manager](https://github.com/osism/openstack-image-manager/) is the tooling that can be used for that (as in, that is part of the reference implementation). With the [OpenStack Image Crawler](https://github.com/pluscloudopen/openstack-image-crawler) there is tooling available to automatically download the proper images from upstream (however only for images that are publically available).

Tests/iaas/scs-0104-v1-images.yaml
```diff
@@ -0,0 +1,15 @@
+images:
+- name: "Ubuntu 22.04"
+  source: https://cloud-images.ubuntu.com/releases/jammy/
+  status: mandatory
+- name: "Kubernetes CAPI 1.27.3"
```
garloff, line 5: k8s-cluster-api-provider uses and expects a different name here, see:
https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/6d08e840273dcdb5f02edcc73f36b3d509822826/terraform/files/bin/upload_capi_image.sh#L24
`ubuntu2204-capi-image-v1.27.3`
This has been like this for a while.
So we either adjust the k8s-capi scripts to make the capi img useful or we adjust the standard to what the scripts use and expect.

Note 1: The canonical name according to the naming recommendations would probably be "Ubuntu 22.04 CAPI K8S v1.27.3" or so. That would require changes on both ends (standards and tooling).
Note 2: These CAPI images will change every month, as we have one patch level per month roughly in k8s ... So we would either have to accept recommending outdated things in the standard or to change the standard every month or so ...

fkr (approved):

Tests/iaas/scs-0104-v1-images.yaml
garloff: Proposal from the IaaS team call: Don't fix the k8s version as recommendation. Just define the naming scheme for the capi images.

josephineSei (approved):

Tests/iaas/scs-0104-v1-images.yaml
garloff: > the script above uses `ubuntu-22.04-capi-image-v1.27.3`

The referenced line would resolve to `ubuntu-2204-kube-v1.27.3` which is the (filename minus the .qcow2) that is used to download the node image from osism's minio.
The OpenStack image name used is defined by the `OPENSTACK_IMAGE_NAME` as defined in clusterctl.yaml.
We should just align our defaults here ... If we use names with a space, we'll need to do some extra testing to ensure the quoting is done correctly in all scripts.

mbuechse: @garloff I opted for that (`ubuntu-22.04-capi-image-v1.27.3`) in the latest version, and if that goes into effect, then the CSPs have to align, and I guess they will (@berendt for instance told me that he doesn't have a strong opinion on that).

garloff: And the default `OPENSTACK_IMAGE_NAME` in k8s-cluster-api-provider is defined here:
https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/65af7217a4c2d849d1a408d36a3c9c6ea6c554aa/terraform/files/template/clusterctl.yaml.tmpl#L15
`OPENSTACK_IMAGE_NAME: ubuntu-capi-image-${kubernetes_version}`
So `ubuntu-capi-image-v1.XX.Y` not just in gx-scs.
My apologies for the confusion!

Going forward, I would like to ensure that:
* The default image-manager settings
* The recommended naming scheme in the std image spec
* The code in k8s-cluster-api-provider
all use the same. 

Right now my preference would be to stick to what we use in k8s-capi: `ubuntu-capi-image-v1.xx.yy`
If we change it, probably we should go to the canonical name. Doing this now, right before the release, is not optimal though.

mbuechse: @garloff So the two code occurrences contradict each other? This should be fixed, no? I'm fine with using either one of the two, and I would not go through the hassle of changing it to a third option. Still, the code occurrences should be harmonized, shouldn't they?

garloff: There is no contradiction, just my own confusion, for which I apologise.
To create a cluster with Kubernetes v1.xx.yy, the current k8s-cluster-api-provider scripts look for an image ubuntu-capi-image-v1.XX.YY. (This has been like this since R2 or so.)
If the image does not exist, a file ubuntu-2204-kube-v1.XX.YY.qcow2 would be downloaded from osism minio and registered as ubuntu-capi-image-v1.XX.YY image. (This has been like this as well for a long time, except that the -2204 was added in when we switched to 22.04.)

So, ubuntu-capi-image-v1.XX.YY it is from a k8s-cluster-api-provider perspective.

fkr: _If_ the image is downloaded if it does not exist (which is a very good fallback, since it is self contained), I see no reason to actually reference it in the Iaas-related standard. Or do I overlook something?

fkr (approved):
LGTM in its current state. 
The kube image is only recommended, so the discussion around that does not block merging this PR imho.

Tests/iaas/scs-0104-v1-images.yaml
garloff: If the image already has been registered before, cluster creation will be a few minutes. Download and registration typically add 10mins to that. In addition you get a private image, which is inefficient if many people / projects use capi. So a recommendation to the IaaS provider would seem useful to me. Better user experience, less wasted storage.

matfechner (approved):
LGTM

PR #318, opened 2023-06-29
Add test for k8s version recency (#288)

### LINKED ISSUE #288: Write tests for standard "Offered K8s version recency"
This issue should create a compliance test for the K8s version recency standard. This is part of the SCS standardization work, which should provide tests for all provided SCS standards.
The relevant Standard can be found [here](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0210-v1-k8s-new-version-policy.md).

This issue should provide
- [ ] a compliance test for the K8s version recency standard

The test itself should
- [ ] check the version of the to-be tested cluster against the current cluster versions, the cadence time is written in the standard
- [ ] return 0 if cluster version is in the cadence time or an error code if it isn't

## Definition of Ready:

- [x] User Story is small enough to be finished within one sprint
- [ ] User Story is clear and understood by the whole team
- [x] Acceptance criteria are defined
- [ ] Acceptance criteria are clear and understood by the whole team

## Definition of Done:

- [ ] All acceptance criteria are met
- [ ] Changes have been reviewed
- [ ] CI tests have run successfully
- [ ] Documentation has been updated
- [ ] Release Notes have been updated

cah-hbaum:
This PR adds a test for the K8s version recency standard.
A user can execute the test script and provide his Kubernetes Cluster config file in order to check if the cluster version
is still in line with the versions demanded by the standard.

The following requirements are set:
* Minor version is not older then 4 months
* Patch version is not older then 1 week
* Versions with CVEs need to be patched faster  

This PR also adds a configuration file for the test script, which allows setting different logging parameters as well as the Github Token required to access the latest K8s release information in a readable format.

Closes #288

### DIFF
--- a/Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
+++ b/Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
@@ -1,0 +1,465 @@
+#!/usr/bin/env python3
+# vim: set ts=4 sw=4 et:
+#
+"""
+K8s Version Recency Checker
+https://github.com/SovereignCloudStack/standards
+
+Return codes:
+0: Version is inside the recency window
+1: Error during script execution
+2: Version isn't inside the recency windows anymore
+3: Version used contains a critical CVE
+
+One or more K8s clusters are checked by providing their kubeconfigs.
+It is determined, if the version on these clusters is still inside
+the recency window, which is determined by the Standard to be 4 months
+for minor versions and 1 week for patch versions. An exception are
+versions with critical CVEs, which should be replaced on a shorter notice.
+
+(c) Hannes Baum <hannes.baum@cloudandheat.com>, 6/2023
+License: CC-BY-SA 4.0
+"""
+
+import aiohttp
+import asyncio
+import datetime
+import getopt
+import kubernetes_asyncio
+import logging
+import logging.config
+import math
+import re
+import requests
+import sys
+import yaml
+
+
+MAJOR_VERSION_CADENCE = None
+MINOR_VERSION_CADENCE = 4  # months
+PATCH_VERSION_CADENCE = 1  # week
+CVE_VERSION_CADENCE = 3  # days
+CVE_SEVERITY = 8  # CRITICAL
+
+logging_config = {
+    "level": "INFO",
+    "version": 1,
+    "disable_existing_loggers": False,
+    "formatters": {
+        "k8s-version-recency-check": {
+            "format": "%(levelname)s: %(message)s"
+        }
+    },
+    "handlers": {
+        "console": {
+            "class": "logging.StreamHandler",
+            "formatter": "k8s-version-recency-check",
+            "stream": "ext://sys.stdout"
+        }
+    },
+    "root": {
+        "handlers": ["console"]
+    }
+}
+
+logger = logging.getLogger("k8s-version-recency-check")
+
+
+class ConfigException(BaseException):
+    """Exception raised in a configuration error occurs"""
+
+
+class HelpException(BaseException):
+    """Exception raised if the help functionality is called"""
+
+
+class Config:
+    config_path = "./config.yaml"
+    kubeconfig = None
+    logging = None
+
+
+def print_usage():
+    print("""
+K8s Version Recency Compliance Check
+
+Usage: k8s-version-recency-check.py [-h] [-k|--kubeconfig PATH/TO/KUBECONFIG]
+
+The K8s version recency check returns 0 if the version of the tested cluster is still acceptable, otherwise
+it returns 1 for an out-of-date minor version, 2 for an out-of-date patch level version or 3 if the currently
+used version should be updated due to a highly critical CVE.
+
+    -c/--config PATH/TOCONFIG          - Path to the config file of the test script
+    -k/--kubeconfig PATH/TO/KUBECONFIG - Path to the kubeconfig of the server we want to check
+    -h                                 - Output help
+    """)
+
+
+def parse_arguments(argv):
+    """Parse cli arguments from the script call"""
+    config = Config()
+
+    try:
+        opts, args = getopt.gnu_getopt(argv, "c:k:h", ["config", "kubeconfig", "help"])
+    except getopt.GetoptError:
+        raise ConfigException
+
+    for opt in opts:
+        if opt[0] == "-h" or opt[0] == "--help":
+            raise HelpException
+        if opt[0] == "-c" or opt[0] == "--config":
+            config.config_path = opt[1]
+        if opt[0] == "-k" or opt[0] == "--kubeconfig":
+            config.kubeconfig = opt[1]
+
+    return config
+
+
+def setup_logging(config_log):
+
+    logging.config.dictConfig(config_log)
+    loggers = [
+        logging.getLogger(name)
+        for name in logging.root.manager.loggerDict
+        if not logging.getLogger(name).level
+    ]
+
+    for log in loggers:
+        log.setLevel(config_log['level'])
+
+
+def initialize_config(config):
+    """Initialize the configuration for the test script"""
+
+    try:
+        with open(config.config_path, "r") as f:
+            config.logging = yaml.safe_load(f)['logging']
+    except OSError:
+        logger.warning(f"The config file under {config.config_path} couldn't be found, "
+                       f"falling back to the default config.")
+    finally:
+        # Setup logging if the config file with the relevant information could be loaded before
+        # Otherwise, we initialize logging with the included literal
+        setup_logging(config.logging or logging_config)
+
+    if not config.kubeconfig:
+        print("A kubeconfig needs to be set in order to test a k8s cluster version.")
+        raise ConfigException
+
+    return config
+
+
+class K8sVersionInfo:
+    """Class that contains a k8s version info.
+
+    Attributes:
+        major (int): Major version of the k8s version
+        minor (int): Minor version of the k8s version
+        patch (int): Patch version of the k8s version
+        date (datetime): release date of the k8s version
+    """
+    def __init__(self, major=0, minor=0, patch=0):
+        self.major = major
+        self.minor = minor
+        self.patch = patch
+
+        self.date = None
+
+    def __eq__(self, other):
+        if not isinstance(other, K8sVersionInfo):
+            return False
+        return self.major == other.major and self.minor == other.minor and self.patch == other.patch
+
+    def __gt__(self, other):
+        if not isinstance(other, K8sVersionInfo):
+            return False
+        patchcomp = self.minor == other.minor and self.patch > other.patch
+        return self.major > other.major or (self.major == other.major and (self.minor > other.minor or patchcomp))
+
+    def __ge__(self, other):
+        if not isinstance(other, K8sVersionInfo):
+            return False
+        patchcomp = self.minor == other.minor and self.patch >= other.patch
+        return self.major > other.major or (self.major == other.major and (self.minor > other.minor or patchcomp))
+
+    def __lt__(self, other):
+        if not isinstance(other, K8sVersionInfo):
+            return False
+        patchcomp = self.minor == other.minor and self.patch < other.patch
+        return self.major < other.major or (self.major == other.major and (self.minor < other.minor or patchcomp))
+
+    def __le__(self, other):
+        if not isinstance(other, K8sVersionInfo):
+            return False
+        patchcomp = self.minor == other.minor and self.patch <= other.patch
+        return self.major < other.major or (self.major == other.major and (self.minor < other.minor or patchcomp))
+
+    @classmethod
+    def extract_version(cls, string, separator="."):
+        components = string.strip().split(separator)
+        return cls(int(components[0]), int(components[1]), int(components[2]))
+
+    def check_for_version(self, major=None, minor=None, patch=None):
+        return (major is None or self.major == major) and \
+               (minor is None or self.minor == minor) and \
+               (patch is None or self.patch == patch)
+
+    def __str__(self):
+        return f"{self.major}.{self.minor}.{self.patch}"
+
+
+class CVEVersionInfo:
+    def __init__(self, lower_version, upper_version, less_than=False, equal=False):
+        self.lower_version = lower_version
+        self.upper_version = upper_version
+
+        self.less_than = less_than
+        self.equal = equal
+
+    def __eq__(self, other):
+        if not isinstance(other, CVEVersionInfo):
+            return False
+        return self.lower_version == other.lower_version and self.upper_version == other.upper_version and \
+            self.less_than == other.less_than and self.equal == self.equal
+
+    def is_version_affected(self, version_info):
+        if self.upper_version:
+            if self.lower_version:
+                gt = self.lower_version <= version_info
+            else:
+                gt = True
+            if self.less_than:
+                if self.equal:
+                    return gt and self.upper_version >= version_info
+                return gt and self.upper_version > version_info
+        else:
+            return self.lower_version == version_info
+
+
+def diff_months(date1, date2):
+    return abs((date1.year - date2.year) * 12 + date1.month - date2.month)
+
+
+def diff_weeks(date1, date2):
+    day1 = (date1 - datetime.timedelta(days=date1.weekday()))
+    day2 = (date2 - datetime.timedelta(days=date2.weekday()))
+    diff = day2 - day1
+    return abs((diff.days / 7) + math.ceil(diff.seconds / 86400))
+
+
+def diff_days(date1, date2):
+    delta = date1 - date2
+    return abs(delta.days)
+
+
+async def request_cve_data(session: aiohttp.ClientSession, cveid: str) -> dict:
+    """Request for a single CVE data item."""
+    resp = await session.request('GET', f"https://cveawg.mitre.org/api/cve/{cveid}",
+                                 headers={"Accept": "application/json"})
+    return await resp.json()
+
+
+def parse_cve_version_information(cve_version_info):
+    """Parse the CVE version information according to their CVE JSON 5.0 schema"""
+    vi_lower_version = None
+    vi_upper_version = None
+    less_than = False
+    equal = False
+
+    # Extract the version if it is viable, but it's not a requirement
+    try:
+        vi_lower_version = K8sVersionInfo().extract_version(cve_version_info['version'].strip("v"))
+    except ValueError:
+        pass
+
+    if 'lessThanOrEqual' in cve_version_info:
+        vi_upper_version = K8sVersionInfo().extract_version(cve_version_info['lessThanOrEqual'].strip("v"))
+        vi_upper_version.patch += 1
+        less_than = True
+        equal = True
+    elif 'lessThan' in cve_version_info:
+        vi_upper_version = K8sVersionInfo().extract_version(cve_version_info['lessThan'].strip("v"))
+        less_than = True
+
+    # This shouldn't happen, but if it happens, we look for non-standard descriptions
+    # According to this(https://www.cve.org/AllResources/CveServices#cve-json-5),
+    # this isn't how the data should be described
+    if not vi_lower_version and not vi_upper_version:
+        if re.search(r'v?\d+.\d+.x', cve_version_info['version']):
+            vdata = cve_version_info['version'].strip("v").split(".")
+            vi_lower_version = K8sVersionInfo(vdata[0], vdata[1], 0)
+            vi_upper_version = K8sVersionInfo(vdata[0], int(vdata[1])+1, 0)
+
+        if re.search(r'v?\d+.\d+.\d+\s+-\s+v?\d+.\d+.\d+', cve_version_info['version']):
+            vdata = cve_version_info['version'].split("-")
+            vi_lower_version = K8sVersionInfo().extract_version(vdata[0].strip("v"))
+            vi_upper_version = K8sVersionInfo().extract_version(vdata[1].strip("v"))
+
+    return CVEVersionInfo(vi_lower_version, vi_upper_version, less_than, equal)
+
+
+async def collect_cve_versions():
+    """Get all relevant CVE versions, that are relevant for the test according to the severity
+    dictated by the Standard.
+    """
+
+    # CVE fix versions
+    cfvs = list()
+
+    # Request latest version
+    cve_list = requests.get("https://kubernetes.io/docs/reference/issues-security/official-cve-feed/index.json",
+                            headers={"Accept": "application/json"}).json()
+
+    async with aiohttp.ClientSession() as session:
+        tasks = []
+        for cve in cve_list['items']:
+            cveid = cve['external_url'].split("=")[-1]
+            tasks.append(request_cve_data(session=session, cveid=cveid))
+        # asyncio.gather() will wait on the entire task set to be
+        # completed.  If you want to process results greedily as they come in,
+        # loop over asyncio.as_completed()
+        cve_data_list = await asyncio.gather(*tasks, return_exceptions=True)
+
+    for cve_data in cve_data_list:
+        try:
+            cve_cna = cve_data['containers']['cna']
+            cve_metrics = cve_cna['metrics']
+            cve_affected = cve_cna['affected']
+            # This data is extracted like this due to the location of the relevant information in the schema.
+            # mitre.org uses CVE schema 5.0, which is described in the link below
+            # https://github.com/CVEProject/cve-schema/tree/master/schema/v5.0
+            # The containers -> cna path contains vulnerability information like severity, which is documented
+            # under the metrics list.
+        except KeyError as e:
+            logger.debug(f"They key {e} couldn't be found in the CVE json data for CVE {cveid}.")
+            continue
+
+        is_high_severity = any(
+            re.search(r'[cC][vV][sS]{1,2}V\d', metric_key) and metric_value['baseScore'] >= CVE_SEVERITY
+            for cve_metric in cve_metrics
+            for metric_key, metric_value in cve_metric.items()
+        )
+
+        if is_high_severity:
+            affected_kubernetes_versions = [
+                parse_cve_version_information(version_info)
+                for aff in cve_affected
+                if aff['product'] == "Kubernetes"
+                for version_info in aff['versions']
+                if version_info['status'] == "affected"
+            ]
+            for cvev in affected_kubernetes_versions:
+                if cvev not in cfvs:
+                    cfvs.append(cvev)
+
+    return cfvs
+
+
+async def get_k8s_cluster_version(kubeconfig):
+    """Get the k8s version of the cluster under test."""
+    cluster_config = await kubernetes_asyncio.config.load_kube_config(kubeconfig)
+
+    async with kubernetes_asyncio.client.ApiClient() as api:
+        version_api = kubernetes_asyncio.client.VersionApi(api)
+        ret = await version_api.get_code()
+
+        version = K8sVersionInfo.extract_version(ret.git_version.strip("v"))
+        version.date = datetime.datetime.strptime(ret.build_date, '%Y-%m-%dT%H:%M:%SZ')
+
+        return version, cluster_config.current_context['name']
+
+
+def is_current_version(version, k8s_version_list, cve_version_list=[]):
+    """Filter the versions depending on the usage times set by the Standard and
+    the times set for CVE versions.
+    """
+    try:
+        if diff_months(version.date, datetime.datetime.now()) >= MINOR_VERSION_CADENCE:
+            return False
+
+        for kv in k8s_version_list:
+            if version.check_for_version(major=kv.major, minor=kv.minor) and \
+               version.patch < kv.patch:
+
+                if diff_weeks(datetime.datetime.now(), kv.date) >= PATCH_VERSION_CADENCE:
+                    return False
+
+                if kv in cve_version_list and \
+                   diff_days(datetime.datetime.now(), kv.date) >= CVE_VERSION_CADENCE:
+                    return False
+    except (KeyError, IndexError, TypeError) as e:
+        logger.debug(f"An error occurred during version filtering: {e}")
+        return False
+    else:
+        return True
+
+
+def collect_accepted_k8s_versions(cve_version_list=[]):
+    """Collect a list of k8s versions that comply to the cadence time set by the standard"""
+
+    k8s_versions = []
+
+    github_headers = {
+        "Accept": "application/vnd.github+json",
+        "X-GitHub-Api-Version": "2022-11-28"
+    }
+
+    # Request latest version
+    resp = requests.get("https://api.github.com/repos/kubernetes/kubernetes/releases/latest",
+                        headers=github_headers).json()
+
+    # Latest version
+    lv = K8sVersionInfo.extract_version(resp['tag_name'].strip("v"))
+    lv.date = datetime.datetime.strptime(resp['published_at'], '%Y-%m-%dT%H:%M:%SZ')
+
+    # Request the latest 100 version (the next are not needed, since these versions are too old)
+    response = requests.get("https://api.github.com/repos/kubernetes/kubernetes/releases?per_page=100",
+                            headers=github_headers).json()
+
+    # Iterate all versions until the first patch of the previous latest version
+    for r in response:
+        v = K8sVersionInfo.extract_version(r['tag_name'].split("-")[0].strip("v"))
+        v.date = datetime.datetime.strptime(r['published_at'], '%Y-%m-%dT%H:%M:%SZ')
+
+        if not r['draft'] and not r['prerelease']:
+            k8s_versions.append(v)
+
+        # Stop adding new version if the version we added is the previous latest minor versions first patch,
+        # since it is most of the time around 3-6 months until a new version comes out or an old one goes into EOL
+        if v.check_for_version(lv.major, (lv.minor-1), 0):
+            break
+
+    return [v for v in k8s_versions if is_current_version(v, k8s_versions, cve_version_list)]
+
+
+async def main(argv):
+    try:
+        config = initialize_config(parse_arguments(argv))
+    except (OSError, ConfigException, HelpException):
+        print_usage()
+        return 1
+
+    cve_versions = await collect_cve_versions()
+    k8s_versions = collect_accepted_k8s_versions(cve_versions)
+    cluster_version, cluster_name = await get_k8s_cluster_version(config.kubeconfig)
+
+    for k8sv in k8s_versions:
+        if k8sv == cluster_version:
+            logger.info("The K8s cluster version %s of cluster '%s' is still in the recency time window." %
+                        (str(cluster_version), cluster_name))
+            return 0
+
+    for cvev in cve_versions:
+        if cvev.is_version_affected(cluster_version):
+            logger.error("The K8s cluster version %s of cluster '%s' is an outdated version "
+                         "with a possible CRITICAL CVE." % (str(cluster_version), cluster_name))
+            return 3
+
+    logger.error("The K8s cluster version %s of cluster '%s' is outdated according to the Standard." %
+                 (str(cluster_version), cluster_name))
+    return 2
+
+
+if __name__ == "__main__":
+    return_code = asyncio.run(main(sys.argv[1:]))
+    sys.exit(return_code)
--- a/Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
+++ b/Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
@@ -24,11 +24,11 @@ License: CC-BY-SA 4.0
 import aiohttp
 import asyncio
 import datetime
+from dateutil import relativedelta
 import getopt
 import kubernetes_asyncio
 import logging
 import logging.config
-import math
 import re
 import requests
 import sys
@@ -36,9 +36,9 @@ import yaml
 
 
 MAJOR_VERSION_CADENCE = None
-MINOR_VERSION_CADENCE = 4  # months
-PATCH_VERSION_CADENCE = 1  # week
-CVE_VERSION_CADENCE = 3  # days
+MINOR_VERSION_CADENCE_MONTHS = 4
+PATCH_VERSION_CADENCE_WEEKS = 1
+CVE_VERSION_CADENCE_DAYS = 3
 CVE_SEVERITY = 8  # CRITICAL
 
 logging_config = {
@@ -62,7 +62,7 @@ logging_config = {
     }
 }
 
-logger = logging.getLogger("k8s-version-recency-check")
+logger = logging.getLogger(__name__)
 
 
 class ConfigException(BaseException):
@@ -83,13 +83,12 @@ def print_usage():
     print("""
 K8s Version Recency Compliance Check
 
-Usage: k8s-version-recency-check.py [-h] [-k|--kubeconfig PATH/TO/KUBECONFIG]
+Usage: k8s-version-recency-check.py [-h] [-c|--config PATH/TO/CONFIG] -k|--kubeconfig PATH/TO/KUBECONFIG
 
 The K8s version recency check returns 0 if the version of the tested cluster is still acceptable, otherwise
-it returns 1 for an out-of-date minor version, 2 for an out-of-date patch level version or 3 if the currently
-used version should be updated due to a highly critical CVE.
+it returns 2 for an out-of date version or 3 if the used version should be updated due to a highly critical CVE.
 
-    -c/--config PATH/TOCONFIG          - Path to the config file of the test script
+    -c/--config PATH/TO/CONFIG         - Path to the config file of the test script
     -k/--kubeconfig PATH/TO/KUBECONFIG - Path to the kubeconfig of the server we want to check
     -h                                 - Output help
     """)
@@ -142,9 +141,8 @@ def initialize_config(config):
         # Otherwise, we initialize logging with the included literal
         setup_logging(config.logging or logging_config)
 
-    if not config.kubeconfig:
-        print("A kubeconfig needs to be set in order to test a k8s cluster version.")
-        raise ConfigException
+    if config.kubeconfig is None:
+        raise ConfigException("A kubeconfig needs to be set in order to test a k8s cluster version.")
 
     return config
 
@@ -167,39 +165,44 @@ class K8sVersionInfo:
 
     def __eq__(self, other):
         if not isinstance(other, K8sVersionInfo):
-            return False
+            raise TypeError
         return self.major == other.major and self.minor == other.minor and self.patch == other.patch
 
     def __gt__(self, other):
         if not isinstance(other, K8sVersionInfo):
-            return False
+            raise TypeError
         patchcomp = self.minor == other.minor and self.patch > other.patch
         return self.major > other.major or (self.major == other.major and (self.minor > other.minor or patchcomp))
 
     def __ge__(self, other):
         if not isinstance(other, K8sVersionInfo):
-            return False
+            raise TypeError
         patchcomp = self.minor == other.minor and self.patch >= other.patch
         return self.major > other.major or (self.major == other.major and (self.minor > other.minor or patchcomp))
 
     def __lt__(self, other):
         if not isinstance(other, K8sVersionInfo):
-            return False
+            raise TypeError
         patchcomp = self.minor == other.minor and self.patch < other.patch
         return self.major < other.major or (self.major == other.major and (self.minor < other.minor or patchcomp))
 
     def __le__(self, other):
         if not isinstance(other, K8sVersionInfo):
-            return False
+            raise TypeError
         patchcomp = self.minor == other.minor and self.patch <= other.patch
         return self.major < other.major or (self.major == other.major and (self.minor < other.minor or patchcomp))
 
     @classmethod
-    def extract_version(cls, string, separator="."):
+    def extract_version(cls, string, separator=".", strip=None):
+        if strip is None:
+            strip = ["v"]
+        for s in strip:
+            string = string.strip(s)
         components = string.strip().split(separator)
         return cls(int(components[0]), int(components[1]), int(components[2]))
 
     def check_for_version(self, major=None, minor=None, patch=None):
+        """Check if a version or part of the version is equal to the given version numbers"""
         return (major is None or self.major == major) and \
                (minor is None or self.minor == minor) and \
                (patch is None or self.patch == patch)
@@ -209,42 +212,57 @@ class K8sVersionInfo:
 
 
 class CVEVersionInfo:
-    def __init__(self, lower_version, upper_version, less_than=False, equal=False):
+    """Class that contains a CVE version info.
+
+    Attributes:
+        upper_version (K8sVersionInfo): Last version with the CVE
+        lower_version (K8sVersionInfo): First version with the CVE; this value will be set if either an affected version
+            is directly set in a CVE dataset or if the CVE dataset is in a non-standard format.
+            If the variable is set, `lower_version` and `upper_version` create a range of affected versions.
+        equal (bool): check if the version is equal to the `upper_version`, (less than is always checked, since the
+            format is build like this)
+    """
+    def __init__(self, lower_version, upper_version, equal=False):
         self.lower_version = lower_version
         self.upper_version = upper_version
 
-        self.less_than = less_than
         self.equal = equal
 
     def __eq__(self, other):
         if not isinstance(other, CVEVersionInfo):
-            return False
-        return self.lower_version == other.lower_version and self.upper_version == other.upper_version and \
-            self.less_than == other.less_than and self.equal == self.equal
+            raise TypeError
+        return self.lower_version == other.lower_version and \
+            self.upper_version == other.upper_version and \
+            self.equal == self.equal
 
     def is_version_affected(self, version_info):
+        # See the following link for more information about the format
+        # https://www.cve.org/AllResources/CveServices#cve-json-5
+
+        # Check if an `upper version` exists
         if self.upper_version:
+            # Check if a `lower version` exists and compare the version against it
             if self.lower_version:
                 gt = self.lower_version <= version_info
             else:
                 gt = True
-            if self.less_than:
-                if self.equal:
-                    return gt and self.upper_version >= version_info
-                return gt and self.upper_version > version_info
+            # Compare the version either with `less than` or `less than or equal` against the `upper version`
+            if self.equal:
+                return gt and self.upper_version >= version_info
+            return gt and self.upper_version > version_info
         else:
+            # If no upper version exists, we only need to check if the version is equal to the `lower version`
             return self.lower_version == version_info
 
 
 def diff_months(date1, date2):
-    return abs((date1.year - date2.year) * 12 + date1.month - date2.month)
+    r = relativedelta.relativedelta(date2, date1)
+    return r.months + (12 * r.years)
 
 
 def diff_weeks(date1, date2):
-    day1 = (date1 - datetime.timedelta(days=date1.weekday()))
-    day2 = (date2 - datetime.timedelta(days=date2.weekday()))
-    diff = day2 - day1
-    return abs((diff.days / 7) + math.ceil(diff.seconds / 86400))
+    delta = date1 - date2
+    return abs(delta.days / 7)
 
 
 def diff_days(date1, date2):
@@ -254,51 +272,49 @@ def diff_days(date1, date2):
 
 async def request_cve_data(session: aiohttp.ClientSession, cveid: str) -> dict:
     """Request for a single CVE data item."""
-    resp = await session.request('GET', f"https://cveawg.mitre.org/api/cve/{cveid}",
-                                 headers={"Accept": "application/json"})
-    return await resp.json()
+    async with session.get(
+        f"https://cveawg.mitre.org/api/cve/{cveid}",
+        headers={"Accept": "application/json"}
+    ) as resp:
+        return await resp.json()
 
 
 def parse_cve_version_information(cve_version_info):
     """Parse the CVE version information according to their CVE JSON 5.0 schema"""
     vi_lower_version = None
     vi_upper_version = None
-    less_than = False
     equal = False
 
     # Extract the version if it is viable, but it's not a requirement
     try:
-        vi_lower_version = K8sVersionInfo().extract_version(cve_version_info['version'].strip("v"))
+        vi_lower_version = K8sVersionInfo.extract_version(cve_version_info['version'])
     except ValueError:
         pass
 
     if 'lessThanOrEqual' in cve_version_info:
-        vi_upper_version = K8sVersionInfo().extract_version(cve_version_info['lessThanOrEqual'].strip("v"))
-        vi_upper_version.patch += 1
-        less_than = True
+        vi_upper_version = K8sVersionInfo.extract_version(cve_version_info['lessThanOrEqual'])
         equal = True
     elif 'lessThan' in cve_version_info:
-        vi_upper_version = K8sVersionInfo().extract_version(cve_version_info['lessThan'].strip("v"))
-        less_than = True
+        vi_upper_version = K8sVersionInfo.extract_version(cve_version_info['lessThan'])
 
     # This shouldn't happen, but if it happens, we look for non-standard descriptions
     # According to this(https://www.cve.org/AllResources/CveServices#cve-json-5),
     # this isn't how the data should be described
-    if not vi_lower_version and not vi_upper_version:
+    if vi_lower_version is None and vi_upper_version is None:
         if re.search(r'v?\d+.\d+.x', cve_version_info['version']):
             vdata = cve_version_info['version'].strip("v").split(".")
             vi_lower_version = K8sVersionInfo(vdata[0], vdata[1], 0)
-            vi_upper_version = K8sVersionInfo(vdata[0], int(vdata[1])+1, 0)
+            vi_upper_version = K8sVersionInfo(vdata[0], vdata[1], 0)
 
         if re.search(r'v?\d+.\d+.\d+\s+-\s+v?\d+.\d+.\d+', cve_version_info['version']):
             vdata = cve_version_info['version'].split("-")
-            vi_lower_version = K8sVersionInfo().extract_version(vdata[0].strip("v"))
-            vi_upper_version = K8sVersionInfo().extract_version(vdata[1].strip("v"))
+            vi_lower_version = K8sVersionInfo.extract_version(vdata[0])
+            vi_upper_version = K8sVersionInfo.extract_version(vdata[1])
 
-    return CVEVersionInfo(vi_lower_version, vi_upper_version, less_than, equal)
+    return CVEVersionInfo(vi_lower_version, vi_upper_version, equal)
 
 
-async def collect_cve_versions():
+async def collect_cve_versions(session: aiohttp.ClientSession):
     """Get all relevant CVE versions, that are relevant for the test according to the severity
     dictated by the Standard.
     """
@@ -307,18 +323,16 @@ async def collect_cve_versions():
     cfvs = list()
 
     # Request latest version
-    cve_list = requests.get("https://kubernetes.io/docs/reference/issues-security/official-cve-feed/index.json",
-                            headers={"Accept": "application/json"}).json()
+    async with session.get(
+        "https://kubernetes.io/docs/reference/issues-security/official-cve-feed/index.json",
+        headers={"Accept": "application/json"}
+    ) as resp:
+        cve_list = await resp.json()
 
-    async with aiohttp.ClientSession() as session:
-        tasks = []
-        for cve in cve_list['items']:
-            cveid = cve['external_url'].split("=")[-1]
-            tasks.append(request_cve_data(session=session, cveid=cveid))
-        # asyncio.gather() will wait on the entire task set to be
-        # completed.  If you want to process results greedily as they come in,
-        # loop over asyncio.as_completed()
-        cve_data_list = await asyncio.gather(*tasks, return_exceptions=True)
+    tasks = [request_cve_data(session=session, cveid=cve['external_url'].split("=")[-1])
+             for cve in cve_list['items']]
+
+    cve_data_list = await asyncio.gather(*tasks, return_exceptions=True)
 
     for cve_data in cve_data_list:
         try:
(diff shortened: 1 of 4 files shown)


cah-hbaum:
The things we also might want to think about here:
* a Github Token needs to be used in order to access the API calls to Github
* the recency variables could also be set/changed in the config, if we want to be flexible with respect for the standard
* `Versions with CVEs need to be patched faster` this section should probably defined better

Some improvements I also thought about:
* this test could be sped up by either 
   a) caching the CVE data for repeated uses somehow, since they take the most time or
   b) adding a `--disable-cve` option, which would disable it for the requested runs (but this would go against the Standard)
* test multiple clusters by providing multiple files (or just call the test multiple times, but this would require the caching solution from before)

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+            config_path = opt[1]
+        if opt[0] == "-k" or opt[0] == "--kubeconfig":
+            config.kubeconfig = opt[1]
+        if opt[0] == "-t" or opt[0] == "--token":
+            config.token = opt[1]
+
+    try:
+        with open(config_path, "r") as fd:
+            config_file = yaml.safe_load(fd)
+            fd.close()
```
mbuechse, line 117: Two things:
* I wouldn't call this a file descriptor, because a file descriptor is an integer used on the os level. This, on the other hand, is the Python file object, so I would call this `fileobj`.
* The call to close is superfluous; this is dealt with for us by the `with` block.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+        return version, cluster_config['contexts'][0]['name']
+
+
+def get_k8s_versions(token, cve_versions=None):
+
+    k8s_versions = []
+
+    github_headers = {
+        "Accept": "application/vnd.github+json",
+        "Authorization": f"Bearer {token}",
```
mbuechse, line 248: I'm wondering if this token business is really necessary. Which of the subsequent requests really need it? Can you add some more commentary about that? I'm asking because the token complicates matters, the information should be publicly available, and at least the "latest" page does indeed work straight from my browser without any special headers.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+        if resp[i]['draft'] is False and resp[i]['prerelease'] is False:
+            k8s_versions.append(v)
+
+        # Stop adding new version if the version we added is the previous latest minor versions first patch,
+        # since it is most of the time around 3-6 months until a new version comes out or an old one goes into EOL
+        if v.check_for_version(lv.major, (lv.minor-1), 0):
+            break
+
+    # Filter the versions depending on the usage times set by the standard
+    for i in range(0, len(k8s_versions)):
```
mbuechse, line 280: So this loop seems to replace certain entries in `k8s_versions` with `None`, only to then use list comprehension to obtain the list of the other entries. I think there are two more preferable ways to achieve this aim:
* Ideally, use list comprehension directly: `filtered = [v for v in k8s_versions if is_admissible(v)]` where you would extract the bulk of the loop's body into the new function `is_admissible`
* Or, at least use `enumerate` (`for i, v in enumerate(k8s_versions):`) -- the use of `k8s_versions.__getitem__(i)` instead of `k8s_versions[i]` is very baffling and (if really necessary) needs to be accompanied with a comment, but if `__getitem__` isn't necessary, then just use `enumerate`.

To summarize: If at all possible, try to do the list comprehension directly.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+
+    try:
+        config = load_config(argv)
+    except ConfigException:
+        return_code = 1
+        return
+
+    cve_versions = check_cve_versions()
+    k8s_versions = get_k8s_versions(config.token, cve_versions)
+    cluster_version, cluster_name = await get_k8s_cluster_version(config.kubeconfig)
```
mbuechse, line 316: Why not get the cluster version first and then, knowing the current version, look for CVEs and Kubernetes releases? Then you don't have to fetch all CVEs and all releases, only to restrict your attention later. Am I missing something?

mbuechse (commented):
I think this implementation does get the job done, and I also think that a lot of important research went into it (for finding the right endpoints for the Kubernetes releases and CVE data). Nevertheless, from my point of view, it needs a bit of polish here and there.

- the naming of the functions should better reflect what they do
- the functions also seem to mix levels of abstraction and concerns that I would prefer to see separated
- async should be capitalized on more
- some of the logic seems bit convoluted

mbuechse:
One thing that I forgot: This PR should also replace the TBD in the standard by just one paragraph explaining the name of the script and how to invoke it properly (this wouldn't warrant a dedicated issue in my opinion).

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Separated the comprehension in a new function.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+    resp = requests.get("https://api.github.com/repos/kubernetes/kubernetes/releases?per_page=100",
+                        headers=github_headers).json()
+
+    # Iterate all versions until the first patch of the previous latest version
+    for i in range(0, len(resp)):
+        v = K8sVersionInfo()
+        v.extract_version(resp[i]['tag_name'].split("-")[0].strip("v"))
+        v.date = datetime.datetime.strptime(resp[i]['published_at'], '%Y-%m-%dT%H:%M:%SZ')
+
+        if resp[i]['draft'] is False and resp[i]['prerelease'] is False:
```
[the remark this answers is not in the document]
cah-hbaum: Changed.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: That can be done, but like I asked in one of the first comments (and this was also an intention here), you could test multiple clusters with this script, if we want the functionality for this.
Yeah we could still do that, but that were my intentions with this order.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+    lv = K8sVersionInfo()
+    lv.extract_version(resp['tag_name'].strip("v"))
+    lv.date = datetime.datetime.strptime(resp['published_at'], '%Y-%m-%dT%H:%M:%SZ')
+
+    # Request the latest 100 version (the next are not needed, since these versions are too old)
+    resp = requests.get("https://api.github.com/repos/kubernetes/kubernetes/releases?per_page=100",
+                        headers=github_headers).json()
+
+    # Iterate all versions until the first patch of the previous latest version
+    for i in range(0, len(resp)):
```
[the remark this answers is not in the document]
cah-hbaum: Changed and I couldn't find a better name.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+        logging.getLogger(name)
+        for name in logging.root.manager.loggerDict
+        if not logging.getLogger(name).level
+    ]
+
+    for log in loggers:
+        log.setLevel(config_log['level'])
+
+
+def load_config(argv):
```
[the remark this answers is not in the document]
cah-hbaum: Separated into 2 functions, more seems overly complicated.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+class K8sVersionInfo:
+    def __init__(self, major=0, minor=0, patch=0):
+        self.major = major
+        self.minor = minor
+        self.patch = patch
+
+        self.date = None
+
+    def __eq__(self, other):
+        if self.major == other.major and \
```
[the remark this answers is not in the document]
cah-hbaum: Done.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+           self.minor == other.minor and \
+           self.patch == other.patch:
+            return True
+        else:
+            return False
```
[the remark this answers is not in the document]
cah-hbaum: Done.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+    try:
+        opts, args = getopt.gnu_getopt(argv, "t:k:h", ["token", "kubeconfig", "help"])
+    except getopt.GetoptError:
+        print_usage()
+        raise ConfigException
+
+    for opt in opts:
+        if opt[0] == "-h" or opt[0] == "--help":
+            print_usage()
+            sys.exit()
```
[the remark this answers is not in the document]
cah-hbaum: Changed with `asyncio.run()`.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+           self.patch == other.patch:
+            return True
+        else:
+            return False
+
+    def extract_version(self, string, separator="."):
+        v_dict = string.strip().split(separator)
+        self.major = int(v_dict[0])
+        self.minor = int(v_dict[1])
+        self.patch = int(v_dict[2])
```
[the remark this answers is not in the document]
cah-hbaum: Done.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+    def check_for_version(self, major=None, minor=None, patch=None):
+        if (major is None or self.major == major) and \
+           (minor is None or self.minor == minor) and \
+           (patch is None or self.patch == patch):
+            return True
+        else:
+            return False
+
+    def get_version(self):
+        return f"{self.major}.{self.minor}.{self.patch}"
```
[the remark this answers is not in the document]
cah-hbaum: Done.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+    # CVE fix versions
+    cfvs = list()
+
+    # Request latest version
+    cve_list = requests.get("https://kubernetes.io/docs/reference/issues-security/official-cve-feed/index.json",
+                            headers={"Accept": "application/json"}).json()
+
+    for cve in cve_list['items']:
+        cveid = cve['external_url'].split("=")[-1]
+        cve_data = requests.get(f"https://cveawg.mitre.org/api/cve/{cveid}",
```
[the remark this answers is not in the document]
cah-hbaum: Parallelized the requests.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+                for v in cve_cna['affected'][0]['versions']:
+                    vi = K8sVersionInfo()
+                    if 'lessThanOrEqual' in v.keys():
+                        vi.extract_version(v['lessThanOrEqual'].strip("v"))
+                        vi.patch += 1
+                    elif 'lessThan' in v.keys():
+                        vi.extract_version(v['lessThan'].strip("v"))
+                    if vi not in cfvs:
+                        cfvs.append(vi)
+        except KeyError as e:
```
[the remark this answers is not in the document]
cah-hbaum: I had multiple errors due to different keys missing, e.g. `cna`, `metrics` or `affected`, that's why I wrapped it around everything.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+        version.date = datetime.datetime.strptime(ret.build_date, '%Y-%m-%dT%H:%M:%SZ')
+
+        with open(kubeconfig, "r") as fd:
+            cluster_config = yaml.safe_load(fd)
+            fd.close()
+
+        return version, cluster_config['contexts'][0]['name']
+
+
+def get_k8s_versions(token, cve_versions=None):
```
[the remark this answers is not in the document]
cah-hbaum: Renamed the functions to better reflect their purpose.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+versions with critical CVEs, which should be replaced on a shorter notice.
+
+(c) Hannes Baum <hannes.baum@cloudandheat.com>, 6/2023
+License: CC-BY-SA 4.0
+"""
+
+import asyncio
+import datetime
+import getopt
+import kubernetes_asyncio
```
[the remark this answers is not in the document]
cah-hbaum: Added it to the requirements.txt

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Changed.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+        version_api = kubernetes_asyncio.client.VersionApi(api)
+        ret = await version_api.get_code()
+
+        version = K8sVersionInfo()
+        version.extract_version(ret.git_version.strip("v"))
+        version.date = datetime.datetime.strptime(ret.build_date, '%Y-%m-%dT%H:%M:%SZ')
+
+        with open(kubeconfig, "r") as fd:
+            cluster_config = yaml.safe_load(fd)
+            fd.close()
```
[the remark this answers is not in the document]
cah-hbaum: Changed it with the whole function.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
mbuechse: But it seems to work without the token...
```
$ curl https://api.github.com/repos/kubernetes/kubernetes/releases/latest
{
  "url": "https://api.github.com/repos/kubernetes/kubernetes/releases/108644552",
  "assets_url": "https://api.github.com/repos/kubernetes/kubernetes/releases/108644552/assets",
  "upload_url": "https://uploads.github.com/repos/kubernetes/kubernetes/releases/108644552/assets{?name,label}",
  "html_url": "https://github.com/kubernetes/kubernetes/releases/tag/v1.27.3",
  "id": 108644552,
  "author": {
    "login": "k8s-release-robot",
    "id": 33505452,
    "node_id": "MDQ6VXNlcjMzNTA1NDUy",
    "avatar_url": "https://avatars.githubusercontent.com/u/33505452?v=4",
    "gravatar_id": "",
    "url": "https://api.github.com/users/k8s-release-robot",
    "html_url": "https://github.com/k8s-release-robot",
    "followers_url": "https://api.github.com/users/k8s-release-robot/followers",
    "following_url": "https://api.github.com/users/k8s-release-robot/following{/other_user}",
    "gists_url": "https://api.github.com/users/k8s-release-robot/gists{/gist_id}",
    "starred_url": "https://api.github.com/users/k8s-release-robot/starred{/owner}{/repo}",
    "subscriptions_url": "https://api.github.com/users/k8s-release-robot/subscriptions",
    "organizations_url": "https://api.github.com/users/k8s-release-robot/orgs",
    "repos_url": "https://api.github.com/users/k8s-release-robot/repos",
    "events_url": "https://api.github.com/users/k8s-release-robot/events{/privacy}",
    "received_events_url": "https://api.github.com/users/k8s-release-robot/received_events",
    "type": "User",
    "site_admin": false
  },
  "node_id": "RE_kwDOAToIks4GecjI",
  "tag_name": "v1.27.3",
  "target_commitish": "master",
  "name": "Kubernetes v1.27.3",
  "draft": false,
  "prerelease": false,
  "created_at": "2023-06-14T09:47:38Z",
  "published_at": "2023-06-14T21:36:39Z",
  "assets": [

  ],
  "tarball_url": "https://api.github.com/repos/kubernetes/kubernetes/tarball/v1.27.3",
  "zipball_url": "https://api.github.com/repos/kubernetes/kubernetes/zipball/v1.27.3",
  "body": "\nSee [kubernetes-announce@](https://groups.google.com/forum/#!forum/kubernetes-announce). Additional binary downloads are linked in the [CHANGELOG](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.27.md).\n\nSee [the CHANGELOG](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.27.md) for more details.\n\n\n\n",
  "reactions": {
    "url": "https://api.github.com/repos/kubernetes/kubernetes/releases/108644552/reactions",
    "total_count": 45,
    "+1": 13,
    "-1": 0,
    "laugh": 2,
    "hooray": 15,
    "confused": 0,
    "heart": 4,
    "rocket": 5,
    "eyes": 6
  }
}
```

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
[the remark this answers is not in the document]
mbuechse: So, if `cna` or `metrics` isn't present, then the next item in `cve_data_list` will be processed, which sounds reasonable. But if `affected` is missing, shouldn't then the next item in `cve_metrics` be processed?

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
mbuechse: Got it.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+    diff = day2 - day1
+    return abs((diff.days / 7) + math.ceil(diff.seconds / 86400))
+
+
+def diff_days(date1, date2):
+    delta = date1 - date2
+    return abs(delta.days)
+
+
+def check_cve_versions():
```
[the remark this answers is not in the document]
mbuechse: The schema is really more helpful than I expected. I'm still wondering why we only process the first entry of the metrics list and the first entry of the affected list. Couldn't more than one product be affected, and then Kubernetes could be listed in second or third place?

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
mbuechse: For reals, calling the object `fd` is misleading. See, e.g., https://docs.python.org/3/library/os.html#os.fdopen

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Interesting, I got an error (I think 403) last time I tried that and then the [Docs](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases) brought me to the conclusion that I needed a token.
But the text also says only the 'draft' releases need a token, so maybe I did something wrong last time.

Then I will change that up.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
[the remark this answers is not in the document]
cah-hbaum: Implemented multiple try-except-blocks in order to catch different scenarios.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+
+    for cve in cve_list['items']:
+        cveid = cve['external_url'].split("=")[-1]
+        cve_data = requests.get(f"https://cveawg.mitre.org/api/cve/{cveid}",
+                                headers={"Accept": "application/json"}).json()
+        try:
+            cve_cna = cve_data['containers']['cna']
+            cve_metrics = cve_cna['metrics'][0]
+            key = None
+            for k in cve_metrics.keys():
```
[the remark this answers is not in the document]
cah-hbaum: Changed that.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,339 @@
...
+                if re.search(r'[cC][vV][sS]{1,2}V\d', k):
+                    key = k
+                    break
+
+            if cve_metrics[key]['baseScore'] >= CVE_SEVERITY:
```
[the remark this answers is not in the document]
cah-hbaum: Didn't implement this, since the solution now looks a bit different and the `continue` wasn't necessary for that.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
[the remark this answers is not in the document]
cah-hbaum: It now iterates over all entries in the `metrics` / `affected` lists and checks if Kubernetes is mentioned before using it.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,369 @@
...
+        date (datetime): release date of the k8s version
+    """
+    def __init__(self, major=0, minor=0, patch=0):
+        self.major = major
+        self.minor = minor
+        self.patch = patch
+
+        self.date = None
+
+    def __eq__(self, other):
```
mbuechse, line 159: I meant to say that if you implement `__eq__`, you also have to implement `__ne__`, BUT apparently, that is no longer true for Python 3. Just FYI.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,369 @@
...
+        if not isinstance(other, K8sVersionInfo):
+            return False
+        return self.major == other.major and self.minor == other.minor and self.patch == other.patch
+
+    @classmethod
+    def extract_version(cls, string, separator="."):
```
mbuechse, line 165: I'm wondering if this method should also strip away the `v`. Maybe even in a more strict manner than just using `strip`, like
```python
if string[:1] == "v":
    string = string[1:]
```

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,369 @@
...
+async def collect_cve_versions():
+    """Get all relevant CVE versions, that are relevant for the test according to the severity
+    dictated by the Standard.
+    """
+
+    # CVE fix versions
+    cfvs = list()
+
+    # Request latest version
+    cve_list = requests.get("https://kubernetes.io/docs/reference/issues-security/official-cve-feed/index.json",
```
mbuechse, line 211: I wouldn't mix `requests` and `aiohttp` in the same program. Also, I think `collect_cve_versions` should take a ClientSession as argument just like `request_cve_data` does.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,369 @@
...
+                            headers={"Accept": "application/json"}).json()
+
+    async with aiohttp.ClientSession() as session:
+        tasks = []
+        for cve in cve_list['items']:
+            cveid = cve['external_url'].split("=")[-1]
+            tasks.append(request_cve_data(session=session, cveid=cveid))
```
mbuechse, line 218: Two points:
- This looks like a classic case of a list comprehension.
- How many tasks are we expecting here? Can we make sure that the number of concurrent requests is bounded (as I said, maybe two or three)?

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,369 @@
...
+            # mitre.org uses CVE schema 5.0, which is described in the link below
+            # https://github.com/CVEProject/cve-schema/tree/master/schema/v5.0
+            # The containers -> cna path contains vulnerability information like severity, which is documented
+            # under the metrics list.
+        except KeyError as e:
+            logger.debug(f"They key {e} couldn't be found in the CVE json data for CVE {cveid}.")
+            continue
+        for i, cve_metric in enumerate(cve_metrics):
+            try:
+                cve_affected_versions = cve_cna['affected'][i]['versions']
```
mbuechse, line 238: I don't think the metrics and the affected products are linked. 
From my point of view, we have to do the following:

- iterate over the metrics and check whether one of them is the type and severity we're interested in
- if yes, iterate over the affected products and check whether one of them is Kubernetes
- if yes, iterate over the affected versions and add those to our list.

Note that affected versions can be given via ranges, such as "v1.25.0 - v1.25.10" [here](https://cveawg.mitre.org/api/cve/CVE-2023-2727). I have the impression that this requires more work.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,369 @@
...
+                   cve_cna['affected'][i]['product'] == "Kubernetes" and \
+                   cve_metric[key]['baseScore'] >= CVE_SEVERITY:
+
+                    for v in cve_affected_versions:
+                        vi = K8sVersionInfo()
+                        if 'lessThanOrEqual' in v:
+                            vi = vi.extract_version(v['lessThanOrEqual'].strip("v"))
+                            vi.patch += 1
+                        elif 'lessThan' in v:
+                            vi = vi.extract_version(v['lessThan'].strip("v"))
```
mbuechse, line 254: The `classmethod` should be called as such:

```suggestion
                        if 'lessThanOrEqual' in v:
                            vi = K8sVersionInfo.extract_version(v['lessThanOrEqual'].strip("v"))
                            vi.patch += 1
                        elif 'lessThan' in v:
                            vi = K8sVersionInfo.extract_version(v['lessThan'].strip("v"))
                        else:
                            vi = K8sVersionInfo()  # not sure what this will do for us actually
```

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,369 @@
...
+
+
+def is_current_version(version, k8s_version_list, cve_version_list=[]):
+    """Filter the versions depending on the usage times set by the Standard and
+    the times set for CVE versions.
+    """
+    try:
+        if diff_months(version.date, datetime.datetime.now()) >= MINOR_VERSION_CADENCE:
+            return False
+        else:
```
mbuechse, line 284: `else` after `return` is unnecessary; we can save a lot of indentation by omitting the `else`.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,369 @@
...
+
+The K8s version recency check returns 0 if the version of the tested cluster is still acceptable, otherwise
+it returns 1 for an out-of-date minor version, 2 for an out-of-date patch level version or 3 if the currently
+used version should be updated due to a highly critical CVE.
+
+    -t/--token GITHUB_TOKEN            - Github token used to check the Github API for the latest K8s releases
+    -k/--kubeconfig PATH/TO/KUBECONFIG - Path to the kubeconfig of the server we want to check
+    -h                                 - Output help
+    """)
+    return
```
mbuechse, line 78: `return` as the final instruction of a function is superfluous.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,369 @@
...
+    for opt in opts:
+        if opt[0] == "-h" or opt[0] == "--help":
+            print_usage()
+            raise HelpException
+        if opt[0] == "-c" or opt[0] == "--config":
+            config_path = opt[1]
+        if opt[0] == "-k" or opt[0] == "--kubeconfig":
+            config.kubeconfig = opt[1]
+
+    return config, config_path
```
mbuechse, line 102: I would add `config_path` to the `Config` class. Maybe in lieu of `log`, which seems to be unused.

cah-hbaum: Changed that up.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Remove the `return`

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Removed the `else`

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Changed that up.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: I did a big rewrite here (separating it into a new function), but I must say it's very bad that the people don't use the schema correctly (https://www.cve.org/AllResources/CveServices#cve-json-5 talks about their intentions). They don't have any checks for that, so that is their problem, but we need to parse the data so its also a problem for us.

We could now either a) try get all possible types of writing version information or b) just read the schema if it is conformant with the intended schema.

Any opinions on that?

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,369 @@
...
+            config_file = yaml.safe_load(fd)
+    except OSError:
+        pass
+
+    if config_file:
+        # Setup logging if the config file with the relevant information could be loaded before
+        setup_logging(config_file['log'])
+    else:
+        # Otherwise, we initialize basic logging
+        logging.basicConfig(format='%(levelname)s: %(message)s', level=logging.INFO)
```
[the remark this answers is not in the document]
cah-hbaum: Created a default dict, which is used if the config file isn't loaded.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,369 @@
...
+
+def initialize_config(config, config_path):
+    """Initialize the configuration for the test script
+    """
+
+    config_file = None
+    try:
+        with open(config_path, "r") as fd:
+            config_file = yaml.safe_load(fd)
+    except OSError:
```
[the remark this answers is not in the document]
cah-hbaum: Kept it, but integrated a warning.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,369 @@
...
+    try:
+        opts, args = getopt.gnu_getopt(argv, "t:k:h", ["token", "kubeconfig", "help"])
+    except getopt.GetoptError:
+        print_usage()
+        raise ConfigException
+
+    for opt in opts:
+        if opt[0] == "-h" or opt[0] == "--help":
+            print_usage()
+            raise HelpException
```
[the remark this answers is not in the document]
cah-hbaum: Moved the `print_usage` call.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,369 @@
...
+    """Parse cli arguments from the script call
+    """
+    config = Config()
+    config_path = "./config.yaml"
+
+    try:
+        opts, args = getopt.gnu_getopt(argv, "t:k:h", ["token", "kubeconfig", "help"])
+    except getopt.GetoptError:
+        print_usage()
+        raise ConfigException
```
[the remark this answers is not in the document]
cah-hbaum: Moved the `print_usage` call.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Forgot it here, changed it now.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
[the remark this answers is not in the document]
mbuechse: Okay, fine, but please use the `logger`, not least because it belongs on `stderr`.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,462 @@
...
+        except KeyError as e:
+            logger.debug(f"They key {e} couldn't be found in the CVE json data for CVE {cveid}.")
+            continue
+
+        is_high_severity = False
+        for cve_metric in cve_metrics:
+            for metric_key in cve_metric:
+                if re.search(r'[cC][vV][sS]{1,2}V\d', metric_key) and \
+                   cve_metric[metric_key]['baseScore'] >= CVE_SEVERITY:
+                    is_high_severity = True
```
mbuechse, line 341:
```suggestion
        is_high_severity = any(
           re.search(r'[cC][vV][sS]{1,2}V\d', metric_key) and
           metric_value['baseScore'] >= CVE_SEVERITY
           for cve_metric in cve_metrics
           for metric_key, metric_value in cve_metric.items()
       )
```

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,462 @@
...
+
+        if is_high_severity:
+            for aff in cve_affected:
+                if aff['product'] == "Kubernetes":
+                    for version_info in aff['versions']:
+                        if version_info['status'] == "affected":
+                            cvev = parse_cve_version_information(version_info)
+                            if cvev not in cfvs:
+                                cfvs.append(cvev)
```
mbuechse, line 350:
```suggestion
            affected_kubernetes_versions = [
                parse_cve_version_information(version_info)
                for aff in cve_affected
                if aff['product'] == "Kubernetes"
                for version_info in aff['versions']
                if version_info['status'] == "affected"
            ]
            for cvev in affected_kubernetes_versions:
                if cvev not in cfvs:
                    cfvs.append(cvev)
```

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
mbuechse: If that works, great. Then please remove any mention of the token from the usage and getopt stuff.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
mbuechse: That's indeed an unfortunate situation to be in, but I'm afraid we have to live with it. We can of course always ask Team Container what they think; maybe we are spending a lot of effort for something they don't deem important...

cah-hbaum: I'm gonna ask in the next container meeting.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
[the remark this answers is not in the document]
cah-hbaum: Put the logger there.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,369 @@
...
+CVE_VERSION_CADENCE = 3  # days
+CVE_SEVERITY = 8  # CRITICAL
+
+
+logger = logging.getLogger("k8s-version-recency-check")
+
+
+class ConfigException(BaseException):
+    """Exception raised in a configuration error occurs
+    """
```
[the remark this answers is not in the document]
cah-hbaum: Changed all one-liner doc strings to the mentioned format.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Changed it to the suggestion.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Changed it to the suggestion.

cah-hbaum:
As per suggestion of @jschoone, I looked into [Trivy](https://github.com/aquasecurity/trivy).
In general, it looks like a nice tool with many possibilities, which can provide an overview for all vulnerabilities.

The only two problems I see is that `trivy` doesn't provide an interface/API for other programs (at least I didn't find one during checking). Maybe their CI pipeline integration could be useful here, but that would need further investigation from my side.
If we then would need to use the output given by the `trivy` application, we would again need to parse this. But at least this would be standardized.

If someone else has opinions on this, I'm open for discussions/suggestions/etc.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,465 @@
...
+            if self.lower_version:
+                gt = self.lower_version <= version_info
+            else:
+                gt = True
+            if self.less_than:
+                if self.equal:
+                    return gt and self.upper_version >= version_info
+                return gt and self.upper_version > version_info
+        else:
+            return self.lower_version == version_info
```
fdobrovolny, line 236: I'm unfamiliar with the source data, but I'm just guessing I could be wrong. Should it be here lower or equal than just equal? Like if the CVE is not fixed in the next version?
```suggestion
            return self.lower_version <= version_info
```

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,465 @@
...
+    return CVEVersionInfo(vi_lower_version, vi_upper_version, less_than, equal)
+
+
+async def collect_cve_versions():
+    """Get all relevant CVE versions, that are relevant for the test according to the severity
+    dictated by the Standard.
+    """
+
+    # CVE fix versions
+    cfvs = list()
```
fdobrovolny, line 307: Just a suggestion, but wouldn't it make here more sense to have a set instead of a list? So you don't have to check for duplicate entries, and the lookup is faster?

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,465 @@
...
+import re
+import requests
+import sys
+import yaml
+
+
+MAJOR_VERSION_CADENCE = None
+MINOR_VERSION_CADENCE = 4  # months
+PATCH_VERSION_CADENCE = 1  # week
+CVE_VERSION_CADENCE = 3  # days
```
fdobrovolny, line 41: It would be great to have them all in the same units or rename them so the period is in the name itself.
```suggestion
MINOR_VERSION_CADENCE_MONTHS = 4
PATCH_VERSION_CADENCE_WEEKS = 1
CVE_VERSION_CADENCE_DAYS = 3
```

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,465 @@
...
+            config.logging = yaml.safe_load(f)['logging']
+    except OSError:
+        logger.warning(f"The config file under {config.config_path} couldn't be found, "
+                       f"falling back to the default config.")
+    finally:
+        # Setup logging if the config file with the relevant information could be loaded before
+        # Otherwise, we initialize logging with the included literal
+        setup_logging(config.logging or logging_config)
+
+    if not config.kubeconfig:
```
fdobrovolny, line 145: As mentioned below.

```suggestion
    if config.kubeconfig is None:
```

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,465 @@
...
+    config_path = "./config.yaml"
+    kubeconfig = None
+    logging = None
+
+
+def print_usage():
+    print("""
+K8s Version Recency Compliance Check
+
+Usage: k8s-version-recency-check.py [-h] [-k|--kubeconfig PATH/TO/KUBECONFIG]
```
fdobrovolny, line 86: If kubeconfig is a required parameter, it should not be here marked as optional.

Also as a suggestion for the future, I would consider using `argparse`.

> The [getopt](https://docs.python.org/3/library/getopt.html#module-getopt) module is a parser for command line options whose API is designed to be familiar to users of the C getopt() function. Users who are unfamiliar with the C getopt() function or who would like to write less code and get better help and error messages should consider using the [argparse](https://docs.python.org/3/library/argparse.html#module-argparse) module instead.

fdobrovolny (changes requested):

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,465 @@
...
+        version_api = kubernetes_asyncio.client.VersionApi(api)
+        ret = await version_api.get_code()
+
+        version = K8sVersionInfo.extract_version(ret.git_version.strip("v"))
+        version.date = datetime.datetime.strptime(ret.build_date, '%Y-%m-%dT%H:%M:%SZ')
+
+        return version, cluster_config.current_context['name']
+
+
+def is_current_version(version, k8s_version_list, cve_version_list=[]):
```
chess-knight, line 372: `cve_version_list=[]` default argument value should not be mutable

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,465 @@
...
+                   diff_days(datetime.datetime.now(), kv.date) >= CVE_VERSION_CADENCE:
+                    return False
+    except (KeyError, IndexError, TypeError) as e:
+        logger.debug(f"An error occurred during version filtering: {e}")
+        return False
+    else:
+        return True
+
+
+def collect_accepted_k8s_versions(cve_version_list=[]):
```
chess-knight, line 397: `cve_version_list=[]` default argument value should not be mutable

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
chess-knight, line 1: please add `chmod +x`

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,465 @@
...
+        self.date = None
+
+    def __eq__(self, other):
+        if not isinstance(other, K8sVersionInfo):
+            return False
+        return self.major == other.major and self.minor == other.minor and self.patch == other.patch
+
+    def __gt__(self, other):
+        if not isinstance(other, K8sVersionInfo):
+            return False
```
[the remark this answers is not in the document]
cah-hbaum: You're right, changed that to raise an exception.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,465 @@
...
+def diff_days(date1, date2):
+    delta = date1 - date2
+    return abs(delta.days)
+
+
+async def request_cve_data(session: aiohttp.ClientSession, cveid: str) -> dict:
+    """Request for a single CVE data item."""
+    resp = await session.request('GET', f"https://cveawg.mitre.org/api/cve/{cveid}",
+                                 headers={"Accept": "application/json"})
+    return await resp.json()
```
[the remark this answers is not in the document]
cah-hbaum: Updated to your suggestion.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,465 @@
...
+               version.patch < kv.patch:
+
+                if diff_weeks(datetime.datetime.now(), kv.date) >= PATCH_VERSION_CADENCE:
+                    return False
+
+                if kv in cve_version_list and \
+                   diff_days(datetime.datetime.now(), kv.date) >= CVE_VERSION_CADENCE:
+                    return False
+    except (KeyError, IndexError, TypeError) as e:
+        logger.debug(f"An error occurred during version filtering: {e}")
```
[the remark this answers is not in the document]
cah-hbaum: This section is now obsolete, since I changed the whole part how the versions got checked.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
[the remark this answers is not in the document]
cah-hbaum: You're correct, the month check was also wrong. I changed both up and in my testing, they both seemed to work correctly now.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,465 @@
...
+            return False
+        patchcomp = self.minor == other.minor and self.patch <= other.patch
+        return self.major < other.major or (self.major == other.major and (self.minor < other.minor or patchcomp))
+
+    @classmethod
+    def extract_version(cls, string, separator="."):
+        components = string.strip().split(separator)
+        return cls(int(components[0]), int(components[1]), int(components[2]))
+
+    def check_for_version(self, major=None, minor=None, patch=None):
```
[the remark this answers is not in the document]
cah-hbaum: Added a comment.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,465 @@
...
+        return (major is None or self.major == major) and \
+               (minor is None or self.minor == minor) and \
+               (patch is None or self.patch == patch)
+
+    def __str__(self):
+        return f"{self.major}.{self.minor}.{self.patch}"
+
+
+class CVEVersionInfo:
```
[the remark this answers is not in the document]
cah-hbaum: Added documentation.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,465 @@
...
+def parse_cve_version_information(cve_version_info):
+    """Parse the CVE version information according to their CVE JSON 5.0 schema"""
+    vi_lower_version = None
+    vi_upper_version = None
+    less_than = False
+    equal = False
+
+    # Extract the version if it is viable, but it's not a requirement
+    try:
+        vi_lower_version = K8sVersionInfo().extract_version(cve_version_info['version'].strip("v"))
```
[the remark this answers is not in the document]
cah-hbaum: Changed that up.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,465 @@
...
+            "formatter": "k8s-version-recency-check",
+            "stream": "ext://sys.stdout"
+        }
+    },
+    "root": {
+        "handlers": ["console"]
+    }
+}
+
+logger = logging.getLogger("k8s-version-recency-check")
```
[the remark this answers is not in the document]
cah-hbaum: I think I tried that and had some problems with that, but now it seems to work.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: No this is correct as it is, for more information, check out the description/video/documentation [here](https://www.cve.org/AllResources/CveServices#cve-json-5)

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: I had some problems with a set here, since the `CVEVersionInfo` class is unhashable. I could probably make it hashable, but using a list and comparing seemed like the easier solution to me.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Changed the names up.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Changed that up.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: We can use `argparse` next time, I just assumed `getopt` was anticipated, since previous scripts also used it.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Changed it to use a non-mutable type and set an empty list later.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Changed it to use a non-mutable type and set an empty list later.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Done.

cah-hbaum:
I think, I fixed/changed most of the problems/things found, please provide more feedback if anymore problems arise. Thank you for your contribution.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
chess-knight: I still get:
```bash
$ ./k8s-version-recency-check.py
bash: ./k8s-version-recency-check.py: Permission denied
```

cah-hbaum: Interesting, for some reason it reset itself.
I pushed again, now its definitely there.

chess-knight: Thank you

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -83,13 +83,13 @@ def print_usage():
     print("""
 K8s Version Recency Compliance Check
 
-Usage: k8s-version-recency-check.py [-h] [-k|--kubeconfig PATH/TO/KUBECONFIG]
+Usage: k8s-version-recency-check.py [-h] [-c|--config PATH/TO/CONFIG] -k|--kubeconfig PATH/TO/KUBECONFIG
 
 The K8s version recency check returns 0 if the version of the tested cluster is still acceptable, otherwise
 it returns 1 for an out-of-date minor version, 2 for an out-of-date patch level version or 3 if the currently
 used version should be updated due to a highly critical CVE.
```
chess-knight: For me, it is a little bit different as you write in header:
```
Return codes:
0: Version is inside the recency window
1: Error during script execution
2: Version isn't inside the recency windows anymore
3: Version used contains a critical CVE
```

Standards/scs-0412-v1-metering-json.md
chess-knight, line 1: Should this document be included in this PR?

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Yeah you're right, that wasn't updated quit some time.

Standards/scs-0412-v1-metering-json.md
cah-hbaum: No this came from the merge done by @jschoone.
I rebased the branch now, so the PR now contains only the files we expect.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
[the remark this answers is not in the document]
mbuechse: @fdobrovolny  Sorry for interjecting here, but out of curiosity, what is the difference?

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,465 @@
...
+        less_than = True
+        equal = True
+    elif 'lessThan' in cve_version_info:
+        vi_upper_version = K8sVersionInfo().extract_version(cve_version_info['lessThan'].strip("v"))
+        less_than = True
+
+    # This shouldn't happen, but if it happens, we look for non-standard descriptions
+    # According to this(https://www.cve.org/AllResources/CveServices#cve-json-5),
+    # this isn't how the data should be described
+    if not vi_lower_version and not vi_upper_version:
```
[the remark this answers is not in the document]
mbuechse: Is this a check for None though? Wouldn't any falsy value be equally bad here?

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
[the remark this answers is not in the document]
mbuechse: While it is indeed often the case that `__name__` is used, I think it is not a must, and less so for stand-alone programs.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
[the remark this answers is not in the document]
fdobrovolny: @mbuechse 

I think it is checking for if the value was not set and therefore is None as set in default.

```
>>> bool(0)
False
```

If the upper version would be integer zero, it would not work.

But it is a practice I use in line with [PEP8](https://peps.python.org/pep-0008/#programming-recommendations), as this is more explicit and sometimes when the code changes down the line it can lead to weird bugs when a falsey value triggers something that they should not. But in this specific code, it would be nicer to handle this completely differently as the extract version can return None and it could be a wanted behaviour, I just wanted to propose more minimal usable solution.

> Also, beware of writing if x when you really mean if x is not None – e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
fdobrovolny: I was not aware of that. Then it is probably better that it is consistent with the rest. However it might be nice to ask in Element and discuss what is better approach next time. If you ping me I can brought it up, I think it would be worthy.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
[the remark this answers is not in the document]
fdobrovolny: I think it does some logic to close the connection as the `with` operator calls "`open`" and "`close`" on the object.

I was just going out of the documentation and how it is intended to be used, never questioning it. If you want to take a look at the source code, I would also be interested.

https://docs.aiohttp.org/en/stable/client_quickstart.html#make-a-request

> Now, let’s try to get a web-page. For example let’s query http://httpbin.org/get:
> 
> ```
> async def main():
>     async with aiohttp.ClientSession() as session:
>         async with session.get('http://httpbin.org/get') as resp:
>             print(resp.status)
>             print(await resp.text())
> 
> asyncio.run(main())
> ```

> To make several requests to the same site more simple, the parameter base_url of [ClientSession](https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientSession) constructor can be used. For example to request different endpoints of http://httpbin.org can be used the following code:
> ```
> async with aiohttp.ClientSession('http://httpbin.org') as session:
>     async with session.get('/get'):
>         pass
>     async with session.post('/post', data=b'data'):
>         pass
>     async with session.put('/put', data=b'data'):
>         pass
> ```

> A session context manager usage is not mandatory but await session.close() method should be called in this case, e.g.:
> 
> ```
> session = aiohttp.ClientSession()
> async with session.get('...'):
>     # ...
> await session.close()
> ```

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,464 @@
...
+        return (major is None or self.major == major) and \
+               (minor is None or self.minor == minor) and \
+               (patch is None or self.patch == patch)
+
+    def __str__(self):
+        return f"{self.major}.{self.minor}.{self.patch}"
+
+
+class CVEVersionInfo:
+    """Class that contains a CVE version info.<
```
fdobrovolny, line 216:
```suggestion
    """Class that contains a CVE version info.
```

typo

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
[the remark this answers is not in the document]
fdobrovolny: It is not a must. I just don't see much reason why not as even in [docs](https://docs.python.org/3/library/logging.html) it is a recommended way. If the script name changes, it gets updated too. 

> The logger name hierarchy is analogous to the Python package hierarchy, and identical to it if you organise your loggers on a per-module basis using the recommended construction `logging.getLogger(__name__)`. That’s because in a module, `__name__` is the module’s name in the Python package namespace.

However, it was meant as a suggestion and definitely would not have anything against keeping it there like this.

fdobrovolny (approved):

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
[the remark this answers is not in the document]
mbuechse: Interesting that the quickstart docs should recommend that, because the [reference](https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponse.release) also states

> It is not required to call release on the response object. When the client fully receives the payload, the underlying connection automatically returns back to pool. If the payload is not fully read, the connection is closed

So it seems to me that the `async with` is kind of redundant for us, because we do consume the payload (by calling `json`).

At any rate, it has nothing to do with being async, and more with returning the resources (the connection in this case) to the pool. It may be more economical with the context manager (though, as mentioned, not in our case), but it's not "more async".

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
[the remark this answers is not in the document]
mbuechse: You are right. I know and support PEP-8, I just wasn't aware that the value in question was an integer. (I guess it's obvious in the code, but Github doesn't provide much in the way of context.) In this case, we must explicitly check for None!

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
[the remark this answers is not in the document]
fdobrovolny: Indeed, it is interesting. I find some philosophical value in it as well, as it promotes a consistent approach that always works, aligning with The Zen of Python principles.

I meant "properly async" to suggest that either the python `requests` library should be used correctly or the `aiohttp` library should be adopted fully. At the time, it appeared to be a partial transition to async(`aiohttp`) based on the quickstart and my experience, hence the term "...properly async it should look more like this." However, since it was functioning, it was meant as a suggestion rather than a requirement.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Fixed the typo.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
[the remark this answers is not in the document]
cah-hbaum: Can this be resolved?

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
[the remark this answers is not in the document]
cah-hbaum: Added the suggestion.

fdobrovolny (approved):

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Added the ClientSession as an argument. Also tried to follow the `requests`/`aiohttp` advise.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
cah-hbaum: Created a list comprehension.
I also bound the number of simultaneous requests to 5.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,369 @@
...
+
+    if config_file:
+        # Setup logging if the config file with the relevant information could be loaded before
+        setup_logging(config_file['log'])
+    else:
+        # Otherwise, we initialize basic logging
+        logging.basicConfig(format='%(levelname)s: %(message)s', level=logging.INFO)
+
+    if not config.kubeconfig:
+        print("A kubeconfig needs to be set in order to test a k8s cluster version.")
```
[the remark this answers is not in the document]
cah-hbaum: Wrote the error message in the Exception and printed it out in the `except` clause with the logger.

Tests/kaas/k8s-version-recency/k8s-version-recency-check.py
```diff
@@ -0,0 +1,472 @@
...
+            # under the metrics list.
+        except KeyError as e:
+            try:
+                logger.debug(
+                    f"They key {e} couldn't be found in the CVE json data for CVE {cve_data['cveMetadata']['cveId']}."
+                )
+            except KeyError:
+                logger.debug(
+                    f"They key {e} couldn't be found in the CVE json data for a CVE."
+                )
```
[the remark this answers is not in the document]
mbuechse: I advocate for this change, but I don't like the Python 3.8+ version very much. What we could also do:

```suggestion
            logger.debug(
                f"They key {e} couldn't be found in the CVE json data for CVE {cve_data.get('cveMetadata', {}).get('cveId', '<ID NOT SET>')}."
            )
```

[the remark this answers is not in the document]
cah-hbaum: Took the version from @mbuechse, but also thanks for the other suggestions.

mbuechse (approved):

PR #319, opened 2023-07-03
Decouple standard flavors from flavor naming

### LINKED ISSUE #267: Split flavor naming and mandatory flavors in two separate standards
Flavor naming and mandatory flavors are currently defined in one standard, called [flavor naming](https://github.com/SovereignCloudStack/standards/blob/main/Drafts/flavor-naming.md), but should be separated.

mbuechse:
Resolves #267


Standards/scs-0103-v1-standard-flavors.md
```diff
@@ -0,0 +1,134 @@
...
+| SCS-8V-32        |      8 | shared core   |         32 |                 |            |
+| SCS-1V-2         |      1 | shared core   |          2 |                 |            |
+| SCS-2V-4         |      2 | shared core   |          4 |                 |            |
+| SCS-4V-8         |      4 | shared core   |          8 |                 |            |
+| SCS-8V-16        |      8 | shared core   |         16 |                 |            |
+| SCS-16V-32       |     16 | shared core   |         32 |                 |            |
+| SCS-1V-8         |      1 | shared core   |          8 |                 |            |
+| SCS-2V-16        |      2 | shared core   |         16 |                 |            |
+| SCS-4V-32        |      4 | shared core   |         32 |                 |            |
+| SCS-1L-1         |      1 | crowded core  |          1 |                 |            |
```
mbuechse, line 39: **TODO**: Resolve the following issues:

- We say what resources the flavors should offer. We don't say what resources should NOT be offered. For instance, none of the flavors explicitly offers a GPU, but we don't disallow a GPU either. However, for a customer, this may make a big difference, not least with regards to pricing. (**Note that this issue exists with the original standard as well.**)
- We say that the mandatory and recommended flavors must come with certain `extra_specs`. We have to say something about the converse: which (other) flavors may carry (any, some subset, or all of) these `extra_specs`. For the time being, I suppose we require that the presence of any one of these properties imply that the flavor is one of those listed here (and consequently, the other properties must be present as well). In the medium run, we have to define in general which of these properties are compulsory for any flavor.

Standards/scs-0103-v1-standard-flavors.md
```diff
@@ -0,0 +1,134 @@
...
+| SCS-2V-16-50     |      2 | shared core   |         16 |              50 | (any)      |
+| SCS-4V-32-100    |      4 | shared core   |         32 |             100 | (any)      |
+| SCS-1L-1-5       |      1 | crowded core  |          1 |               5 | (any)      |
+
+### Guarantees and properties
+
+The following guarantees must be met:
+
+- A "shared core" means _at least 20% of a core in >99% of the time_, measured over the
+  course of one month (1% is 7,2 h/month).
```
mbuechse, line 66: We had a discussion within our team, where we wondered the following: Isn't the question here, **how much is 100 %**?

There could be a big difference between `SCS-1V-4-10` on one cluster and `SCS-1V-4-10` on another, depending on the model of the CPU, even if the sharing is the same. In this vein, wouldn't it be better for the customer to know what to expect in terms of computing power?

Maybe we could introduce performance classes with respect to TFLOPs or something similar, and then x amount of TFLOPs (what the customer needs) is a function of both the CPU model and the sharing model (what the customer needn't care about).

Standards/scs-0103-v1-standard-flavors.md
artificial-intelligence: > We say that the mandatory and recommended flavors must come with certain extra_specs. We have to say something about the converse: which (other) flavors may carry (any, some subset, or all of) these extra_specs.

I disagree, we don't have to say anything about the converse, because there is an infinite number of combinations of flavors and extra_specs. We only define and standardise what we define, everything else is up to the CSP.

mbuechse: @artificial-intelligence The problem is that we want discoverability. People want to find their flavor by looking at extra_specs. So when some flavor carries these extra_specs, that has to mean something!

Standards/scs-0100-v4-flavor-naming.md
```diff
@@ -0,0 +1,521 @@
...
+Note that all letters are case-sensitive.
+
+Typical flavor names look like `SCS-4V-16-50` for a flavor with 4vCPUs (with limited
+oversubscription), 16GiB RAM and a 50GB disk (of unspecified type).
+
+## Proposal Details
+
+### [REQUIRED] CPU Suffixes
+
+Next to the number of vCPUs, these vCPUs need to be characterized ti describe their nature.
```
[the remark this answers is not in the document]
mbuechse: I will incorporate typo corrections such as this one. Other discussions should be postponed, because this PR is **only meant for splitting** the document. (Be aware that all  your remarks apply to the original document.)

Standards/scs-0100-v4-flavor-naming.md
```diff
@@ -0,0 +1,521 @@
...
+CPU types and generations, high-frequency models, GPU support and GPU types as
+well as Infiniband support. (More extensions may be appended in the future.)
+
+Using the systematic naming approach ensures that two providers that offer flavors
+with the same specific features will use the same name for them, thus simplifying
+life for their customers when consuming these flavors.
+
+Note that there is no need to indicate all details and extra features this way.
+Flavors may always perform better or have more features than indicated in a name.
+Underperformance (CPU suffices `L` or `i` or memory suffices `o` and `u`) on the other
```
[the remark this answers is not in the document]
mbuechse: I guess the correction should be "suffixes".

Standards/scs-0100-v4-flavor-naming.md
```diff
@@ -0,0 +1,521 @@
...
+NAMEV2=$(echo "$NAMEV1" | sed -e 's/\-/_/g' -e 's/:/-/g' -e 's/^SCS_/SCS-/')
+```
+
+and the way back can be done with
+
+```shell
+NAMEV1=$(echo "$NAMEV2" | sed -e 's/\-/:/g' -e 's/_/-/g' -e 's/^SCS:/SCS-/')
+```
+
+Considerations for how providers can ensure a smooth transition for their customers
```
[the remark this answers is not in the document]
mbuechse: --> <https://github.com/SovereignCloudStack/standards/issues/327>

Standards/scs-0100-v4-flavor-naming.md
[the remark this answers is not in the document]
mbuechse: --> <https://github.com/SovereignCloudStack/standards/issues/327>

Standards/scs-0100-v4-flavor-naming.md
```diff
@@ -0,0 +1,521 @@
...
+- SCS-2C-4-3x- <- Cloud decides disk type and size and creates three of them (FIXME: Is this useful?)
+- SCS-2C-4-3xs <- Cloud decides size and creates three local SSD volumes (FIXME: useful?)
+- SCS-2C-4-3x10 <- Cloud decides type and creates three 10GB volumes
+- ~~SCS-2C-4-**1.5n**~~ <- You must not specify disk sizes which are not in full GiBs
+
+## Naming policy compliance
+
+To be certified as an SCS compliant x86-64 IaaS platform, you must offer all standard mandatory SCS
+flavors according to the previous section. (We may define a mechanism that allows exceptions to be
+granted in a way that makes this very transparent and visible to clients.)
```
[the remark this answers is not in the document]
mbuechse: --> <https://github.com/SovereignCloudStack/standards/issues/327>

Standards/scs-0100-v4-flavor-naming.md
```diff
@@ -0,0 +1,521 @@
...
+- SCS-2C-4-**n** <- Cloud decides disk size (min_disk from image or larger)
+- SCS-2C-4-**n**\_bms_3
+- SCS-2C-4- <- Cloud decides disk type and size
+- SCS-2C-4-\_bms_z3
+- SCS-2C-4-\_bms_z3h_GNa-64_ib
+- SCS-2C-4-\_ib
+- SCS-2C-4 <- You need to specify a boot volume yourself (boot from volume, or use `block_device_mapping_v2`)
+- SCS-2C-4_bms_z3
+- SCS-2C-4-3x- <- Cloud decides disk type and size and creates three of them (FIXME: Is this useful?)
+- SCS-2C-4-3xs <- Cloud decides size and creates three local SSD volumes (FIXME: useful?)
```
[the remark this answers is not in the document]
mbuechse: --> <https://github.com/SovereignCloudStack/standards/issues/327>

Standards/scs-0100-v4-flavor-naming.md
```diff
@@ -0,0 +1,521 @@
...
+#### Higher oversubscription
+
+Must be indicated with a `L` vCPU type (low performance for > 5x/core or > 3x/thread oversubscription and
+the lack of workload management that would prevent worst case performance <20% in more than 7.2h per month).
+
+#### Insufficient microcode
+
+Not using these mitigations must be indicated by an additional `i` suffix for insecure
+(weak protection against CPU vulns through insufficient microcode, lack of disabled hyperthreading
+on L1TF susceptible CPUs w/o effective core scheduling or disabled protections on the host/hypervisor).
```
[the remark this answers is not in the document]
mbuechse: --> <https://github.com/SovereignCloudStack/standards/issues/327>

Standards/scs-0100-v4-flavor-naming.md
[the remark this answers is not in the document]
mbuechse: --> <https://github.com/SovereignCloudStack/standards/issues/327>

Standards/scs-0103-v1-standard-flavors.md
```diff
@@ -0,0 +1,145 @@
...
+to have a guaranteed set of flavors available on all SCS clouds, so these need not be discovered.
+
+## Properties (extra specs)
+
+The following extra specs are recognized, together with the respective semantics:
+
+- `scs:name-vN=NAME` (where `N` is `1` or `2`, and `NAME` is some string) means that the
+  flavor is one of the
+  standard SCS flavors, and the requirements of Section "Standard SCS flavors" below apply.
+- `scs:cpu-type=shared-core` means that _at least 20% of a core in >99% of the time_,
```
garloff, line 26: Don't we need to provide the full list of allowable and standardized options here?
I.e. `crowded-core` (corresponding to an `L` encoded in the name), `shared-core` (`V), `dedicated-thread` (`T`), or `dedicated-core` (`C`)?

Standards/scs-0103-v1-standard-flavors.md
```diff
@@ -0,0 +1,145 @@
...
+  measured over the course of one month (1% is 7,2 h/month).
+- `scs:diskN-type=ssd` (where `N` is a nonnegative integer, usually `0`) means that the
```
garloff, line 28: Likewise, we should spell out the allowable `disk`N`-type` settings, no?
`network` (`n`), `hdd` (`h`), `ssd` (`s`), `nvme` (`p`) and also mention that is is allowable to advertize NVMes as SSDs (as understatement is allowable).

We could of course reference another standard here, if we prefer so.
(And as we have not yet a complete list of extra specs, maybe we do two sets of them to have the first one ready quickly?)

Tests/iaas/scs-0103-v1-flavors.yaml
```diff
@@ -0,0 +1,192 @@
+meta:
+  name_key: name-v2
+flavor_groups:
+  - status: mandatory
+    list:
+    - name: SCS-1V-4
```
garloff, line 6: One of the design principles in the old way of listing Mandatory (and Recommended) flavors was to NOT keep long lists of settings in a manually managed YAML file.
Instead we only had the names and had code that derived all settings from the name to generate the verbose form. This had the big advantage to keep the human-managed data short and avoid any divergence between the settings encoded in the file. I would consider a list here with just names. The amount of `cpus`, `ram` as well as `disk1 and `cpu-type` and `name-v1` and `name-v2` should be extracted from the name and generated by code.

The other option would be to have tooling that is included in our CI that validates the input for consistency. I would consider it less preferable over the condensed format, but would not veto it, as the ones who do the main work obviously have to feel comfortable with the design and tooling.

Tests/testing/scs-compatible-test.yaml
```diff
@@ -1,15 +1,37 @@
...
+      - name: Standard flavors
+        url: https://raw.githubusercontent.com/SovereignCloudStack/standards/main/Standards/scs-0103-v1-standard-flavors.md
+        check_tools:
+          - executable: ./iaas/standard-flavors/flavors-openstack.py
+            args: "./iaas/scs-0103-v1-flavors.yaml"
+      - name: Flavor naming
+        url: https://raw.githubusercontent.com/SovereignCloudStack/standards/main/Standards/scs-0100-v3-flavor-naming.md
+        check_tools:
+          - executable: ./iaas/flavor-naming/flavor-names-openstack.py
+            args: "--mand=./iaas/scs-0100-v3-flavors.yaml"
```
garloff, line 15: So we call the script twice and it only tests one of the aspects each time?
0103: Is the list of mandatory SCS flavors complete?
0100: Are the names starting with `SCS-` all according to the syntax rules and do the discoverable properties match what the names claim?
The scripts has previously done both -- separating concerns here is good.
(Obviously in CI testing, cycle time is an important goal, so running the script only once doing both tests is an option if cycle time would be heavily impacted. This test is rather fast, so the cleaner separation of concerns wins over performance optimization ...)

garloff (commented):
Great work overall!
I would still like the comments to be considered and responded to, so we get the best out of this ...

fkr:
ping @frosty-geek @jnull - please review (since you belong to the 'crowd of the CSPs')

Standards/scs-0103-v1-standard-flavors.md
mbuechse: This standard is about standard flavors. Therefore, I only listed the values that we need for the standard flavors.

Standards/scs-0103-v1-standard-flavors.md
mbuechse: See above. Yes, we can go ahead and standardize the extra specs fully, but I would rather do one step at a time. What we need is a proof of concept that discoverability in this way does indeed work.

Tests/testing/scs-compatible-test.yaml
mbuechse: Right, see https://github.com/SovereignCloudStack/standards/pull/319#discussion_r1302938346

(except we are talking two separate scripts, not one script being called twice)

Tests/iaas/scs-0103-v1-flavors.yaml
artificial-intelligence: fwiw I tend to agree with @mbuechse here, because I find it cleaner to separate code - which does the checks - and data - the flavors to be checked -.

As a middle ground to keep the line count lower I'd suggest the usage of yaml anchors and aliases like this:

```yaml
  - foo: &SCS-2V
    - name: SCS-2V-8
      cpus: 2
      cpu-type: shared-core
      ram: 8
  - foo: *SCS-2V
    - name: SCS-2V-4-20s
      ram: 4
      disk: 20
  - foo: *SCS-2V
    - name: SCS-2V-16
      ram: 16
```

artificial-intelligence (approved):
LGTM

as said in the comment, I guess we can keep the yaml line count lower with some clever usage of yaml anchors and aliases, I just made up a random example, there may be very likely even better schemes.

fkr (approved):

Tests/iaas/scs-0103-v1-flavors.yaml
garloff: > I strongly oppose deriving the metadata from the name. One of the main principles behind my approach is separation of concerns, that is, avoiding coupling where possible:
> 
> - From the point of view of this standard, the names are opaque/arbitrary.
> - It is up to the naming standard to check whether the properties and the name are consistent. (Which the script for that standard actually does, except for the extra_specs.)

Fair enough.
Some consistency checks should be done as CI job then I believe, because invariably we will screw up sooner or later otherwise.

garloff: And I'm fine with pushing this consistency check out into a new issue, so we can get this PR merged now.

Standards/scs-0103-v1-standard-flavors.md
garloff: OK, we discussed this in the Team IAM call and I pushed https://github.com/SovereignCloudStack/standards/pull/319/commits/074b12a7102c74a47db74609456a3389c5348589 as a consequence.
Rationale: Without having the full extra_specs discoverability/transparency thing defined, we have introduced the extra_specs `scs:name-v1`, `scs:name-v2`, `scs:cpu-type` and `scs:diskN-type` here on the fly. If we do so, let's cover them completely and leave all other extra_specs in a future standard. This avoids confusion for the reader and avoids providers doing wild things that are hard to change once they are out in the wild.

mbuechse: I don't disagree, but I strongly oppose the breaking of due process here. It takes you a full week to write your comment, and then it can't happen quickly enough, so you push the change yourself. This is very disturbing.

Standards/scs-0103-v1-standard-flavors.md
garloff: I never meant to do all of the `extra_specs` definitions that we would want.
But the ones that we touch anyway (`scs:name-v1`, `scs-name-v2`, `scs:cpu-type` and `scs:diskN-type`) feel half-defined, so let's have those complete. And do all the others in a future step.

mbuechse: Yeah, that comment of mine is obsolete by now. I understand your reasoning, and I don't oppose it.

Standards/scs-0103-v1-standard-flavors.md
mbuechse: I wouldn't have minded a direct contribution if the back and forth hadn't been as sluggish. I had a feeling of being left out, no real cooperation happening, and your commit to the branch was only adding to that feeling.

PR #320, opened 2023-07-04
Create glossary.md

alexander-diab:
first draft of a glossary explaining the key terms in SCS certification

### DIFF
--- a/glossary.md
+++ b/glossary.md
@@ -1,0 +1,25 @@
+Standard: 
+A standard in SCS is a definition of how a certain cloud computing stack is to be configured or operated or functioning in order to be SCS compliant. Each standard is described in an architectural decision record (ADR). 
+
+Layer: 
+An SCS standard refer to either one of two different Layers - Iaas (Infrastructur as a Service) or KaaS (Kubernetes as a Service), which are independent of each other. 
+
+Certification bundle (pool): 
+Several single standards for a certain layer build a certification bundle. In order to award an SCS certification for a certain Layer ALL standards in the certification bundle must be verifiably implemented in the target cloud environment. 
+
+Version:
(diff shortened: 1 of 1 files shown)


mbuechse (commented):
Thanks for providing this overview! It certainly gets me thinking. I have a few remarks.

- Are layer and level really orthogonal, or is the concept of a layer mainly applicable with the first level (SCS compatible)? My impression is that we should focus on SCS compatible and drop the whole concept of levels for the time being, until we know what the next level could look like and which of the concepts make sense on the next level.
- As far as I understand, a standard is not linked to any layer per se, but only to a track (an SCS-internal way of organising things). It is only the certification bundle that associates the standards it consists of with a layer.
- Also (as far as I understand), "standard" and "ADR" are two distinct types of documents (see scs-0001).
- I would say that the most important point to make about the concept of a standard from the point of view of certification: we just want to make the description of certification requirements modular, so that it's easier to work on independent things independently and to track what changes from one version to the next; therefore the certification requirements are split into multiple documents, and these documents we call standards. We could also call them certification components.
- I'm not sure if "SCS compliant" is a good term. I would say the cloud layer X (IaaS or CaaS) is SCS compatible (if and only if it has the certificate for any valid version of SCS-compatible).

maxwolfs:
I thought adding the definitions to the already existing glossary [here](https://github.com/SovereignCloudStack/docs/blob/main/glossary.md).

fkr:
> * Also (as far as I understand), "standard" and "ADR" are two distinct types of documents (see scs-0001).

Correct. The types are 'Standard', 'Decision Record' and 'Procedural'. 
See https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0001-v1-sovereign-cloud-standards.md#types-of-documents

Issue #265, opened 2023-04-21
Define Standard for the naming of volume types


josephineSei:
Volumes are commonly used either as a root disk for a VM or to attach to a VM for additional disk space. Plus they are persistent, even after a VM is deleted.
There are multiple storage solutions for volumes and also the performance and whether they are encrypted or not can vary. These differences and a few other things are defined in the volume type. However, there is always a default volume type which will be used and can vary. And there is no API endpoint in OpenStack so far, that can show you all defined properties of a volume type. Uses right now can only look at the names of the types and have to decide only with this in mind, which type they choose for their volumes.

Therefore we want to define a standard for the naming of volume types. The names should include performance aspects and optional encryption type and maybe other properties.

- [x] Find consensus about the properties, which should be stated in the volume type name
- [ ] Write Standard for the naming of volume types

garloff:
I guess next to performance (bandwidth, latency) and encryption, users want to have an indication of redundancy/availability. In terms of "survives a disk crash, survives a node crash, survives a rack going down, survives an AZ going down" or so. (Maybe it's just two classes: Survives single disk/node failure and survives AZ failure.)
Speaking of AZs: An indication whether the volume is local to an AZ or available in all AZs of the region is also an important distinction that should be discoverable. (Not necessarily in the name.)

fkr:
@josephineSei thanks for the write up. I've added it to next weeks IaaS call's agenda (https://input.scs.community/2023-scs-team-iaas?view#Define-Standard-for-the-naming-of-volume-types) - are you going to be there for the call?

Upcoming week is Sprint Review & Planning, however I'll make sure we have a timebox of 10 Minutes for the topic - if needed we can breakout from there.

josephineSei:
@fkr i will attend the next meeting

artificial-intelligence:
when/how will the breakout session happen?

josephineSei:
I would like to have a breakout session this week. 

 @artificial-intelligence , @ppkuschy , @frosty-geek , @jnull , @berendtcon , @fkr : Which of the following slots could you attend?
 
 Wednesday: 9:05 - 10:00  or 11:05 - 12:00  
 Thursday: 9:00 - 10:00 or 10:00 - 11:00
 Friday: 9:00 - 10:00 or 10:00 - 11:00

artificial-intelligence:
all suggested timeslots are fine for me :+1:

jnull:
[...]

PR #323, opened 2023-07-12
Add document for the volume type decisions

josephineSei:
This document describes all options, that are configurable in volume types. It also shows our decisions and a few still open questions. This will be the base for the volume type standard.


josephineSei:
This is part of ticket #265

Standards/scs-0111-v1-volume-type-decisions.md
```diff
@@ -0,0 +1,90 @@
+---
+title: Decisions for the Volume Type Standard
+type: Decision Record
+status: Draft
+track: IaaS
+---
+
+## Introduction
+
+Volume Types are used to classify volumes and provide a basic decision for what kind of volume should be created. These volume types can sometimes very be backend-specific and it might be hard for a user to choose the most suitable volume type, if there is more than one default type.
```
berendt, line 10: We should add a definition of what a "volume" is from our point of view.

Standards/scs-0111-v1-volume-type-decisions.md
```diff
@@ -0,0 +1,90 @@
...
+
+2. Volume Types that can be attached to multiple Nova Azs: This option can be seen in the extra specs of a volume type also by normal users. Another option is to use backend specific options, as for example with ceph that directly interacts with nova for this. In that case there will not be any visible extra specs for the users.
+
+Another question is how many providers use one of these options or both.
+
+**Conclusion**: The first part doesn't make much sense to standardize, as migration between the volume types can only be done by admins. However the second part might be noteable, but due to the variety of configuration options very hard to standardize.
+
+#### Multiattach
+
+It is possible in a few backends to attach a volume to multiple VMs. This hat to be configured in the Volume Type and this information is also accessable for users. Nevertheless this option also needs a lot of work from users, as those types of volumes have to have a file system, that is capable of multiattach. Many providers don't provide multiattach.
```
berendt, line 54: What about Cross Attach AZ?

Standards/scs-0111-v1-volume-type-decisions.md
```diff
@@ -0,0 +1,90 @@
...
+
+**Conclusion**: It might be noteable, that this already is a discoverable option.
+
+#### Replication
+
+Replication states, whether or not there are multiple replicas of a volume. Thus answers the question, whether the data could survive a node outage. Again there are different ways to achive replicated volumes. It can either be defined in the volume type and is discoverable also by normal users or it is configured in the backend. The last option is usually used with ceph for example. This makes it hard to discover, whether a volume is replicated or not. Another point is the number of replicas, that exist.
```
berendt, line 60: How to handle Erasure Coding?

Standards/scs-0111-v1-volume-type-decisions.md
```diff
@@ -0,0 +1,90 @@
...
+
+**Conclusion**: This should not be standardized.
+
+#### Availability Zones
+
+Availability Zones are used in Nova and Cinder seperatly to provide an often also physical separation of compute hosts or storage nodes. This leads to two options to consider:
+
+1. Multiple Volume AZs: This might be used if there are different backends present in one IaaS structure. The different volume types are usually used for the different volume AZs. This makes migration between those AZs only be possible for administrators.
+
+2. Volume Types that can be attached to multiple Nova Azs: This option can be seen in the extra specs of a volume type also by normal users. Another option is to use backend specific options, as for example with ceph that directly interacts with nova for this. In that case there will not be any visible extra specs for the users.
```
berendt, line 46: Is this Cross AZ Attach?

Standards/scs-0111-v1-volume-type-decisions.md
```diff
@@ -0,0 +1,90 @@
...
+
+**Conclusion**: Replication is a good option to be standardized. Whether this should be done as a boolean option or if the number of replicas is also something users need to know should still be discussed. Nevertheless due to the different options to configure replication this will be quite complex.
+
+#### QoS
+
+Quality of Service parameters can be stated in a volume qos object. These objects can then be associated to a volume type (or directly to a volume as an admin only option). But this is optional and thus even good or very good volume QoS parameters that are aquired through hardware configuration and storage parameters, might go by unmentioned.
+Furthermore the indirection makes it harder to discover the qos for a volume type. Only admins will see the associated qos ID and will have to take a closer look at the qos after discovering the volume type. PLUS: there can only be one qos association for one volume type. But a qos can be used for multiple volumes.
+
+**Conclusion**: The benefit of displaying qos parameters is clear, thus this option should be noted. But are volume qos objects widely used? If not, standardization process would be too much work.
+
```
berendt, line 70: What about things like latency?

Standards/scs-0111-v1-volume-type-decisions.md
```diff
@@ -0,0 +1,90 @@
...
+| Aspect | Standardize? | Discoverability | other Things |
+| ---- | ---- | ---- | ------ |
+| encryption | **Recommended** | work needed | extra_spec: encrypted=True/False |
+| Backend name | - | - | - |
+| AZs | - | - | describe as optional and backend-dependend |
+| multiattach | - | yes | describe as optional |
+| Replication | **Recommended** | lot of work | either get from backend to OS or as extra_spec defined by deployer |
+| Number of Replicas, etc | ? | lot of work | optional, work on it after Replication is standardized |
+| Volume QoS | ? | admin only | needs further discussion, should be at least described as optional |
+
```
berendt, line 87: Should storage backend features (like snapshotting, CoW, ..) be visible?

berendt: Should it be possible to have something for the purpose of the storage backend? Read-optimized, write-optmized, ...

fkr:
Due to PTO this is on hold. Of course it is welcome to be further reviewed and commented, finalisation will happen once the participants is back.

Standards/scs-0111-v1-volume-type-decisions.md
josephineSei: What excactly do you mean with cross AZ, especially in this part? I tried to describe, what is possible to configure and what is possible to see for users. This does not match excactly. And we have two types of AZ that to me are quite different. A) Volumes can reside in different Volume AZs e.g. mapped to different backends. B) Volumes can be attached to VMs in specific AZs, that is configured for Ceph in Ceph and Nova.

Standards/scs-0111-v1-volume-type-decisions.md
josephineSei: There is not much documentation about volume qos - https://bugs.launchpad.net/cinder/+bug/1785735
It seems latency is not something, that can be defined in ths volume qos object. Also latency does really depend on the physical parameters of the deployment. 
So there are not many more thing to describe here.

Standards/scs-0111-v1-volume-type-decisions.md
fkr: @josephineSei Reading over this and the comment from @berendt: In the one session we did on this (6th of July) we did discuss this iirc (since in that session @jnull raised this question).

josephineSei:
@berendt @fkr i added a section with "other options" to conclude, that backend specific but notable options should ideally be noted in the description by the deployers themself. Are you okay with that? It would solve many discussion points.

fkr:
It was discussed that more and explicit feedback from operators is desired.
As such @jnull and @maliblatt were added as reviewers. Please review in the course of the next six days.

Standards/scs-0111-v1-volume-type-decisions.md
```diff
@@ -0,0 +1,96 @@
...
...y the storage service Cinder, which abstracts creation and usage of many different storage backends. While it is possible to use a backend like lvm which can reside on the sam host as the hypervisor, the SCS wants to make a more clear differenciation between volumes and the ephemeral storage of a virtual machine. For all SCS deployments we want to assume that volumes are always residing in a storage backend that is NOT on the same host as a hypervisor - in short terms: Volumes are network storage. Ephemeral storage on the other hand is the only storage residing on a compute host. It is created by creating a VM directly from an Image and is automatically los as soon as the VM cease to exist. Volumes on the other hand have to be created from Images and only after that can be used for VMs. They are persistant and will remain in the last state a VM has written on them before they cease to exit. Being persistant and not relying on the host where the VM resides, Volumes can easily be attached to another VM in case of a node outage and VMs be migrated way more easily, because only metadata and data in RAM has to be shifted to another host, accelerating any migration or evacuation of a VM.
```
fkr, line 10:
```suggestion
Volumes in OpenStack are virtual drives. They are managed by the storage service Cinder, which abstracts creation and usage of many different storage backends. While it is possible to use a backend like lvm which can reside on the same host as the hypervisor, the SCS wants to make a more clear differenciation between volumes and the ephemeral storage of a virtual machine. For all SCS deployments we want to assume that volumes are always residing in a storage backend that is NOT on the same host as a hypervisor - in short terms: Volumes are network storage. Ephemeral storage on the other hand is the only storage residing on a compute host. It is created by creating a VM directly from an Image and is automatically los as soon as the VM cease to exist. Volumes on the other hand have to be created from Images and only after that can be used for VMs. They are persistant and will remain in the last state a VM has written on them before they cease to exit. Being persistant and not relying on the host where the VM resides, Volumes can easily be attached to another VM in case of a node outage and VMs be migrated way more easily, because only metadata and data in RAM has to be shifted to another host, accelerating any migration or evacuation of a VM.
```

berendt:
My questions were all answered during an SCS call in which the topic was discussed.

maliblatt (approved):
Looks good to me from Operators point of view.

fkr (approved):

PR #329, opened 2023-08-10
Extend scs-0003: more context and process (resolves issues/#343)

mbuechse:
- added lots of context in the introduction
- introduced the distinction between "certificate" and "certificate type"
- improved misleading wording in the motivation
- rectified incorrect section levels
- added process description (governance)

### DIFF
--- a/Standards/scs-0003-v1-sovereign-cloud-standards-yaml.md
+++ b/Standards/scs-0003-v1-sovereign-cloud-standards-yaml.md
@@ -39,33 +39,33 @@ and so on (but usually, we aim to keep at most two versions in effect).
 
 This decision record describes two main points:
 
-1. How we denote our certification types by means of a YAML file.
-2. Our process for constructing and progressing the certification types.
+1. How we denote our certificate types by means of a YAML file.
+2. Our process for constructing and progressing the certificate types.
 
 ## Motivation
 
 This decision record establishes a mechanism (by means of the YAML file) with the following three main objectives:
 
-- to provide an overview of the mandatory standards for the different SCS certification levels
-- to make the lifecycle of certification levels traceable
+- to provide an overview of the mandatory standards for the different SCS certificate types
+- to make the lifecycle of certificate types traceable
 - to provide a machine-readable document for further processing (e.g. for a compliance tool suite or continuous integration).
 
 ### Overview of mandatory SCS standards
 
 Digging through a repository of draft, stable, replaced and rejected standards becomes increasingly challenging with a growing number
-documents and decision records. A central document that lists all mandatory standards to acquire a certain certification level can
+documents and decision records. A central document that lists all mandatory standards to acquire a certificate of a certain type can
 resolve this issue. It provides clarity for providers as well as users and helps to understand the value
 proposition of SCS.
 
-### Lifecycle of certification levels
+### Lifecycle of certificate types
 
-Standards and therefore certifications will evolve over time. To provide transparency and traceability for the lifecycle of SCS certification
-levels, the whole history of our certifications should be recorded. Pre-notification of changes to our certification levels allows
+Standards and therefore certifications will evolve over time. To provide transparency and traceability for the lifecycle of SCS certificate
+types, the whole history of our certifications should be recorded. Pre-notification of changes to our certificate types allows
 users to adapt their environments or deployment automation to the new standards in advance.
 
 ### Machine-readability for further processing
 
-By providing a machine-readable document, we can generate web-friendly overviews of our certification levels as well as create a tool suite
+By providing a machine-readable document, we can generate web-friendly overviews of our certificate types as well as create a tool suite
 that checks environments against all described standards.
 
 ## SCS Certification YAML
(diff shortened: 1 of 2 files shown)


Standards/scs-0003-v1-sovereign-cloud-standards-yaml.md
```diff
@@ -10,19 +10,47 @@ track: Global
...
-certification levels and the mandatory standards therein.
+SCS plans to offer six types of certificates, which are sorted into two dimensions:
+
+1. _certification level_, of which there are three:
+   - SCS-compatible
+   - SCS-open
+   - SCS-sovereign
+2. _cloud layer_, of which there are two:
+   - infastructure as a service (IaaS)
+   - Kubernetes as a service (KaaS)
```
maxwolfs, line 21: I think we need a decision on how to call this layer and implement it everywhere. The team is currently called `Team Container` and often CaaS is preferred over KaaS, as it doesn't only hold Kubernetes.

mbuechse: Yes, let's talk about that in the SIG meeting on 2023-08-24. As of right now, `kaas` is the term used in the yaml and this document, and I stuck to it.

garloff (approved):
/lgtm
Thanks @mbuechse

mbuechse:
@alexander-diab This PR is good to be merged; Kurt is satisfied, apparently highly so. Do you have any objections?

Standards/scs-0003-v1-sovereign-cloud-standards-yaml.md
```diff
@@ -10,34 +10,62 @@ track: Global
...
+
+This decision record describes two main points:
+
+1. How we denote our certificate types by means of a YAML file.
+2. Our process for constructing and progressing the certificate types.
 
 ## Motivation
 
-This decision record has three main objectives:
+This decision record establishes a mechanism (by means of the YAML file) with the following three main objectives:
```
anjastrunk, line 47: Isn't this of type `Procedural` and not of `Decision Record`?

```suggestion
This standard establishes a mechanism (by means of the YAML file) with the following three main objectives:
```

mbuechse: The verbiage in question didn't stem from me. Anyway, following your argumentation, we should change "decision record" into "procedural", right? However, I suppose "decision record" could be construed as a kind of general term (a superset) that includes Standards, Procedurals, and any other document. From this point of view, every standard is a decision record, but not every decision record is a standard. If we expressly designate a document as "decision record" in the header, then this mostly means that it doesn't fit into any of the subcategories. Just one way of looking at it.

anjastrunk:
Regarding yaml file:

Do we have **one** yaml file with **all** layers and levels at the end, or **one** yaml file per level and layer, such as SCS-compatible-IaaS.yaml for layer IaaS and level "SCS-compatible"

I prefer the second option, as one file may be grow over time and gets confusting. We may even think to have one file or each version, such as  SCS-compatible-IaaS-v1.yaml for version 1 of  layer "SCS-compatible" certificate for IaaS.

mbuechse:
> Regarding yaml file:
> 
> Do we have **one** yaml file with **all** layers and levels at the end, or **one** yaml file per level and layer, such as SCS-compatible-IaaS.yaml for layer IaaS and level "SCS-compatible"

Currently, it's one file per level, containing both layers. I would prefer one file per type of certificate, that is to say, one file per combination of level and layer. But this has been an open topic in the SIG for months now.

anjastrunk:
I like your improvements. It makes thing clearer.

One question, which came to may mind: Is there already a standard for receiving a certificate as CSP? I mean, how das a CSP apply for a certificate, how often will a re-certification take place and what are the consequences if certificate rules are violated?

mbuechse:
> One question, which came to may mind: Is there already a standard for receiving a certificate as CSP? I mean, how das a CSP apply for a certificate, how often will a re-certification take place and what are the consequences if certificate rules are violated?

@anjastrunk Something like that is being worked on in https://github.com/SovereignCloudStack/standards/issues/337

anjastrunk:
> > Regarding yaml file:
> > Do we have **one** yaml file with **all** layers and levels at the end, or **one** yaml file per level and layer, such as SCS-compatible-IaaS.yaml for layer IaaS and level "SCS-compatible"
> 
> Currently, it's one file per level, containing both layers. I would prefer one file per type of certificate, that is to say, one file per combination of level and layer. But this has been an open topic in the SIG for months now.
> 
> @anjastrunk see here: https://input.scs.community/2023-scs-sig-standardization#2023-06-29 I'm pretty sure that I had raised the matter even earlier, but it got postponed at least once, and even when it got discussed, it was postponed again.

We should but this on next SIG' agenda and make a decision!

anjastrunk (approved):

anjastrunk:
> > One question, which came to may mind: Is there already a standard for receiving a certificate as CSP? I mean, how das a CSP apply for a certificate, how often will a re-certification take place and what are the consequences if certificate rules are violated?
> 
> @anjastrunk Something like that is being worked on in #337

Yes. Thanks.

mbuechse:
> We should put this on next SIG' agenda and make a decision!

I prepared the agenda accordingly.

PR #330, opened 2023-08-14
Drop paragraph recommending digital image signature (resolves #328)

### LINKED ISSUE #328: No longer recommend image signatures
As shown by @josephineSei in https://input.scs.community/RID5cVHpTaG_AEDYsB9gVQ# the recommendation of image signatures in the image metadata standard is not feasible.

Definition of ready:
- [ ] It has been decided whether this change requires a new version of the standard (because the current version is stable).

Definition of done:
- [ ] A new version of the standard has been created with the paragraph in question dropped.
- [ ] The test script has been adjusted accordingly (use a command-line option for that).
- [ ] The adjusted script invocation has been incorporated into <https://github.com/SovereignCloudStack/standards/blob/issue/267/Tests/testing/scs-compatible-test.yaml>
- [ ] The new version of the standard has been added to https://github.com/SovereignCloudStack/standards/issues/317

mbuechse:
This change does not render any conformant implementation non-conformant.
Being mere recommendations, the properties in question were not even checked by the script.

### DIFF
--- a/Standards/scs-0102-v1-image-metadata.md
+++ b/Standards/scs-0102-v1-image-metadata.md
@@ -232,10 +232,6 @@ upstream images, they should point to the place where they document and offer th
   for the image file. (This references the image file in the format it is stored in, we
   recommend raw over qcow2 for systems that use ceph.) Note that these values are
   typically generated automatically upon image registration.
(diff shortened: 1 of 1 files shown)


josephineSei (approved):
Thank you for removing this

PR #334, opened 2023-08-21
Update standard K8s version policy

cah-hbaum:
The standard for the K8s version policy is updated in order to add the support period for Kubernetes versions into it. Since this is a bit contradictive with the K8s version recency, the whole standard needed to be updated.

### DIFF
--- a/Standards/scs-0210-v2-k8s-version-policy.md
+++ b/Standards/scs-0210-v2-k8s-version-policy.md
@@ -1,0 +1,81 @@
+---
+title: SCS K8S Version Policy
+type: Standard
+status: Draft
+track: KaaS
+---
+
+## Introduction
+
+The Kubernetes project maintains multiple release versions including their patched versions.
+In the project, the three most recent minor releases are actively maintained, with a fourth
+version being in development. As soon as a new minor version is officially released,
+the oldest version is dropped out of the support period.
+Kubernetes supports its releases for around 14 months. 12 of these are the standard
+support period. The remaining 2 months are the end-of-life support period for things like:
+
+- CVEs (under the advisement of the Security Response Committee)
+- dependency issues (including base image updates)
+- critical core component issues
+
+More information can be found under [Kubernetes Support Period].
+
+The Kubernetes release cycle is set around 4 months, which usually results in about
+**3 minor** releases per year [Kubernetes Release Cycle](https://kubernetes.io/releases/release/#the-release-cycle).
+
+Patches to these releases are provided monthly, with the exception of the first patch,
+which is usually provided 1-2 weeks after the initial release [Patch Release Cadence](https://kubernetes.io/releases/patch-releases/#cadence).
+
+## Motivation
+
+Kubernetes is a living, fast-paced project, which follows a pre-defined release cycle.
+This enables forward planning with regards to releases and patches, but also implies a
+necessity to upgrade to newer versions quickly, since these often include new features,
+important security updates or especially if a previous version falls out of the support
+period window.
+
+We want to achieve an up-to-date policy, meaning that providers should be mostly in
+sync with the upstream and don't fall behind the official Kubernetes releases.
+This is achievable, since new versions are released periodical on a well communicated
+schedule, enabling providers and users to set up processes around it.
+Being up to date ensures that security issues and bugs are addressed and new features
+are made available when using SCS compliant clusters.
+
+It is nevertheless important to at least support all Kubernetes versions that are still
+inside the support period, since users could depend on specific versions or may need
+longer to upgrade their workloads to a newer version.
+
+The standard therefore should provide a version recency policy as well as a support
+window period.
+
+## Decision
+
+In order to keep up-to date with the latest Kubernetes features, bug fixes and security improvements,
+the provided Kubernetes versions should be kept up to date with the upstream.
+
+- The latest minor version MUST be provided no later than 4 months after release.
+- The latest patch version MUST be provided no later than 1 week after release.
+- This time period MUST be even shorter for patches that target critical CVEs (CVSS >= 8).
+  It is RECOMMENDED to provide a new patch version in a 2 day time period after their release.
+- New versions MUST be tested before being rolled out on productive infrastructure;
+  at least the CNCF E2E tests should be passed beforehand.
+
+At the same time, providers must support Kubernetes versions at least as long as the
+official sources as mentioned in the [Kubernetes Support Period](https://kubernetes.io/releases/patch-releases/#support-period).
+
+- Kubernetes versions MUST be supported as long as the official sources support them.
+  The current support period can therefore be found in [Kubernetes Support Period].
+- It is RECOMMENDED to not support versions after this period in order to not encourage
+  usage of out-of-date versions.
+
+## Related Documents
+
+All documents regarding versioning, releases, etc. for the official Kubernetes projects can be found here:
+[Kubernetes Releases](https://kubernetes.io/releases/)
+[Kubernetes Support Period](https://kubernetes.io/releases/patch-releases/#support-period)
+[Kubernetes Release Cycle](https://kubernetes.io/releases/release/#the-release-cycle)
+[Patch Release Cadence](https://kubernetes.io/releases/patch-releases/#cadence)
+
+## Validation / Conformance
+
+*This section will be updated when the conformance tests are written.*


Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,75 @@
...
+which is usually provided 1-2 weeks after the initial release.
+[Patch Release Cadence](https://kubernetes.io/releases/patch-releases/#cadence)
+
+## Motivation
+
+Kubernetes is a living, fast-paced project, which follows a pre-defined release cycle.
+This enables forward planning with regards to releases and patches, but also implies a
+necessity to upgrade to newer versions quickly.
+
+We want to achieve an up to date policy, which means that providers should be mostly in sync with the upstream and don't
```
garloff, line 38: Language hint:
up-to-date (with dashes).

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,75 @@
...
+fall behind the official Kubernetes releases. In our opinion, this should be achievable, since new versions are released
```
garloff, line 39: "In our opinion" is not the language I would expect in a standard.
Suggestion:
This is achievable since new versions are released ...
(without comma)

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,75 @@
...
+periodical on a well communicated schedule. This should enable processes to be set up around this.
+Being up to date ensures, that security issues and bugs are addressed and new features are made available when using SCS compliant clusters.
```
garloff, line 41: no comma in front of that

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,75 @@
...
+
+On the other hand, providers should support minor Kubernetes versions at least as long as
+the official support period is, since users could depend on specific versions.
```
garloff, line 44: OTOH, since users may depend on specific versions, providers should support minor Kubernetes versions *for the duration of the official support period*.

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,75 @@
...
+This could be a bit contradicting in some cases, because the necessity to upgrade versions
+would theoretically drop the support period after that upgrade period.
```
garloff, line 46: I don't get it. What i contradicting?
Month 0: New minor k8s version N
Month ~2: Introduction of the version by service provider
Month 12: End of generic upstream maintenance.
Month 12: Recommendation by providers to users to upgrade (to version N+3 (if already provided), N+2 or N+1)
Month 14: End of upstream security fixes
Month 14: Final warning and end of support for minor by provider

Where's the contradiction?

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,75 @@
...
+
+It is therefore more reasonable, to recommend a version upgrade and make the version
+support period a necessity.
+
+## Decision
+
+In order to keep up-to date with the latest Kubernetes features, bug fixes and security improvements,
+the provided Kubernetes versions should be kept up to date with the upstream.
+
+- It is RECOMMENDED, that the latest minor version is provided no later than 4 months after release.
```
garloff, line 56: We wanted to have stronger language here in the past.
Providers MUST provide the latest minor version no later than ...

With that language, we can make this a hard SCS-compatible Container certification requirement.

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,75 @@
...
+- It is RECOMMENDED, that the latest patch version is provided no later than 1 week after release.
+- It is RECOMMENDED, that this time period should be even shorter for patches that target critical CVEs (CVSS >= 8). It is probably recommended to have a patch period around 1 to 3 days for these cases.
+- It is RECOMMENDED, that new versions SHOULD be tested before being rolled out on an infrastructure. How this testing will be done won't be defined in this document.
```
garloff, line 59: We could add that at least the CNCF conformance testsuite must be passed.

Again on RECOMMENDED vs MUST: The detail with faster patching for CVEs probably should stay in the RECOMMENDED state, but I see no reason not to mandate the rest.

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,75 @@
...
+support period a necessity.
+
+## Decision
+
+In order to keep up-to date with the latest Kubernetes features, bug fixes and security improvements,
+the provided Kubernetes versions should be kept up to date with the upstream.
+
+- It is RECOMMENDED, that the latest minor version is provided no later than 4 months after release.
+- It is RECOMMENDED, that the latest patch version is provided no later than 1 week after release.
+- It is RECOMMENDED, that this time period should be even shorter for patches that target critical CVEs (CVSS >= 8). It is probably recommended to have a patch period around 1 to 3 days for these cases.
```
garloff, line 58: You could leave out the last sentence. If you want to keep it, remove the "probably"

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,75 @@
...
+- It is RECOMMENDED, that new versions SHOULD be tested before being rolled out on an infrastructure. How this testing will be done won't be defined in this document.
+
+At the same time, provider should support Kubernetes versions at least as long as the official sources as mentioned
+in the Kubernetes Support Period document.
+
+- Kubernetes versions SHOULD be supported as long as the official sources support them. At the moment, this support period
```
garloff, line 64: Isn't this statement a duplication of the previous paragraph?

garloff (commented):
Two points on language:
* Standards should not make extensive use of "probably" or "we believe". We need to clarify these things until we are convinced. And then just make statements what is expected.
* relative clause "The reason that XYZ can be done is ..." do not have commas in English

Two on the content:
* This has been toned down into mere recommendations while the original intention was to have stronger language MUST. Did we agree as a team to really make it so soft? How much value remains then if customers can not rely on anything even if the container SCS-compatible cert has been granted?
* There seems to be confusion about support period for minor versions that puzzles me. Let's clarify this!

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: I agree, but that was also more of an example/proposal, because I don't know what would be the best timescale for such a critical patch.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: Depends on which paragraph you mean here. I thought the "decision" paragraph should contain the actual standard decisions, which IMO could "duplicate" something from a previous paragraph.

cah-hbaum:
Answering to https://github.com/SovereignCloudStack/standards/pull/334#pullrequestreview-1592868816 and https://github.com/SovereignCloudStack/standards/pull/334#discussion_r1303869016 since they ask the same things:

So I had a misunderstanding on how this K8s providing should work, since I thought that would be the provided version for things like managed K8s clusters.
If the K8s version means the version made available to the users of a cloud by the provider, this would imply two problems to me:
1) The previously made test for `k8s version recency` is wrong, since we just test a version of a cluster. This wouldn't matter in the case described above, since a cluster wouldn't need to be updated right away (only the version the provider needs to offer) and the failure on the test if a version is delayed (according to the standard) would be false in this case as well.
2) If a new version is published by Kubernetes, I see no reason to not offer it right away. I think the versions are sufficiently tested and having a delay between the official release date and a "provided" release date just feels not right.

I think what I would want to know here to clear that up:
What is the use-case?
How are the Kubernetes versions provided?
And can this even be tested (in the next step), since we can't check if a version is "provided", only the version of a cluster (?)

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: Added the MUST keyword on the critical patch sentence.

I would leave the last sentence in and the document on `Draft` until tests are written and we can write in what tests are used.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: Fixed.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: Fixed.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: See my comment https://github.com/SovereignCloudStack/standards/pull/334#issuecomment-1691253644

garloff:
> So I had a misunderstanding on how this K8s providing should work, since I thought that would be the provided version for things like managed K8s clusters.

I assume that even in managed k8s offerings, the k8s version updates are normally done only with the user's consent.
At least for minor updates (like going from v1.24.xx to v1.25.xx), the upgrade tends to require the user (who manages the workloads) to review the changelogs with the deprecations ... to ensure the workload is still working after such an upgrade.

So, looking at a cluster and determine it's level of outdatedness is indeed not very useful.

Instead, we need to ensure that the latest patchlevel (with a delay of less than a week) is *available*.
Same for latest minor versions (with a max delay of a generous 4 months).

I agree that testing this is difficult.
* In our reference implementation, users can select v1.27.x and get v1.27.<LATEST> in return.
  For this, we need to adjust https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/65af7217a4c2d849d1a408d36a3c9c6ea6c554aa/terraform/files/bin/openstack-kube-versions.inc#L6
  and the node image must be available. This one comes from OSISMs minio: 
  https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/65af7217a4c2d849d1a408d36a3c9c6ea6c554aa/terraform/files/bin/upload_capi_image.sh#L30
* If we fail to touch the `openstack-kube-versions.inc` file, the user could still successfully force a version. As long as the node image is available, this will work.
* Some providers might serve their customers well by registering those node images centrally as opposed to having each cluster user download them as needed. (The scripts will use those pre-registered images then.)

Now, this is all specific to our reference implementation, unfortunately.
Tests can be developed to check `openstack-kube-versions.inc` and the image availability.
In a generic approach, we have not defined a way to discover available k8s versions, so we can not yet cover other providers. Maybe we'll have a mechanism for all clusterstack users when we move to next-gen Ref.Impl.

cah-hbaum:
> Now, this is all specific to our reference implementation, unfortunately. Tests can be developed to check `openstack-kube-versions.inc` and the image availability. In a generic approach, we have not defined a way to discover available k8s versions, so we can not yet cover other providers. Maybe we'll have a mechanism for all clusterstack users when we move to next-gen Ref.Impl.

Ok I think I understand the intention behind all of this, but doesn't this still make the first test for k8s cluster recency *incorrect*, since the test is only applied to a cluster and we shouldn't look for a version recency in a cluster, since this is seemingly up to the user?

cah-hbaum:
Made an update to the text.
I still think it will be very hard to test this (if its even possible), but we will see. This is only the standard for now.

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,81 @@
...
+## Introduction
+
+The Kubernetes project maintains multiple release versions including their patched versions.
+In the project, the three most recent minor releases are actively maintained, with a fourth
+version being in development. As soon as a new minor version is officially released,
+the oldest version is dropped out of the support period.
+Kubernetes supports its releases for around 14 months. 12 of these are the standard
+support period. The remaining 2 months are end-of-life support for things like:
+
+- CVEs (under the advisement of the Security Response Committee)
```
anjastrunk, line 18: What does "CVE" stand for?

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,81 @@
...
+- dependency issues (including base image updates)
+- critical core component issues
+
+More information can be found under [Kubernetes Support Period].
```
anjastrunk, line 22: There is a hyperlink missing, isn't it?

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,81 @@
...
+
+The Kubernetes releases are planned for about 4 months, which usually results in about
+**3 minor** releases per year [Kubernetes Release Cycle].
```
anjastrunk, line 25: d
```suggestion
**3 minor** releases per year [Kubernetes Release Cycle](https://kubernetes.io/releases/release/).
```

anjastrunk: I just recognized, that links are listed at the end of the document. Is this a common practice in SCS standards?

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Decision
+
+In order to keep up-to date with the latest Kubernetes features, bug fixes and security improvements,
+the provided Kubernetes versions should be kept up to date with the upstream.
+
+- The latest minor version SHOULD be provided no later than 4 months after release.
+- The latest patch version SHOULD be provided no later than 1 week after release.
+- This time period MUST be even shorter for patches that target critical CVEs (CVSS >= 8).
+  It is RECOMMENDED to provide the version in a 2 day time period after release.
```
anjastrunk, line 60: A new "version" of what: patch or minor release?

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,81 @@
...
+- New versions SHOULD be tested before being rolled out on productive infrastructure.
```
anjastrunk, line 61: A new "version" of what: patch or minor release?

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,81 @@
...
+window period.
+
+## Decision
+
+In order to keep up-to date with the latest Kubernetes features, bug fixes and security improvements,
+the provided Kubernetes versions should be kept up to date with the upstream.
+
+- The latest minor version SHOULD be provided no later than 4 months after release.
+- The latest patch version SHOULD be provided no later than 1 week after release.
+- This time period MUST be even shorter for patches that target critical CVEs (CVSS >= 8).
```
anjastrunk, line 59: How short MUST support time for critical patches be?

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,81 @@
...
+  It is RECOMMENDED to provide the version in a 2 day time period after release.
+- New versions SHOULD be tested before being rolled out on productive infrastructure.
+
+At the same time, provider should support Kubernetes versions at least as long as the official sources as mentioned
+in the Kubernetes Support Period document.
+
+- Kubernetes versions SHOULD be supported as long as the official sources support them.
+  At the moment, this support period is 14 months.
+- It is RECOMMENDED to not support versions after this period in order to not encourage
+  usage of out-of-date versions.
```
anjastrunk, line 69: As line 57 allows to delay latest minor version for 4 month, I would take this into account with respect to shut of oldest minor version. I can image, that customers may not adapt their applications to latest minor version, but may skip ones. Hence, I would rephrase this sentence in order to not remove older verisons, before new versions will be supported.


```suggestion
- It is RECOMMENDED to not support versions after this period in order to not encourage
  usage of out-of-date versions, but not before the latest minor version is supported.
```

garloff:
> If the K8s version means the version made available to the users of a cloud by the provider, this would imply two problems to me:
> 
>     1. The previously made test for `k8s version recency` is wrong, since we just test a version of a cluster. This wouldn't matter in the case described above, since a cluster wouldn't need to be updated right away (only the version the provider needs to offer) and the failure on the test if a version is delayed (according to the standard) would be false in this case as well.

Right. The mere existence of a cluster with outdated k8s version in it does not indicate a failure to comply with the SCS-compatibility. (For a managed k8s scenario, I would expect a provider to push customers to agree to upgrading at some point though, but we have no recommended let alone mandatory policies for this at this point.)

A test would look like this:
* We try to create a cluster with a version that should be available according to the policy.
  - If it succeeds (i.e. a working cluster is created), we are good.
  - It it fails, we are likely not.

garloff:
> 2\. If a new version is published by Kubernetes, I see no reason to not offer it right away. I think the versions are sufficiently tested and having a delay between the official release date and a "provided" release date just feels not right.

Well ... we currently don't offer k8s-v1.28.x officially yet, as capo does not officially support k8s-v1.28 yet.
(Yes, we provide it, it's tested and works, but we have protected it with an `--allow-preview-versions` flag to makr it as TechPreview.)
The up to four months is a compromise that came out of the discussion with providers. While it seems long to me, this is what everyone was willing to commit to.
For patch levels, one week seems reasonable to me -- you need to build node images and run the E2E tests at least.

Standards/scs-0210-v2-k8s-version-policy.md
garloff: https://en.wikipedia.org/wiki/Common_Vulnerabilities_and_Exposures

Standards/scs-0210-v2-k8s-version-policy.md
garloff: https://kubernetes.io/releases/
https://endoflife.date/kubernetes

Standards/scs-0210-v2-k8s-version-policy.md
garloff: No. Linking docs that explain a statement or term is best be done in place. A link I would click if I need an explanation or definition or want to see the source of truth for a statement.

A references section at the end is more for further reading or background reading (in case I'm new to the topic).
Also link to related content, e.g. related standards.

Standards/scs-0210-v2-k8s-version-policy.md
garloff:
```suggestion
  It is RECOMMENDED to provide a new patch version in a 2 day time period after release in this case.
```

Standards/scs-0210-v2-k8s-version-policy.md
garloff: Both.
We could add here that they need to pass the CNCF E2E tests at least.

Standards/scs-0210-v2-k8s-version-policy.md
garloff: The statement here says MUST be shorter than one week (i.e. < 168 hours), and the next sentence goes on to recommend smaller than 2d (<48hrs).

Standards/scs-0210-v2-k8s-version-policy.md
garloff: If I do the math:
1.N is released and provided in month 0, extended support till month 14
1.N+1 is released in month 4 and provided in month 8, extended support till month 18
1.N+2 is released in month 8 and provided in month 12, extended support till month 22
So, skipping one level from 1.N to 1.N+2 is still possible with a 2months overlap window.
Sidenote: Technically, a rolling upgrade will need to do the intermediate step (but of course that could be very short-lived).
I would not complicate the policy, as skipping one version does not require it.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: The hyperlinks are at the end of the document in their own section.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: Updated the line, but changed the second suggestion a bit since it sounded not quite right.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: Added this line.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: As @garloff said, the next sentence provides the information about the support time for the critical patches.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: I would also not add this clause, since it complicates things as @garloff said, and the versions are provided in regular time intervals. Also we can't control what customers want to do with their applications; as long as the providers give them access to the latest versions (as the standard says), the can decide themselves what to do.

cah-hbaum:
@garloff thanks for your comments, they make this issue and standard more clear to me.
I will make a small update today in order to fix the issues coming from the comments and then make a slightly bigger update next week in order to make things more clear with regard to the overall story of this standard.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: Should we really put links in different locations depending on their relation to the standard? I would rather try to put everything in the final section and check if the link nevertheless works in the document itself.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: Check [here](https://github.com/SovereignCloudStack/standards/pull/334#discussion_r1341266362)

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: Thank you for providing the link, I will put in a glossary section next week.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: Changed this section, so closing the conversation.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: Changed this section, so closing the conversation.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: Changed this section, so closing the conversation.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: Changed this section, so closing the conversation.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: Agree. I wrote this into the document in the latest commit.

cah-hbaum:
Update the text a bit more and rebased the whole branch in order to include the latest changes from main.

anjastrunk (approved):
LGTM

Standards/scs-0210-v2-k8s-version-policy.md
anjastrunk: In favor of readability, I would add links in place. And put, as @garloff mentioned, additional documents, not related to a special term, at the bottom of the document.

cah-hbaum: Put the links in the text.

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,81 @@
...
+
+The standard therefore should provide a version recency policy as well as a support
+window period.
+
+## Decision
+
+In order to keep up-to date with the latest Kubernetes features, bug fixes and security improvements,
+the provided Kubernetes versions should be kept up to date with the upstream.
+
+- The latest minor version SHOULD be provided no later than 4 months after release.
```
garloff, line 56: Why is this not a MUST?
4 months is plenty of time to get the latest minor release built, validated, fixed (if needed, which is normally not the case) and published.

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,81 @@
...
+- The latest patch version SHOULD be provided no later than 1 week after release.
```
garloff, line 57: I believe we can make this a MUST as well.

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,81 @@
...
+- This time period MUST be even shorter for patches that target critical CVEs (CVSS >= 8).
+  It is RECOMMENDED to provide a new patch version in a 2 day time period after their release.
+- New versions SHOULD be tested before being rolled out on productive infrastructure;
```
garloff, line 60: MUST?

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,81 @@
...
+  at least the CNCF E2E tests should be passed beforehand.
```
garloff, line 61: could be a MUST as well (and if we want to be generous, we could allow failed e2e tests to be listed in a release note rather than requiring 100% pass rate).

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,81 @@
...
+
+At the same time, provider should support Kubernetes versions at least as long as the
```
garloff, line 63: Again, I would make this a MUST.

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -0,0 +1,81 @@
...
+official sources as mentioned in the [Kubernetes Support Period](https://kubernetes.io/releases/patch-releases/#support-period).
+
+- Kubernetes versions SHOULD be supported as long as the official sources support them.
```
garloff, line 66: Hmm, looks duplicate to the sentence before.
And should be MUST from my PoV.

garloff (commented):
Sorry if I have missed a prior answer or discussion.
But I believe that we've had discussed mandatory requirements for version recency before. These are useful, as it allows users to know which versions she can rely on as available on all KaaS platforms that are certified as SCS-compatible on the container layer. As user, it would make my live easier, as I can then decide for validating my workload on a k8s versions that I know must be available on all SCS-compatible KaaS solutions. A should does not provide this guarantee.
So were there reasons for watering this down where I missed a discussion? 
Was there a discussion that resulted in g=convincing reasons that we are really

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: To answer all of the MUST questions, I had a grave misunderstanding with the way how MUST and SHOULD are used. In my mind, they were equal, but after reading the RFC document regarding MUST / SHOULD / ... I can see my mistake.
And you're right, all of the ones you pointed out should be MUST.

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: See https://github.com/SovereignCloudStack/standards/pull/334#discussion_r1392522638

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: See https://github.com/SovereignCloudStack/standards/pull/334#discussion_r1392522638

Standards/scs-0210-v2-k8s-version-policy.md
cah-hbaum: See https://github.com/SovereignCloudStack/standards/pull/334#discussion_r1392522638

Also I think you're right about the generosity of passed tests, but I'm not really sure how to properly display this in Release notes.

cah-hbaum:
Since this issue is now open for 3 months and it got one Approval by @anjastrunk and I fixed all issues opened by @garloff, I'm gonna merge this now.

garloff:
Thanks, @cah-hbaum!

PR #336, opened 2023-08-25
Anti-affinity for k8s nodes (#226)

cah-hbaum:
Adds a standard decision for K8s node anti-affinity. This should lead to a standard, which ensures anti-affinity, and therefore high-availability and fault-tolerance on systems of SCS users.

### DIFF
--- a/Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
+++ b/Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
@@ -1,0 +1,123 @@
+---
+title: Kubernetes Nodes Anti Affinity
+type: Decision Record
+status: Draft
+track: KaaS
+---
+
+## Introduction
+
+A Kubernetes instance is provided as a cluster, which consists of a set of worker machines,
+so called nodes. A cluster is composed of a control plane and at least one worker node.
+The control plane manages the worker nodes and therefore the pods in the cluster by making
+decisions about scheduling, event detection and global decisions. Inside the control plane,
+multiple components exist, which can be duplicated and distributed over multiple machines
+inside the cluster. Typically, no user containers are run on these machines in order to
+separate the control plane from the live system.
+
+## Motivation
+
+In a productive environment, the control plane usually runs across multiple machines and
+a cluster usually contains multiple worker nodes in order to provide fault-tolerance and
+high availability.
+
+In order to ensure availability and scaling of workloads, even if some machines in the cluster
+could fail, they should be distributed over multiple nodes on different machines.
+This can be steered with the Affinity or Anti Affinity features, which are separated by
+Kubernetes into two features:
+
+Node Affinity
+The Node Affinity feature allows to match pods according to logical matches of
+key-value-pairs referring to labels of nodes.
+These can be defined with different weights or preferences in order to allow fine-grained
+selection of nodes. The feature works similar to the Kubernetes nodeSelector.
+It is defined in the PodSpec using the nodeAffinity field in the affinity section.
+
+Pod Affinity
+Pod Affinity or Pod Anti Affinity allows the constraint of pod scheduling based on the
+labels of pods already running on a node.
+This means the constraint will match other pods on a node according to their labels key-value-pairs
+and then either schedule the pod to the same (Affinity) or another (Anti Affinity) node.
+This feature is also defined in the PodSpec using the podAffinity field in the affinity section.
+
+Both features allow the usage of "required" or "preferred" keywords, which create
+"hard" or "soft" affinities. By using a hard affinity, a pod would need to be scheduled
+according to the rules set. If this possibility is not given, the pod can't be scheduled.
+A soft affinity would allow scheduling even if the requirements are not fulfilled, but
+they would be preferred if possible.
+
+These features allow efficient resource usage (e.g. by scheduling workloads to evenly
+distribute across nodes) and provide fault-tolerance and therefore high availability.
+But they also require more work during the setup of a service architecture, since nodes
+and pods need to be labelled and described consistently.
+
+In the case of SCS, affinity of the workloads themselves is not relevant, since this
+feature is mostly used by the customers of the providers.
+Instead, the expected standard should enable the Kubernetes cluster to handle Anti Affinity
+rules with a real physical separation as well as distributing the control plane over
+multiple machines in order to provide fault-tolerance during system outages.
+If the control plane survives an outage, a Kubernetes cluster can recover later on.
+
+## Design considerations
+
+SCS plans to require a Hard Anti Affinity and/or Redundancy for the control plane and
+a Soft Anti Affinity for workers. This means, that Anti Affinity would be required for
+the control planes and their pods and only optional (but encouraged) for workers.
+
+In order to achieve the goals for these components, meaning availability and fault tolerance
+for the control plane, in order to achieve an outage resistant cluster, and the availability
+promise given with Anti Affinity for pods on the worker nodes, a separation of nodes
+on the hardware or hypervisor level would need to be achieved. Since running multiple
+hypervisors on the same machine is in most cases not feasible and this approach
+wouldn't change availability for outages of individual machines, this should not be recommended.
+
+For the control plane, a reference to the "Kubernetes High Availability" [1]
+can be useful, since it provides two ways to set up a highly available cluster.
+Both approaches are very similar. The "Stacked Control Plane" requires less infrastructure,
+but also runs the risk of failed coupling, where if one node fails, the redundancy could be
+compromised due to the loss of a complete control plane instance.
+The "External ETCD" solves this problem, but also requires double the infrastructure, due
+to the externally incorporated etcd clusters.
+
+This also shows, that the wording "anti affinity" as used with Kubernetes pods is probably
+slightly misleading in the context of a Kubernetes control plane. It may consist of multiple
+pods with individual tasks, but distributing them over different nodes through Anti Affinity
+would probably still cascade the whole control plane into a complete failure, if one of
+the used nodes goes down. It could be possible to replicate specific important pods and
+assign them to different nodes, but at this point, a redundant setup like presented in [1] could be used.
+So Anti Affinity in this context probably means more like distribution over multiple
+physical machines, which needs to be planned beforehand on the machine/server level.
+
+Therefore would it be preferred for the control plane to use a redundant setup, which
+is separated over different physical machines, meaning at least half of the control plane
+nodes runs on a different physical machine as the rest.
+
+For worker nodes, the whole idea of Anti Affinity is more of a preferable situation.
+The nodes themselves should at best be distributed over different machines, but this
+is not a requirement, especially since smaller providers wouldn't have the capacity to
+provide enough machines for these distributed Kubernetes clusters. Since customers that
+use the Affinity feature would be especially interested in the worker nodes that host
+their workloads, it should be achieved that a good labeling system is provided for the
+nodes in order to see if two nodes are hosted on the same machine.
+
+## Decision
+
+The future standard should define a few important things in order to provide a solid base
+for the usage and advantages of workloads with Anti Affinity rules.
+
+Control planes SHOULD be made redundant in order to provide fault-tolerance and security
+against fatal errors on this layer, in the case of node failures. How this redundancy
+is achieved SHOULD be left to the providers, but since failure must be avoided, it is
+REQUIRED to at least duplicate control plane components. Half of every component SHOULD
+also be located on a different node on a different physical machine than the other half
+of them. This should provide at least the minimum requirements for a fault-tolerant control plane.
+
+The worker nodes are RECOMMENDED to be distributed over different machines. In order to
+provide clear information to the users, the nodes should be labeled to reflect the
+mapping to the underlying clusters. It should be noted, that it is NOT REQUIRED to have
+this anti affinity for the worker nodes due to the requirements of infrastructure and
+complexity associated with this.
+
+## Documents
+
+Kubernetes High Availability Documentation [1](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/high-availability/)
--- a/Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
+++ b/Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
@@ -15,13 +15,23 @@ multiple components exist, which can be duplicated and distributed over multiple
 inside the cluster. Typically, no user containers are run on these machines in order to
 separate the control plane from the live system.
 
+### Glossary
+
+The following special terms are used throughout this decision record document:
+
+| Term | Meaning |
+|---|---|
+| Worker | Virtual or bare-metal machine, which hosts workloads of customers |
+| Control Plane | Virtual or bare-metal machine, which hosts the container orchestration layer that exposes the API and interfaces to define, deploy, and manage the lifecycle of containers. |
+| Machine | Virtual or bare-metal entity with computational capabilities |
+
 ## Motivation
 
 In a productive environment, the control plane usually runs across multiple machines and
 a cluster usually contains multiple worker nodes in order to provide fault-tolerance and
 high availability.
 
-In order to ensure availability and scaling of workloads, even if some machines in the cluster
+In order to ensure availability and scaling of workloads, even if some nodes in the cluster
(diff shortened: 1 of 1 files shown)


Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
```diff
@@ -0,0 +1,123 @@
...
+to the externally incorporated etcd clusters.
+
+This also shows, that the wording "anti affinity" as used with Kubernetes pods is probably
+slightly misleading in the context of a Kubernetes control plane. It may consist of multiple
+pods with individual tasks, but distributing them over different nodes through Anti Affinity
+would probably still cascade the whole control plane into a complete failure, if one of
+the used nodes goes down. It could be possible to replicate specific important pods and
+assign them to different nodes, but at this point, a redundant setup like presented in [1] could be used.
+So Anti Affinity in this context probably means more like distribution over multiple
+physical machines, which needs to be planned beforehand on the machine/server level.
```
matofeder, line 89: Agree, IMO, the k8s control plane (CP) components, and the k8s "anti-affinity" feature should not be mentioned together. 

AFAIK CP components are distributed over the k8s nodes via a mechanism called  `static pod manifests` (in the case when `kubeadm` is used for k8s cluster bootstraping).   `kubeadm` generates static pod manifests for CP components to the `/etc/kubernetes/manifests` directory on each node. Then the `kubelet` watches this directory for Pods to create on startup, see [docs](https://kubernetes.io/docs/reference/setup-tools/kubeadm/implementation-details/#generate-static-pod-manifests-for-control-plane-components).

Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
```diff
@@ -0,0 +1,123 @@
...
+
+Therefore would it be preferred for the control plane to use a redundant setup, which
+is separated over different physical machines, meaning at least half of the control plane
+nodes runs on a different physical machine as the rest.
```
matofeder, line 93: KaaS reference implementation uses the ClusterAPI approach. ClusterAPI approach knows the concept of “failure domains”.  “failure domain” is used to control where to place the k8s nodes and physically distribute them. For a HA k8s cluster with three control plane nodes, we can distribute the nodes into three different availability zones, to survive one availability zone failure. 

Currently, Multi-AZ deployment is not SCS KaaS default but could be implemented see [docs](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/doc/usage/multi-az-and-multi-cloud-environments.md). 

Consider to mention the above, as IMO it is related to the sentence above.

Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
```diff
@@ -0,0 +1,123 @@
...
+
+For worker nodes, the whole idea of Anti Affinity is more of a preferable situation.
+The nodes themselves should at best be distributed over different machines, but this
+is not a requirement, especially since smaller providers wouldn't have the capacity to
+provide enough machines for these distributed Kubernetes clusters. Since customers that
+use the Affinity feature would be especially interested in the worker nodes that host
+their workloads, it should be achieved that a good labeling system is provided for the
+nodes in order to see if two nodes are hosted on the same machine.
```
matofeder, line 101: IMO, in the case of workload pods it is worth also mentioning the k8s concept called [Pod Topology Spread Constraints](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/).

This concept is very close to the anti-affinity feature, but there are some differences, nicely explained e.g. [here](https://stackoverflow.com/questions/73157345/kubernetes-spread-pods-across-nodes-using-podantiaffinity-vs-topologyspreadconst)

Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
```diff
@@ -0,0 +1,123 @@
...
+These can be defined with different weights or preferences in order to allow fine-grained
+selection of nodes. The feature works similar to the Kubernetes nodeSelector.
+It is defined in the PodSpec using the nodeAffinity field in the affinity section.
+
+Pod Affinity
+Pod Affinity or Pod Anti Affinity allows the constraint of pod scheduling based on the
+labels of pods already running on a node.
+This means the constraint will match other pods on a node according to their labels key-value-pairs
+and then either schedule the pod to the same (Affinity) or another (Anti Affinity) node.
+This feature is also defined in the PodSpec using the podAffinity field in the affinity section.
```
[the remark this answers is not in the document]
cah-hbaum: Good idea, I will include that in the next iteration.

Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
```diff
@@ -0,0 +1,123 @@
...
+Both features allow the usage of "required" or "preferred" keywords, which create
+"hard" or "soft" affinities. By using a hard affinity, a pod would need to be scheduled
+according to the rules set. If this possibility is not given, the pod can't be scheduled.
+A soft affinity would allow scheduling even if the requirements are not fulfilled, but
+they would be preferred if possible.
+
+These features allow efficient resource usage (e.g. by scheduling workloads to evenly
+distribute across nodes) and provide fault-tolerance and therefore high availability.
+But they also require more work during the setup of a service architecture, since nodes
+and pods need to be labelled and described consistently.
```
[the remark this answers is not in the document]
cah-hbaum: Will do, thanks for the proposal.

Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
```diff
@@ -0,0 +1,123 @@
...
+In order to achieve the goals for these components, meaning availability and fault tolerance
+for the control plane, in order to achieve an outage resistant cluster, and the availability
+promise given with Anti Affinity for pods on the worker nodes, a separation of nodes
+on the hardware or hypervisor level would need to be achieved. Since running multiple
+hypervisors on the same machine is in most cases not feasible and this approach
+wouldn't change availability for outages of individual machines, this should not be recommended.
+
+For the control plane, a reference to the "Kubernetes High Availability" [1]
+can be useful, since it provides two ways to set up a highly available cluster.
+Both approaches are very similar. The "Stacked Control Plane" requires less infrastructure,
```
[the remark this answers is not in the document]
cah-hbaum: Will do, thanks for the proposal.

Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
```diff
@@ -0,0 +1,123 @@
...
+but also runs the risk of failed coupling, where if one node fails, the redundancy could be
+compromised due to the loss of a complete control plane instance.
+The "External ETCD" solves this problem, but also requires double the infrastructure, due
```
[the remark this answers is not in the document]
cah-hbaum: Will do, thanks for the proposal.

Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
```diff
@@ -0,0 +1,123 @@
...
+## Introduction
+
+A Kubernetes instance is provided as a cluster, which consists of a set of worker machines,
+so called nodes. A cluster is composed of a control plane and at least one worker node.
+The control plane manages the worker nodes and therefore the pods in the cluster by making
+decisions about scheduling, event detection and global decisions. Inside the control plane,
+multiple components exist, which can be duplicated and distributed over multiple machines
+inside the cluster. Typically, no user containers are run on these machines in order to
+separate the control plane from the live system.
+
```
anjastrunk, line 17: I suggest to add a dedicated glossary section for important terms (e.g. control plane, worker, pods) as in https://github.com/SovereignCloudStack/standards/pull/343.

Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
cah-hbaum: Since this document is only a Decision Record, I'm fine with mentioning them here together (just to address the Issue description), but in the final standard, I would choose another description/name for everything, like "Distribution" or just "High Availability", since IMO this is what this whole idea comes from.

Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
cah-hbaum: Updated the section with your suggested text (but I changed it a little bit).

Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
cah-hbaum: Thats a nice concept and good to know it exists. I don't think its that relevant in the end for this DR/Standard, but I mentioned it nonetheless.

Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
```diff
@@ -0,0 +1,123 @@
...
+SCS plans to require a Hard Anti Affinity and/or Redundancy for the control plane and
+a Soft Anti Affinity for workers. This means, that Anti Affinity would be required for
+the control planes and their pods and only optional (but encouraged) for workers.
+
+In order to achieve the goals for these components, meaning availability and fault tolerance
+for the control plane, in order to achieve an outage resistant cluster, and the availability
+promise given with Anti Affinity for pods on the worker nodes, a separation of nodes
+on the hardware or hypervisor level would need to be achieved. Since running multiple
+hypervisors on the same machine is in most cases not feasible and this approach
```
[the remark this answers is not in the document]
cah-hbaum: Yeah you're right I was a bit confused there too. Fixed that section up.

Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
```diff
@@ -0,0 +1,123 @@
...
+In the case of SCS, affinity of the workloads themselves is not relevant, since this
+feature is mostly used by the customers of the providers.
+Instead, the expected standard should enable the Kubernetes cluster to handle Anti Affinity
+rules with a real physical separation as well as distributing the control plane over
+multiple machines in order to provide fault-tolerance during system outages.
+If the control plane survives an outage, a Kubernetes cluster can recover later on.
+
+## Design considerations
+
+SCS plans to require a Hard Anti Affinity and/or Redundancy for the control plane and
```
anjastrunk, line 63: In line 56 you exclude affinity for worker nodes and here you include it. Isn't this contradicting?

cah-hbaum: Maybe its not perfectly formulated. What I meant is that (Anti) Affinity isn't relevant for the SCS internally, since SCS in the end only standardizes the infrastructure, not the workload. But it is nonetheless important to standardize the possibility for (Anti) Affinity usage in the clusters.

Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
matofeder: It makes sense, thanks for clarifying

matofeder (approved):
/lgtm

Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
cah-hbaum: Updated the "Control Plane" part to reflect the K8s docs a bit more.

garloff:
Hmmm, reading through the document, it looks more complicated than it needs to be.

Maybe my view of the world is too simple and I'm happy to learn. But here's my understanding:
* Users of k8s clusters can schedule containers/pods such that they run on multiple worker nodes (e.g. by using a daemonset). For properly designed workloads, the loss of a worker node can then be survived without problems.
* The control plane -- when deployed on multiple nodes -- runs a clustered etcd (and multiple stateless components such as the kube-apiserver), so it survives the failure of a control plane node and can schedule a new one to run and join the cluster to achieve full redundancy again.

Both mechanisms are not very useful if several worker nodes or several control planes nodes are VMs that are running on the same host (hypervisor) system. Because then, one event (a host failure) would affect several k8s nodes of one type at once.

We have thus determined that
* Control plane node VMs of one cluster must never be scheduled to the same host. (This applies to solutions that have dedicated control plane nodes -- for a Gardener style solution with a shared control plane, we would require it to also run on several hosts, though the scheduling happens beforehand not on cluster creation.)
For our cluster-API based implementation on OpenStack, we can simply set hard anti-affinity for the VMs that host the control plane nodes. This ensures survival of the etcd data store and thus the ability for users to make API calls to do CRUD operations in the cluster.
(You are touching on an advanced topic here: If there are several availabiity zones / failure domains, going a step further and not just requiring different hosts but different AZs would be even better.)
* Worker node VMs of one cluster should not be scheduled on the same host. This is a soft requirement -- there are valid cases, where one woud want to have more worker nodes than there are hosts of the required type. In the reference implementation this is achieved by using soft anti-affiinity for the VMs that host the worker nodes.

I seem to be missing something here, as the proposed standards document looks a lot more complicated than my explanation here. You seem to imply that the control plane still does not survive a host failure?

Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
```diff
@@ -0,0 +1,123 @@
...
+These features allow efficient resource usage (e.g. by scheduling workloads to evenly
+distribute across nodes) and provide fault-tolerance and therefore high availability.
+But they also require more work during the setup of a service architecture, since nodes
+and pods need to be labelled and described consistently.
+
+In the case of SCS, affinity of the workloads themselves is not relevant, since this
+feature is mostly used by the customers of the providers.
+Instead, the expected standard should enable the Kubernetes cluster to handle Anti Affinity
+rules with a real physical separation as well as distributing the control plane over
+multiple machines in order to provide fault-tolerance during system outages.
```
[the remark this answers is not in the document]
garloff: On OpenStack, `openstack server  show` shows you the `hostId`. While it is mangled (and thus does not expose the real hypervisor Id to the user), the mangling is consistent. So if two VMs within the same project are on the same host/hypervisor, they will have the same `hostId`. If two VMs within the same project have different `hostId`s, they are not on the same host/hypervisor.

cah-hbaum:
> I seem to be missing something here, as the proposed standards document looks a lot more complicated than my explanation here. You seem to imply that the control plane still does not survive a host failure?

The main goal I had with this Decision Record (that is also something we should keep in mind, this is only a DR) was to set clearer goals and guidelines for the following standard. You're completely right with your observation about the essence of this document and that this could probably be cut a bit shorter. But I also wanted to make clear, that this document is more about "availability" with/trough "cluster distribution" and not really about "Anti Affinity", since this is (in my opinion) more widely used in the context of workloads, at least in the K8s space. So this document was more about clearing that up and finding the right concepts to apply to the standard following it.

About the the depth of commitments to get an (high)-available, fault-tolerant cluster: this is something I would put and discuss in the standard. I think there's always a scenario, where a cluster could go down due to some unexpected problems. Wider distribution of the underlying infrastructure (different physical machines, buildings, AZs, regions) most likely help, but this obviously increases the costs and isn't feasible for every business. I think the final decision how to implement this aspect should be left to the providers, but we can probably give out a recommendation.

Another aspect of the length of this document can be found here:
> Worker node VMs of one cluster should not be scheduled on the same host. This is a soft requirement -- there are valid cases, where one woud want to have more worker nodes than there are hosts of the required type. In the reference implementation this is achieved by using soft anti-affiinity for the VMs that host the worker nodes.

In the document, I wanted to make clear that the distribution of the worker nodes over several physical nodes isn't very useful, if the customers/users can't use them properly. They need clear information about the worker nodes and their distribution, otherwise all workloads could be scheduled (maybe unintentionally) on the same node, in which case a single worker node failure would bring down the whole environment of a single customer. Not good! Thats why I mentioned the labeling and communication in the DR.

Also, just to ask the question about this:
> Both mechanisms are not very useful if several worker nodes or several control planes nodes are VMs that are running on the same host (hypervisor) system. Because then, one event (a host failure) would affect several k8s nodes of one type at once.

I thought (at least from my experience and experiments), if all control plane nodes except one go down, the cluster should still be working. It's maybe not very responsable due to a high request load, but at least the data would still be secure. Rebuilding the cluster would take some time of course. 


I hope that made it more clear.

cah-hbaum:
I now created the follow-up PR for the real standard [here](https://github.com/SovereignCloudStack/standards/pull/354)

PR #341, opened 2023-09-06
Update SCS-Spec.MandatoryFlavors.verbose.yaml according to scs-0103-v1

### LINKED ISSUE #340: flavor-manager-input.py uses strings instead of real types for default values
Currently, the ``flavor-manager-input.py`` produces yaml files which contain strings instead of booleans for certain default values. Namely ``public`` and ``disabled``:
https://github.com/SovereignCloudStack/standards/blob/92b3e8cc85cfff4c424bee722961609103bf4ece/Tests/iaas/flavor-naming/flavor-manager-input.py#L41-L49

We are currently working on unifying the type parsing for the osism-flavor-manager and have encountered the problem, that the yaml file in this repo (``SCS-Spec.MandatoryFlavors.verbose.yaml``), which will be automatically downloaded by the osism-flavor-manager at the users discretion, contains strings instead of booleans because of the python script mentioned above:
https://github.com/SovereignCloudStack/standards/blob/92b3e8cc85cfff4c424bee722961609103bf4ece/Tests/iaas/SCS-Spec.MandatoryFlavors.verbose.yaml#L1-L13

We propose changing the ``flavor-manager-input.py`` in this repo and regenerating the yaml file with correctly typed values.

For reference, this is how the osism flavor.yaml file looks like: https://raw.githubusercontent.com/osism/openstack-flavor-manager/main/flavors.yaml

mbuechse:
Resolves #340

### DIFF
--- a/Tests/iaas/SCS-Spec.MandatoryFlavors.verbose.yaml
+++ b/Tests/iaas/SCS-Spec.MandatoryFlavors.verbose.yaml
@@ -88,14 +88,6 @@ mandatory:
   scs:cpu-type: crowded-core
   scs:name-v1: SCS-1L:1
   scs:name-v2: SCS-1L-1
-- name: SCS-4V-16-100s
-  cpus: 4
-  ram: 16384
-  disk: 100
-  scs:cpu-type: shared-core
-  scs:disk0-type: ssd
-  scs:name-v1: SCS-4V:16:100s
-  scs:name-v2: SCS-4V-16-100s
 - name: SCS-2V-4-20s
   cpus: 2
   ram: 4096
@@ -104,6 +96,14 @@ mandatory:
   scs:disk0-type: ssd
   scs:name-v1: SCS-2V:4:20s
   scs:name-v2: SCS-2V-4-20s
+- name: SCS-4V-16-100s
+  cpus: 4
+  ram: 16384
+  disk: 100
+  scs:cpu-type: shared-core
+  scs:disk0-type: ssd
+  scs:name-v1: SCS-4V:16:100s
+  scs:name-v2: SCS-4V-16-100s
 recommended:
 - name: SCS-1V-4-10
   cpus: 1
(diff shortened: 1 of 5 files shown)


mbuechse:
@gndrmnn Can you please check the new yaml file from this branch?
Are the types correct now?
Are the additional fields (`scs:name-v1` etc.) problematic?
I will also have to adjust the GitHub action to use the correct script.

mbuechse:
@gndrmnn I used the other script, because that will be the state of the art with the standard scs-0103-v1-standard-flavors in place. However, I shall correct the former script as well. As noted, the Github action should also be changed to use the new one.

mbuechse:
@gndrmnn @garloff I think this is ready for review now. In this case, it's advised to look at the total change rather than the individual commits, because the total change is in fact smaller than the intermediate versions (the order of some fields was shuffled, which is only a cosmetic issue, and I corrected that).

Tests/iaas/SCS-Spec.MandatoryFlavors.verbose.yaml
```diff
@@ -1,155 +1,199 @@
...
   cpus: 4
   ram: 16384
   disk: 100
-  description: alias=SCS-4V:16:100s
+  scs:cpu-type: shared-core
+  scs:disk0-type: ssd
+  scs:name-v1: SCS-4V:16:100s
+  scs:name-v2: SCS-4V-16-100s
 recommended:
 - name: SCS-1V-4-10
```
gndrmnn, line 108: Wouldn't ``optional_prefix: SCSX-`` imply that we have to use ``SCSX-[...]`` from this line on in the ``recommended`` section of the file? Or does ``optional_prefix: SCSX-`` mean something else?

mbuechse: I have no idea what it means, to be quite frank. I only copied it from a file that I saw in the wild. Does it mean anything at all? I mean, it says "optional". I don't seem to find any mention of it in the standard. I think we should just drop that line for good. Give me a minute.

mbuechse: I think I copied it from here https://github.com/osism/openstack-flavor-manager/blob/main/flavors.yaml
It's gone now :)

gndrmnn (approved):
LGTM, but someone with better knowledge of the actual standard in question should have a look on this, too

mbuechse:
@garloff @anjastrunk I think this is good to be merged. Please have a look and feel free to merge yourself (after rebasing onto main).

anjastrunk (approved):
LGTM!

PR #342, opened 2023-09-11
Amend scs-0003 according to 2023-09-07 SIG meeting

mbuechse:
- have one yaml file per type of certificate
- mention that votes have to be announced in advance

Related to https://github.com/SovereignCloudStack/issues/issues/343 .


mbuechse:
The markdown link check is failing, but that is not a problem in this case. The links used are only meant as examples, and there is no need for them to exist (yet). In one of the two instances, it will exist once this branch is merged into main.

anjastrunk (approved):
Looks good for me.

PR #354, opened 2023-10-05
K8s node distribution standard (#434)

cah-hbaum:
This MR adds a standard for distribution of k8s nodes over multiple machines, which should help alleviate failures of whole clusters and the ensuing loss of services and data, therefore ensuring availability of the resources.

Closes #434

### DIFF
--- a/Standards/scs-0214-v1-k8s-node-distribution.md
+++ b/Standards/scs-0214-v1-k8s-node-distribution.md
@@ -1,0 +1,92 @@
+---
+title: Kubernetes Node Distribution and Availability
+type: Standard
+status: Draft
+track: KaaS
+---
+
+## Introduction
+
+A Kubernetes instance is provided as a cluster, which consists of a set of machines,
+so called nodes. A cluster is composed of a control plane and at least one worker node.
+The control plane manages the worker nodes and therefore the pods in the cluster by making
+decisions about scheduling, event detection and rights management. Inside the control plane,
+multiple components exist, which can be duplicated and distributed over multiple nodes
+inside the cluster. Typically, no user workloads are run on these nodes in order to
+separate the controller component from user workloads, which could pose a security risk.
+
+### Glossary
+
+The following terms are used throughout this document:
+
+| Term | Meaning |
+|---|---|
+| Worker | Virtual or bare-metal machine, which hosts workloads of customers |
+| Control Plane | Virtual or bare-metal machine, which hosts the container orchestration layer that exposes the API and interfaces to define, deploy, and manage the lifecycle of containers. |
+| Machine | Virtual or bare-metal entity with computational capabilities |
+
+## Motivation
+
+In normal day-to-day operation, it is not unusual for some operational failures, either
+due to wear and tear of hardware, software misconfigurations, external problems or
+user errors. Whichever was the source of such an outage, it always means down-time for
+operations and users and possible even data loss.
+Therefore, a Kubernetes cluster in a productive environment should be distributed over
+multiple "failure zones" in order to provide fault-tolerance and high availability.
+This is especially important for the control plane of the cluster, since it contains the
+state of the whole cluster. A failure of this component could mean an unrecoverable failure
+of the whole cluster.
+
+## Design Considerations
+
+Most design considerations of this standard follow the previously written Decision Record
+[Kubernetes Nodes Anti Affinity] as well as the Kubernetes documents about [High Availability]
+and [Best practices for large clusters].
+
+SCS wishes to prefer distributed, highly-available systems due to their obvious advantages
+like fault-tolerance and data redundancy. But it also understands the costs and overhead
+for the providers associated with this effort, since the infrastructure needs to have
+hardware which will just be used to provide fail-over safety or duplication.
+
+The document [Best practices for large clusters] describes the concept of a failure zone.
+This term isn't defined any further, but can in this context be described as a number of
+physical (computing) machines in such a vicinity to each other (either through physical
+or logical interconnection in some way), that specific problems inside this zone would put
+all these machines at risk of failure/shutdown. It is therefore necessary for important
+data or services to not be present just on one failure zone.
+How such a failure zone should be defined is dependent on the risk model of the service/data
+and its owner as well as the capabilities of the provider. Zones could be set from things
+like single machines or racks up to whole datacenters or even regions, which could be
+coupled by things like electrical grids. They're therefore purely logical entities, which
+shouldn't be defined further in this document.
+
+## Decision
+
+This standard formulates the requirement for the distribution of Kubernetes nodes in order
+to provide a fault-tolerant and available Kubernetes cluster infrastructure.
+Since some providers only have small environments to work with and therefore couldn't
+comply with this standard, it will be treated as a RECOMMENDED standard, where providers
+can OPT-OUT.
+
+If the standard is used by a provider, the following decisions are binding and valid:
+
+- The control plane nodes MUST be distributed over multiple physical machines. Kubernetes
+  provides best-practices on this topic, which are also RECOMMENDED by SCS.
+- At least one control plane instance MUST be run in each "failure zone", more are
+  RECOMMENDED in each "failure zone" to provide fault-tolerance for each zone.
+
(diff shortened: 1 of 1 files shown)


mxmxchere:
I agree on the distribution of multiple control-plane nodes for large and production clusters. 

Maybe we should enlarge our view for other use-cases of k8s-clusters. There can also be very small dev-clusters which may be even disposed on a quick-basis where other controlplane scenarios (single-node or even shared-nodes) may be more suitable (cheaper).

We could reflect those attributes (shared control-plane, dedicated controlplane, dedicated and guaranteed to be split over various failure zones) in different cluster-flavors just as we did with the machine-flavors for openstack.

anjastrunk (approved):
LGTM

Standards/scs-0214-v1-k8s-node-distribution.md
```diff
@@ -0,0 +1,91 @@
...
+track: KaaS
+---
+
+## Introduction
+
+A Kubernetes instance is provided as a cluster, which consists of a set of machines,
+so called nodes. A cluster is composed of a control plane and at least one worker node.
+The control plane manages the worker nodes and therefore the pods in the cluster by making
+decisions about scheduling, event detection and rights management. Inside the control plane,
+multiple components exist, which can be duplicated and distributed over multiple machines
```
NotTheEvilOne, line 14: IMHO we should speak of nodes most of the times until the reader reaches glossary when describing components of an Kubernetes cluster. It may be provided by physical machines or machine flavors on hyperscalers but may also be VM instances also named "instance" instead of a "machine" provided by a classical IaaS. The glossary follows below, so at this time "machine" is fine but still I personally like the k8s name "node" more.

Standards/scs-0214-v1-k8s-node-distribution.md
```diff
@@ -0,0 +1,91 @@
...
+
+This standard formulates the requirement for the distribution of Kubernetes nodes in order
+to provide a fault-tolerant and available Kubernetes cluster infrastructure.
+Since some providers only have small environments to work with and therefore couldn't
+comply with this standard, it will be treated as a RECOMMENDED standard, where providers
+can OPT-OUT.
+
+If the standard is used by a provider, the following decisions are binding and valid:
+
+- The control plane nodes SHOULD be distributed over multiple physical machines. Kubernetes
```
NotTheEvilOne, line 73: In the section above the term "failure zone" is used multiple times. I personally have the feeling that "multiple physical machines" does not fit the intention perfectly. We would like that machines (as defined in the glossary, so we mean nodes here) should be placed in different failure zones to not being affected by issues on routers, burning facilities etc. as mentioned below this comment. That means I would remove the word "physical".

Standards/scs-0214-v1-k8s-node-distribution.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Decision
+
+This standard formulates the requirement for the distribution of Kubernetes nodes in order
+to provide a fault-tolerant and available Kubernetes cluster infrastructure.
+Since some providers only have small environments to work with and therefore couldn't
+comply with this standard, it will be treated as a RECOMMENDED standard, where providers
+can OPT-OUT.
+
+If the standard is used by a provider, the following decisions are binding and valid:
```
joshmue, line 71: Is "opting-out" of documents defined somewhere?

> it will be treated as a RECOMMENDED standard, where providers can OPT-OUT.

and 

> If the standard is used by a provider, the following decisions are binding and valid:

suggests that whole documents can be non-mandatory.

If that is the case, `MUST`/... wording should be used within the document, because they can opt-out at document-level, right?

If this is not the case, this paragraph could be rephrased to make matters more clear:

```suggestion
This standard formulates the requirement for the distribution of Kubernetes nodes in order
to provide a fault-tolerant and available Kubernetes cluster infrastructure.
Since some providers only have small environments to work with and therefore couldn't
comply with this standard, this standard only contains RECOMMENDATIONS, which are not
strictly required to be implemented.

If the standard is used by a provider, the following decisions are binding and valid:
```

Standards/scs-0214-v1-k8s-node-distribution.md
```diff
@@ -0,0 +1,91 @@
...
+
+- The control plane nodes SHOULD be distributed over multiple physical machines. Kubernetes
+  provides best-practices on this topic, which are also RECOMMENDED by SCS.
+- At least one control plane instance SHOULD be run in each "failure zone", more are
+  RECOMMENDED in each "failure zone" to provide fault-tolerance for each zone.
```
joshmue, line 76: I think that it is important to make this more concrete, even if the linked Kubernetes docs already say that "three or more" machines are required for HA.

For example, if only two "failure zones" are present, hosting one or more etcd members each, this would have no real advantage from a robustness POV because the etcd quorum requires more than half of its members to be online.

So, AFAIU, this standard should mandate at least 3 "failure zones" with at least one "control plane instance" (or etcd member) running inside.
Running more probably should be considered non-mandatory (or "less strictly recommended").

Maybe it should also be considered to enable implementers to split their etcd operations from apiserver/scheduler/controller-manager/... operation, but I understand that this might make things more complicated, while covering an unusual case.

Standards/scs-0214-v1-k8s-node-distribution.md
cah-hbaum: I will write `node` in most of this section now, but just to write why I choose `machine`. This was based on the Kubernetes documentation [here](https://kubernetes.io/docs/concepts/architecture/nodes/), which states physical or virtual machines as the base of a node.

Standards/scs-0214-v1-k8s-node-distribution.md
cah-hbaum: IMO, I would leave the part about `physical machines` in. Even if the concept of a `failure zone` is the important term here to achieve availability security, it's nonetheless dependent on physical separation underneath (since a machine could always fail for some reason). If this wouldn't be defined, a failure zone could just be a few VMs and the next failure zone could be another set of VMs on the same machine. This isn't how failure zones should be used, but it could happen nonetheless.
That's why I would keep the physical seperation in there

Standards/scs-0214-v1-k8s-node-distribution.md
cah-hbaum: As far as I understood the requirements set for this document beforehand, the whole standard is OPT-OUT, especially since some providers could potentially not have the infrastructure required for this.

But you're right, MUST is the right wording here. I needed to reread the RFC definition for MUST vs SHOULD and came to the conclusion, that I used them ambiguously, which is not correct.

Standards/scs-0214-v1-k8s-node-distribution.md
cah-hbaum: I agree with the part about the failure zones; this should be integrated in the standard.

What I'm not sure about is the separation of the etcd operation. I understand where you coming from here, but this adds quite a bit more complexity and also sets even stricter requirements for the providers.
I would like to have a broader discussion about this (especially with the providers) and then maybe put this part in the second version of this standard (or even a separate one, if it comes to this).

Standards/scs-0214-v1-k8s-node-distribution.md
```diff
@@ -0,0 +1,91 @@
...
+Therefore, a Kubernetes cluster in a productive environment should be distributed over
+multiple "failure zones" in order to provide fault-tolerance and high availability.
+This is especially important for the control plane of the cluster, since it contains the
+state of the whole cluster. A failure of this component could mean an unrecoverable failure
+of the whole cluster.
+
+## Design Considerations
+
+Most design considerations of this standard follow the previously written Decision Record
+"Kubernetes Nodes Anti Affinity" as well as the Kubernetes documents about [High Availability]
```
matofeder, line 43: It would be useful to link to this standard https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0213-v1-k8s-nodes-anti-affinity.md

cah-hbaum: You're right, added the link to this standard/DR.

NotTheEvilOne (approved):

PR #355, opened 2023-10-06
Rephrase to accommodate the change of wording from type to scope

### LINKED ISSUE #347: Rename certificate type in scs-0003
According to [scs-0003-v1-sovereign-cloud-standards-yaml.md](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0003-v1-sovereign-cloud-standards-yaml.md) SCS supports two dimensions with respect to certification:

1. _certification level_, of which there are three:
   - SCS-compatible
   - SCS-open
   - SCS-sovereign
2. _cloud layer_, of which there are two:
   - infastructure as a service (IaaS)
   - Kubernetes as a service (KaaS)

A question came up: how do we call the combination of level and cloud layer. There are the following candidates:

- certificate type
- certificate subject
- certificate topic
- certificate scope

**edit** We used this issue to find the new name, and it is now **certificate scope**. Now scs-0003 has to be adapted accordingly.

mbuechse:
(resolves #347)

### DIFF
--- a/Standards/scs-0003-v1-sovereign-cloud-standards-yaml.md
+++ b/Standards/scs-0003-v1-sovereign-cloud-standards-yaml.md
@@ -10,7 +10,7 @@ track: Global
 The [Sovereign Cloud Stack (SCS)](https://scs.community) provides standards for a range of cloud infrastructure types.
 It strives for interoperable and sovereign cloud offerings which can be deployed and used by a wide range of organizations and individuals.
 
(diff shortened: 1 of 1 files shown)


mbuechse:
Once again, the markdown-link-check fails, but the problematic link is just an illustrative example and it need not exist. Please ignore that!

fkr (approved):
LGTM! Thanks for adjusting.

Issue #345, opened 2023-09-21
Implementation of kaas conformance tests to be executed as Sonobuoy plugin


tonifinger:
This issue aims to set up the SCS kaas conformance test infrastructure to be executed by  [sonobuoy](https://sonobuoy.io/docs/v0.56.17/) using the [k8s-cluster-api-provider](https://github.com/SovereignCloudStack/k8s-cluster-api-provider).  The **k8s-cluster-api-provider** already utilizes the sonobuoy e2e conformance tests from the offical [k8s github repository](https://github.com/kubernetes/kubernetes/tree/master/test/e2e). 

Future conformance tests derived from standards could already be part of the k8s e2e tests and could be provided by running the integrated sonobuoy [e2e test plugin](https://sonobuoy.io/docs/v0.56.17/plugins/). However for all tests not already provided by this plugin we have to first write our own tests and second make them executable with sonobuoy. Therefore sonobuoy provides the concept of writing your own plugins.

The main task of a [sonobuoy plugins](https://sonobuoy.io/docs/v0.56.17/plugins/) is to present test results and their status in a consolidated way. This is essentially done by applying the test to a pod that is applied to the k8s cluster under test. A sonobuoy worker then supervises this pod and forwards all test results to the aggregator. It does this by waiting for a "done file" to be created. Once this file is created, it forwards the results to the aggregator, using a predefined location of the results file within the done folder, as seen in following image:
![image search api](https://sonobuoy.io/img/plugin-contract.png)

Therefore, in order to apply the conformance tests as a plugin, we need a tooling/ wrapper to implement the individual test scripts that:
* Gathers all test results and provides them in the results file.
* Run tests in sequence and signal the worker when it is finished by writing the "done file".

Apart from providing the test results, a plugin container must also pass on the status of each test by setting the status flag in the results file.
In addition, to run the tests as a sonobuoy plugin, we need to create an image that can be run inside the k8s cluster under test.

# Suggestion:

There is more than one option to implement tests as Sonobuoy plugin. Two of them might be of most interst to us:

### Option A

Writing our tests in go using the well defined [plugin-helper](https://github.com/vmware-tanzu/sonobuoy-plugins/tree/main/plugin-helper) as skeleton for developing tests.

### Option B

[...]

PR #360, opened 2023-10-16
Add tests for default storage class properties

### LINKED ISSUE #289: Write tests for standard "Default storage class properties"
Write a test for the [Standard](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0211-v1-kaas-default-storage-class.md) "Default storage class properties"

## Definition of Ready:

- [x] Standard for default storage classes has been defined
- [x] A starting point for the tests has been identified

## Definition of Done:

- [x] A tests for "ReadWriteOnce must be a supported [access mode](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes)" exists as compliance check

tonifinger:
This PR adds a test for the K8s default storage class.

A user can either run the test script by providing their Kubernetes cluster config, or use it as a Sonobuoy plug-in, in order to check the default storage class of a cluster.
Following tests are implemented:

* Check if a default storage class is defined
* Check if the default storage class has set a provisioner 
* Check if ReadWriteOnce is set as the supported access mode

This PR also adds configuration files to create the Docker image and plugin description in such a way that they can be executed by sonobuoy.
Closes #289


Tests/kaas/k8s-default-storage-class/run_checks.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/sh
```
anjastrunk, line 1: Why do we need an shell script here. Couldn't this be implemented as python file with even better error handling?

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -0,0 +1,311 @@
...
+logger = logging.getLogger("k8s-default-storage-class-check")
+
+
+def setup_k8s_client(kubeconfigfile=None):
+
+    if kubeconfigfile:
+        logger.debug(f"using kubeconfig file '{kubeconfigfile}'")
+        config.load_kube_config(kubeconfigfile)
+    else:
+        logger.debug(" useing system kubeconfig")
```
anjastrunk, line 43:
```suggestion
        logger.debug(" using system kubeconfig")
```

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -0,0 +1,311 @@
...
+        end="",
+        file=file,
+    )
+
+
+def initialize_logging():
+    logging.basicConfig(format="%(levelname)s: %(message)s", level=logging.INFO)
+
+
+class SCSTestException(Exception):
```
anjastrunk, line 73: Maybe it makes sense to move this kind of code in a global helper.py to 

1. make it available in other test scripts
2. standardize output and exceptions of SCS conformity tests

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -0,0 +1,311 @@
...
+Options:
+ [-k/--kubeconfig PATH_TO_KUBECONFIG] sets kubeconfig file to access kubernetes api
+ [-d/--debug] enables DEBUG logging channel
+""",
+        end="",
+        file=file,
+    )
+
+
+def initialize_logging():
```
anjastrunk, line 69: This could also be a candidate for global test helpers. See [comment below](https://github.com/SovereignCloudStack/standards/pull/360/files#r1366733259)

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -0,0 +1,311 @@
...
+        storage_class_name = item["metadata"]["name"]
+        annotations = item["metadata"]["annotations"]
+
+        if annotations["storageclass.kubernetes.io/is-default-class"] == "true":
+            ndefault_class += 1
+            default_storage_class = storage_class_name
+            provisioner = item["provisioner"]
+
+    if provisioner == "kubernetes.io/no-provisioner":
+        raise SCSTestException(
```
anjastrunk, line 98: I do not favor to exit script at first error. This makes debugging and error fixing really hard for CSPs. I prefer to collect all errors and print an error report as result of the script.

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -0,0 +1,311 @@
...
+
+from kubernetes import client, config
+from helper import gen_sonobuoy_result_file
+
+import logging.config
+
+logger = logging.getLogger("k8s-default-storage-class-check")
+
+
+def setup_k8s_client(kubeconfigfile=None):
```
anjastrunk, line 37: As we will have several k8s conformity tests, I suggest to move this code to a extra module for re-use in other scripts.

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -0,0 +1,311 @@
...
+    k8s_api_client = client.CoreV1Api()
+    k8s_storage_client = client.StorageV1Api()
+
+    return (
+        k8s_api_client,
+        k8s_storage_client,
+    )
+
+
+def print_usage(file=sys.stderr):
```
anjastrunk, line 55: We should unify this method among all conformity tests.

anjastrunk:
We should think about to create some helper classes and method containing source code, that is re-used be several conformity tests. Possible candidates are:

- print_usage()
- set_up_logging()
- Exceptions
- Format of test results

anjastrunk (changes requested):
See my comments inline.

Tests/kaas/k8s-default-storage-class/run_checks.sh
tonifinger: The shell script is derived from the official Sonobuoy [plugin Repository](https://github.com/vmware-tanzu/sonobuoy-plugins).  More precisely from the cmd-runner example: [example /cmd-runner/run.sh](https://github.com/vmware-tanzu/sonobuoy-plugins/blob/main/examples/cmd-runner/run.sh). I did some minor adjustments to this script in order to let it run the python scripts of this issue.
I do agree upon writing this part in python as it would be more beneficial in error handling.
My suggestion would be to address this in following Issue which targets on the integration of Sonobuoy as framework for every kaas conformance tests: #345

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
tonifinger: I moved both to helper.py to be enabled to transfer the helper script to a global point in a later iteration.

tonifinger:
> We should think about to create some helper classes and method containing source code, that is re-used be several conformity tests. Possible candidates would be:
> 
> * print_usage()
> * set_up_logging()
> * setup_k8s_client(kubeconfigfile=None):
> * Exceptions
> * Format of test results

@anjastrunk Thank you for your comments! I have taken up this matter and brought it up in #345. #345 involves creating a framework around all kaas tests.  Therefore, I suggest that this would be the best place for creating a source code structure to provide global helper classes and methods. As preparation, I have moved all the functions listed in the [comment above](https://github.com/SovereignCloudStack/standards/pull/360#issuecomment-1772425977) to the help script so that they can be replaced later with a globalized approach.

anjastrunk (approved):

anjastrunk:
@tonifinger: Please merge.

PR #361, opened 2023-10-17
[Standardization] K8s cluster robustness features (#414)

cah-hbaum:
This commit adds the standard for K8s robustness features, including Kube-API rate limiting, ETCD compaction as well as CA expiration avoidance.

Closes https://github.com/SovereignCloudStack/issues/issues/414

### DIFF
--- a/Standards/scs-0215-v1-robustness-features.md
+++ b/Standards/scs-0215-v1-robustness-features.md
@@ -17,10 +17,10 @@ the SCS standards. These would harden the cluster infrastructure against several
 
 The following special terms are used throughout this standard document:
 
-| Term                        | Meaning                                                        |
-|-----------------------------|----------------------------------------------------------------|
-| Certificate Authority       | Trusted organization that issues digital certificates entities |
-| Certificate Signing Request | Request in order to apply for a digital identity certificate   |
+| Term                        | Abbreviation | Meaning                                                        |
+|-----------------------------|--------------|----------------------------------------------------------------|
+| Certificate Authority       | CA           | Trusted organization that issues digital certificates entities |
+| Certificate Signing Request | CSR          | Request in order to apply for a digital identity certificate   |
 
 ## Motivation
 
@@ -124,48 +124,58 @@ In order to secure the communication of a Kubernetes cluster, (TLS) certificates
 Certificate Authority (CA) can be used.
 Normally, these certificates expire after a set period of time. In order to avoid expiration and failure of a cluster,
 these certificates need to be rotated regularly and at best automatically.
-It is important to either set `--rotate-server-certificates` as a command line parameter or set `rotateCertificates: true`
-in the kubelet config or the `kubeletExtraArgs` of the cluster-template.yaml file. This activates the rotation of the
-kubelet server certificate. Another recommendation is to set `serverTLSBootstrap: true`, which also enables the request
-and rotation of the certificate for the kubelet by requesting them from `certificates.k8s.io`
-according to the documentation under [Kubeadm certs].
+Certificates can either be rotated manually (a reference for manually working with certificates can be found
+[here](https://github.com/kelseyhightower/kubernetes-the-hard-way/blob/79a3f79b27bd28f82f071bb877a266c2e62ee506/docs/04-certificate-authority.md))
+or automatically, which requires other things to care about in a deployment.
 
-A clusters certificates can be rotated by updating the cluster, which according to the Kubernetes documentation
-automatically renews the certificates.
-Another method would be the manual renewal, which can be done through multiple methods, depending on the Kubernetes cluster
-used. An example for a Kubernetes cluster set up with `kubeadm` would be to execute the command
+Some tools or clusters provide possibilities to rotate certificates manually.
+For example, `kubeadm` and `k3s` provides the following commands
 
 ```bash
+# kubeadm
 kubeadm certs renew all
+
+# k3s
+k3s certificate rotate
 ```
 
-Since clusters conformant with the SCS standards would probably be updated within the time period described in the
-standard [SCS-0210-v2](https://github.com/SovereignCloudStack/standards/tree/main/Standards/scs-0210-v2-k8s-version-policy.md),
-this rotation can probably be assumed to happen, because of the cluster update functionality.
+A CA might also expire. Unfortunately, not all Kubernetes tools have functionality to renew these certificates.
+Instead, documentation is provided to manually rotating a CA ([Manual rotation of ca certificate]).
 
-It is also important to mention, that CSRs may need to be approved manually with the commands
+#### Automatic certificate rotation
+
+kubelet can be configured to obtain properly signed certificates from the `certificates.k8s.io` API of Kubernetes.
+To do this, set `serverTLSBootstrap: true` in the configuration file of kubelet, which enables both the certificate request
+during bootstrapping and the rotation mechanism. Setting `rotateCertificates: true` only enables the certificate rotation [Kubeadm certs].
+`--rotate-certificates` or `--rotate-server-certificates` shouldn't be used as command line arguments to set these flags,
+since both parameters are deprecated according to [Certificate rotation].
+
+It is also important to note that some Kubernetes clusters or admin tools provide additional ways to rotate certificates.
+For example, `kubeadm` automatically rotates certificates, if the cluster is updated with the tool (see [Automatic Certificate renewal]).
+This would also mean, that at least `kubeadm`-based clusters can be assumed to rotate their certificates regularly,
+since they would probably be updated within the time period described in the
+standard [SCS-0210-v2](https://github.com/SovereignCloudStack/standards/tree/main/Standards/scs-0210-v2-k8s-version-policy.md).
+
+If an automatic certificate rotation happens, these certificates need to be approved either manually or by a third party 
+controller like the [kubelet csr approver](https://github.com/postfinance/kubelet-csr-approver), which can be deployed on 
+a Kubernetes cluster to automate this process.
+
+A manual approval of these CSRs could be done with the commands
 
 ```bash
 kubectl get csr
 kubectl certificate approve <CSR>
 ```
 
-in order to complete a certificate rotation. This is most likely dependent on the Kubernetes cluster solution in use.
-`kubectl get csr` allows to check, if this is the case; a `Pending` CSR would need to be approved.
+in order to complete a certificate rotation. 
+But it should be noted, that this is also most likely dependent on the Kubernetes cluster solution in use.
+
+`kubectl get csr` allows to check, if a CSR needs to be approved; a `Pending` CSR would need to be approved.
 
 ```bash
 NAME        AGE     SIGNERNAME                        REQUESTOR                      CONDITION
 csr-9wvgt   112s    kubernetes.io/kubelet-serving     system:node:worker-1           Pending
 ```
-
-Another option to approve the CSRs would be to use a third-party controller that automates the process. One example for
-this would be the [kubelet csr approver](https://github.com/postfinance/kubelet-csr-approver), which can be deployed on
-a Kubernetes cluster and requires `serverTLSBootstrap` to be set to true. Other controllers with a similar functionality might
-have other specific requirements, which won't be explored in this document.
-
-Another problem is that the CA might expire. Unfortunately, not all Kubernetes tools have functionality to renew these
-certificates with the help of commands. Instead, there is documentation for manually rotating the CA, which can be found
-under [Manual rotation of ca certificate](https://kubernetes.io/docs/tasks/tls/manual-rotation-of-ca-certificates/).
 
 Further information and examples can be found in the Kubernetes documentation:
 [Kubeadm certs](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/)
@@ -235,11 +245,6 @@ How this is done is up to the operator.
 It should be avoided, that certificates expire either on the whole cluster or for single components.
 To avoid this scenario, certificates MUST be rotated regularly; in the
 case of SCS, we REQUIRE at least a yearly certificate rotation.
-To achieve a complete certificate rotation, the parameters `serverTLSBootstrap` and `rotateCertificates`
-MUST be set in the kubelet configuration.
-
-The certificates can be rotated by either updating the Kubernetes cluster, which automatically
-renews certificates, or by manually renewing them. How this is done is dependent on the used Kubernetes cluster.
 
 It is also RECOMMENDED to renew the CA regularly to avoid an expiration of the CA.
 This standard doesn't set an exact timeline for a renewal, since it is dependent on lifetime and
@@ -254,6 +259,9 @@ therefore expiration date of the CA in question.
 [Upgrade etcd](https://kubernetes.io/docs/tasks/administer-cluster/configure-upgrade-etcd/)
 [Kubeadm certs](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/)
 [Kubelet TLS bootstrapping](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-tls-bootstrapping/)
+[Certificate rotation](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-tls-bootstrapping/#certificate-rotation)
+[Manual rotation of ca certificate](https://kubernetes.io/docs/tasks/tls/manual-rotation-of-ca-certificates/)
+[Automatic Certificate renewal](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#automatic-certificate-renewal)
 
 ## Conformance Tests
 
--- a/Standards/scs-0215-v1-robustness-features.md
+++ b/Standards/scs-0215-v1-robustness-features.md
@@ -143,14 +143,20 @@ Since clusters conformant with the SCS standards would probably be updated withi
 standard [SCS-0210-v2](https://github.com/SovereignCloudStack/standards/tree/main/Standards/scs-0210-v2-k8s-version-policy.md),
 this rotation can probably be assumed to happen, because of the cluster update functionality.
 
-It is also important to mention, that due to security reasons, the CSR needs to be approved manually with the commands
+It is also important to mention, that CSRs may need to be approved manually with the commands
 
 ```bash
 kubectl get csr
 kubectl certificate approve <CSR>
 ```
 
-in order to complete a certificate rotation.
+in order to complete a certificate rotation. This is most likely dependent on the K8s cluster solution in use.
+`kubectl get csr` allows to check, if this is the case; a `Pending` CSR would need to be approved.
+
+```bash
+NAME        AGE     SIGNERNAME                        REQUESTOR                      CONDITION
+csr-9wvgt   112s    kubernetes.io/kubelet-serving     system:node:worker-1           Pending
+```
 
 Another option to approve the CSRs would be to use a third-party controller that automates the process. One example for
 this would be the [kubelet csr approver](https://github.com/postfinance/kubelet-csr-approver), which can be deployed on
--- a/Standards/scs-0215-v1-robustness-features.md
+++ b/Standards/scs-0215-v1-robustness-features.md
@@ -127,12 +127,13 @@ these certificates need to be rotated regularly and at best automatically.
 It is important to either set `--rotate-server-certificates` as a command line parameter or set `rotateCertificates: true`
 in the kubelet config or the `kubeletExtraArgs` of the cluster-template.yaml file. This activates the rotation of the
 kubelet server certificate. Another recommendation is to set `serverTLSBootstrap: true`, which also enables the request
-and rotation of the certificate for the kubelet according to the documentation.
+and rotation of the certificate for the kubelet by requesting them from `certificates.k8s.io` 
+according to the documentation under [Kubeadm certs].
 
-A clusters certificates can either be rotated by updating the cluster, which according to the Kubernetes documentation
+A clusters certificates can be rotated by updating the cluster, which according to the Kubernetes documentation
 automatically renews the certificates.
 Another method would be the manual renewal, which can be done through multiple methods, depending on the K8s cluster
-used. An example for a default K8s cluster would be to execute the command
+used. An example for a K8s cluster set up with `kubeadm` would be to execute the command
 
 ```bash
 kubeadm certs renew all
@@ -140,16 +141,18 @@ kubeadm certs renew all
 
 Since clusters conformant with the SCS standards would probably be updated within the time period described in the
 standard [SCS-0210-v2](https://github.com/SovereignCloudStack/standards/tree/main/Standards/scs-0210-v2-k8s-version-policy.md),
-this rotation can probably be assumed to happen. Nevertheless, the alternative can still be mentioned in the standard.
-Additionally, the Certificate Signing Request (CSR) need to be approved manually due to security reasons with the commands
+this rotation can probably be assumed to happen, because of the cluster update functionality.
+
+It is also important to mention, that due to security reasons, the CSR needs to be approved manually with the commands
 
 ```bash
 kubectl get csr
 kubectl certificate approve <CSR>
 ```
+in order to complete a certificate rotation.
 
 Another option to approve the CSRs would be to use a third-party controller that automates the process. One example for
-this would be the [Kubelet CSR approver](https://github.com/postfinance/kubelet-csr-approver), which can be deployed on
+this would be the [kubelet csr approver](https://github.com/postfinance/kubelet-csr-approver), which can be deployed on
 a K8s cluster and requires `serverTLSBootstrap` to be set to true. Other controllers with a similar functionality might
 have other specific requirements, which won't be explored in this document.
 
@@ -157,7 +160,7 @@ Another problem is that the CA might expire. Unfortunately, not all K8s tools ha
 certificates with the help of commands. Instead, there is documentation for manually rotating the CA, which can be found
 under [Manual rotation of ca certificate](https://kubernetes.io/docs/tasks/tls/manual-rotation-of-ca-certificates/).
 
-Further information can be found in the Kubernetes documentation:
+Further information and examples can be found in the Kubernetes documentation:
 [Kubeadm certs](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/)
 [Kubelete TLS bootstrapping](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-tls-bootstrapping/)
 
@@ -223,7 +226,7 @@ How this is done is up to the operator.
 ### Certificate rotation
 
 It should be avoided, that certificates expire either on the whole cluster or for single components.
-To avoid this scenario, certificates SHOULD be rotated regularly; in the
+To avoid this scenario, certificates MUST be rotated regularly; in the
 case of SCS, we REQUIRE at least a yearly certificate rotation.
 To achieve a complete certificate rotation, the parameters `serverTLSBootstrap` and `rotateCertificates`
 MUST be set in the kubelet configuration.
@@ -234,7 +237,8 @@ renews certificates, or by manually renewing them. How this is done is dependent
 After this, new CSRs MUST be approved manually or with a third-party controller, e.g. the [kubelet-csr-approver](https://github.com/postfinance/kubelet-csr-approver).
 
 It is also RECOMMENDED to renew the CA regularly to avoid an expiration of the CA.
-This standard doesn't set a timeline for this, since it is dependent on the CA.
+This standard doesn't set an exact timeline for a renewal, since it is dependent on lifetime and 
+therefore expiration date of the CA in question.
 
 ## Related Documents
 
--- a/Standards/scs-0215-v1-robustness-features.md
+++ b/Standards/scs-0215-v1-robustness-features.md
@@ -235,9 +235,6 @@ MUST be set in the kubelet configuration.
 The certificates can be rotated by either updating the Kubernetes cluster, which automatically
 renews certificates, or by manually renewing them. How this is done is dependent on the used K8s cluster.
 
-After this, new CSRs MUST be approved manually or with a third-party controller,
-e.g. the [kubelet-csr-approver](https://github.com/postfinance/kubelet-csr-approver).
-
 It is also RECOMMENDED to renew the CA regularly to avoid an expiration of the CA.
 This standard doesn't set an exact timeline for a renewal, since it is dependent on lifetime and
 therefore expiration date of the CA in question.
--- a/Standards/scs-0215-v1-robustness-features.md
+++ b/Standards/scs-0215-v1-robustness-features.md
@@ -1,5 +1,5 @@
 ---
-title: Robustness features for K8s clusters
+title: Robustness features for Kubernetes clusters
 type: Standard
 status: Draft
 track: KaaS
@@ -132,8 +132,8 @@ according to the documentation under [Kubeadm certs].
 
 A clusters certificates can be rotated by updating the cluster, which according to the Kubernetes documentation
 automatically renews the certificates.
-Another method would be the manual renewal, which can be done through multiple methods, depending on the K8s cluster
-used. An example for a K8s cluster set up with `kubeadm` would be to execute the command
+Another method would be the manual renewal, which can be done through multiple methods, depending on the Kubernetes cluster
+used. An example for a Kubernetes cluster set up with `kubeadm` would be to execute the command
 
 ```bash
 kubeadm certs renew all
@@ -150,7 +150,7 @@ kubectl get csr
 kubectl certificate approve <CSR>
 ```
 
-in order to complete a certificate rotation. This is most likely dependent on the K8s cluster solution in use.
+in order to complete a certificate rotation. This is most likely dependent on the Kubernetes cluster solution in use.
 `kubectl get csr` allows to check, if this is the case; a `Pending` CSR would need to be approved.
 
 ```bash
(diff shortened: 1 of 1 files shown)


Standards/scs-0215-v1-robustness-features.md
```diff
@@ -0,0 +1,265 @@
...
+Additionally, the ETCD database should be strengthened with these features in order to provide a secure and robust
+backend for the Kubernetes cluster.
+
+## Design Considerations
+
+In order to provide a conclusive standard, some design considerations need to be set beforehand:
+
+### Kube-API rate limiting
+
+Rate limiting is the practice of preventing too many simultaneous requests to the same server. This can help prevent
```
joshmue, line 38:
```suggestion
Rate limiting is the practice of preventing too many requests to the same server in some time frame. This can help prevent
```

Standards/scs-0215-v1-robustness-features.md
```diff
@@ -0,0 +1,265 @@
...
+[Flow Control](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/)
+
+### etcd compaction/defragmentation
+
+etcd is a strongly consistent, distributed key-value store that provides a reliable way to store data that needs to be
+accessed by a distributed system or cluster of machines. For these reasons, etcd was chosen as the default database
+for Kubernetes.
+In order to remain reliable, an etcd cluster needs periodic maintenance. This is necessary to maintain the etcd keyspace;
+failure to do so could lead to a cluster-wide alarm, which would put the cluster into a limited-operation mode.
+To mitigate this scenario, the etcd keyspace can be compacted and defragmented.
```
joshmue, line 87: I think defragmentation should be dropped from the document. AFAIU, the docs cited below do not indicate that periodic defragmentation would improve reliability (if anything, it disrupts operations for a period of time).

Already had a discussion about this [here](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/401#discussion_r1151471155).

cah-hbaum: After reading the provided link and the docs @matofeder wrote, I understand where you're coming from.
I don't know if I would drop it completely, but instead make the whole defragmentation part optional.

I will rewrite the part and then we can maybe get a third opinion on that.

Standards/scs-0215-v1-robustness-features.md
```diff
@@ -0,0 +1,265 @@
...
+
+etcd needs to be cleaned up regularly, so that it functions correctly and doesn't take
+up too much space, which happens because of its increase of the keyspace.
+
+To compact the etcd keyspace, the following flags/parameters MUST be set for etcd:
+
+* auto-compaction-mode = periodic
+* auto-compaction-retention = 8h
+
+A cluster defragmentation MUST be carried out regularly.
```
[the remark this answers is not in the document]
cah-hbaum: See my answer above.

Standards/scs-0215-v1-robustness-features.md
cah-hbaum: Updated this part.

Standards/scs-0215-v1-robustness-features.md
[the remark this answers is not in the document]
joshmue: :+1: 

Resolving this thread in favor of the above one, then.

Standards/scs-0215-v1-robustness-features.md
joshmue: > I will rewrite the part and then we can maybe get a third opinion on that.

Thanks! Cannot argue with the new revised form :)

Still, if defragmentation is mentioned, I think that the trade-off's of periodic defragmentation should be clear.

AFAIU Pros:
- a cluster that ran out of space for some reason (probably due to subpar disk planning or some unmitigated malicious DoS actions) might "repair" itself with defragmentation at e. g. 2 AM

AFAIU Cons:
- Effort of maintaining defragmentation automation
- Disrupts client operations for a short time (especially if done on all members at the same time)
- If proper disk planning, client limits etc. are in pace, preventing out-of-space conditions, periodic defragmentation has little to no value

matofeder: I would like to comment that even though @joshmue pointed out various drawbacks, I still find defragmentation to be valuable, my points:

-  If an infrastructure monitoring solution detects high disk utilization on the node housing etcd, I, as an OPS engineer, expect that performing maintenance on etcd should resolve the alert for high disk utilization.
 
- In addition, see https://github.com/etcd-io/etcd/issues/14267 which complains that compaction alone does not make free space usable. Also see the comment written by etcd maintainer: https://github.com/etcd-io/etcd/issues/14267#issuecomment-1193404153 

In SCS KaaS we did solid work on how to defragment etcd cluster to avoid cases like @joshmue pointed out: "Disrupts client operations for a short time (especially if done on all members at the same time)" 
- see related docs https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/doc/Maintenance_and_Troubleshooting.md#etcd-maintenance

Standards/scs-0215-v1-robustness-features.md
```diff
@@ -0,0 +1,267 @@
...
+configuration file of the etcd cluster, if it is managed independent from the Kubernetes cluster.
+Possible flags, that can be set for this feature, are:
+
+* auto-compaction-mode
+* auto-compaction-retention
+
+etcd cluster defragmentation unfortunately can't be done automatically. Instead the user would need to manually call
+the defrag command on the cluster. In order to mitigate this, a systemd (or similar) job could be created, which
+periodically calls the defragmentation procedure.
+An example for such a job and its helpers could be the following:
```
matofeder, line 101: Would be worth mentioning that defragmentation to a live etcd member blocks the system from reading and writing data while rebuilding its states. 

Alternatively, (my preferable option), would be great to recommend a defragmentation strategy as is described e.g. [here](https://docs.openshift.com/container-platform/4.9/scalability_and_performance/recommended-host-practices.html#automatic-defrag-etcd-data_recommended-host-practices) or also described and implemented [here](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/doc/Maintenance_and_Troubleshooting.md), 
i.e.:
- Defragment the non leader etcd members first
- Change the leadership to the defragmentation completed etcd member
- Defragment the last (ex-leader) etcd member

Standards/scs-0215-v1-robustness-features.md
joshmue: :+1: 

> `In addition, see Compaction does not make free space usable etcd-io/etcd#14267 which complains that compaction alone does not make free space usable. Also see the comment written by etcd maintainer: Compaction does not make free space usable etcd-io/etcd#14267 (comment)`

I'ts worth adding that this issue is concerned/complains about the documented behavior that defragmentation is required after running out of space - to make the cluster operable again. Hence, it is not strictly related to periodic defragmentation.
I actually do not know how to understand the maintainer's response regarding to usual operation (as well as the OP over there). The mentioned enhancement sounds more like a change in calculation of the space quota - which would block after an out-of-space condition right now (as documented), but (ideally) should not.

Standards/scs-0215-v1-robustness-features.md
```diff
@@ -0,0 +1,267 @@
...
+### CA expiration avoidance
+
+In order to secure the communication of a Kubernetes cluster, (TLS) certificates signed by a controlled
+Certificate Authority (CA) can be used.
+Normally, these certificates expire after a set period of time. In order to avoid expiration and failure of a cluster,
+these certificates need to be rotated regularly and at best automatically.
+It is important to either set `--rotate-server-certificates` as a command line parameter or set `rotateCertificates: true`
+in the kubelet config or the `kubeletExtraArgs` of the cluster-template.yaml file. This activates the rotation of the
+kubelet server certificate. Another recommendation is to set `serverTLSBootstrap: true`, which also enables the request
+and rotation of the certificate for the kubelet according to the documentation
```
matofeder, line 151: IMO I find the explanation regarding the rotation of kubelet server certificates somewhat unclear. In addition, according to the documentation, it appears that the approval of Certificate Signing Requests (CSRs) should be done either manually or by using a custom controller, as indicated in this note [here](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-tls-bootstrapping/#certificate-rotation).

Overall, I do not see the recommendation to set the kubelet parameters to automate server certificate rotation as a complete solution because it is just a part of Kubernetes [PKI](https://kubernetes.io/docs/setup/best-practices/certificates/). 

Instead, I would prefer 2 approaches you described below, i.e. 

1. Update the cluster at least once a year to rotate certificates automatically
2.  Renew all certificates with `kubeadm certs renew all`
     - You only need to do this when you don't upgrade your cluster
     
the above is also in sync with SCS KaaS reference implementation, see [docs](https://github.com/SovereignCloudStack/k8s-cluster-api-provider/blob/main/doc/Maintenance_and_Troubleshooting.md#client-certificates-in-kubernetes-expire-after-one-year)

Standards/scs-0215-v1-robustness-features.md
cah-hbaum: I updated this section to include your recommendations. I already read them before and they make sense, so I don't really know why I didn't include them before.

Standards/scs-0215-v1-robustness-features.md
cah-hbaum: I probably wrote this section a bit unclear then and I also probably understand something wrong in the certificate rotation documentation (and some other documents).
I planned this section to basically recommend the cluster renewal or certificate renewal approaches you described. But I was under the impression, that the flags described in the first subsection are required to have all the certificates exchanged correctly.

Standards/scs-0215-v1-robustness-features.md
cah-hbaum: > It's worth adding that this issue is concerned/complains about the documented behavior that defragmentation is required after running out of space - to make the cluster operable again. Hence, it is not strictly related to periodic defragmentation.

You're right that it's not strictly related to periodic defragmentation, but if the "free space" isn't usable after compaction without defragmentation, I would argue that compaction isn't necessarily that useful all together.
I would still prefer to do it, but it wouldn't have the same "value" as before.

What I would also like to know (regarding the previous comment), how exactly "disk planning" could help prevent out-of-space conditions. I guess forward planning is important in this regard, meaning that an etcd cluster should always have enough disk space to accompany the possible growth. I guess the question here would be how much "disk" a provider would be willing to put into a cluster as a kind of security so that no out-of-space would occur and what scenarios will be expected. From my experience in the field people won't have an eye on everything and can't prevent every problem before it happens. So I think such an automation would probably be a good thing.

I also added @matofeder recommendation about the avoidance of service disruption.

Standards/scs-0215-v1-robustness-features.md
matofeder: > But I was under the impression, that the flags described in the first subsection are required to have all the certificates exchanged correctly.

After rereading the docs (again), you are right. Kubeadm does not manage the kubelet cert rotation. 
Based on this [docs](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#kubelet-serving-certs), my understanding is as follows:

1. The command `kubeadm certs renew all` will renew Kubernetes control plane certificates, but it won't renew the certificates used for authenticating the kubelet against the k8s API server and won't renew the kubelet server certificate
2. The kubelet parameter `rotateCertificates` (defaults to true) will enable the full automatic renewal of kubelet client certificate for authenticating the kubelet against the k8s API
3.  The kubelet parameter  `TLSBootstrap: true` will enable the bootstrap of kubelet server certificates by requesting them from the certificates.k8s.io API. 
 ! One known limitation is that the CSRs (Certificate Signing Requests) for these certificates cannot be automatically approved. This will require manual action from the user or a third-party controller like [kubelet-csr-approver](https://github.com/postfinance/kubelet-csr-approver)

cah-hbaum: I updated the section a bit more to include the possibility for a CSR approver controller and also a better representation for the `kubeadm certs renew all` command.

matofeder (approved):
LGTM

Standards/scs-0215-v1-robustness-features.md
joshmue: > You're right that it's not strictly related to periodic defragmentation, but if the "free space" isn't usable after compaction without defragmentation, I would argue that compaction isn't necessarily that useful all together.
> I would still prefer to do it, but it wouldn't have the same "value" as before.

Yes, compaction alone does not seem to help at all if the storage quota was reached and etcd does not allow any more writes. My point was just that this is a very different situation than running a cluster in normal operation, when compaction alone usually would prevent (combined with other limits or just adequately little load) the storage quota to be reached.

Maybe there are indeed cases where defragmentation really frees up space that would not be used otherwise - e. g. if there are many tiny gaps in the backend database, but all new data is bigger than any gap - like the OP of the mentioned issue theorized.
I know that the [OpenShift docs](https://docs.openshift.com/container-platform/4.14/scalability_and_performance/recommended-performance-scale-practices/recommended-etcd-practices.html#etcd-defrag_recommended-etcd-practices) strongly recommend it and implement it as automatic process, but I do not really understand why. Also seems to be [sometimes](https://access.redhat.com/solutions/6840041) [problematic](https://access.redhat.com/solutions/6960380) for such installations.

It's just that the documentation AFAIU does not hint that *periodic* defragmentation is of any advantage or recommended. Nor do I know of any POC or case (either by reading about it or by experience) where it was actually needed. The only plain statement I found was from a former etcd maintainer in https://github.com/etcd-io/etcd/issues/7607#issuecomment-289351605:

> This is actually about disk fragmentation. But if you do not suddenly remove a lot of keys and want to reclaim the disk space immediately, you do not need to defrag.

I'm not strictly opposed to adding docs how to do (periodic) defragmentation on etcd - just want everyone to have the same understanding on why or why not it should be mentioned/recommended. Also, if good reasons are found for implementing automatic defragmentation, the recommendation should also go into the etcd documentation.

---

> What I would also like to know (regarding the previous comment), how exactly "disk planning" could help prevent out-of-space conditions. I guess forward planning is important in this regard, meaning that an etcd cluster should always have enough disk space to accompany the possible growth. I guess the question here would be how much "disk" a provider would be willing to put into a cluster as a kind of security so that no out-of-space would occur and what scenarios will be expected. From my experience in the field people won't have an eye on everything and can't prevent every problem before it happens. So I think such an automation would probably be a good thing.

Again (AFAIU and under the assumption that etcd generally behaves as documented), defragmentation generally does not prevent out-of-space conditions.
In contrast: If disk planning ensures that etcd has always at least 2GB (or whatever storage size quota is configured) disk space available, the disk will not run out of space, as etcd hits its own quota first (AFAIU, assuming that the storage quota equals database file size).

cah-hbaum: For now, I wrote the part about the defragmentation as completely optional into the `Decision` section.

joshmue: Thanks for [adding](https://github.com/SovereignCloudStack/standards/compare/d8b01baa2aacc34c8f4eeabaa93d5a9d254b60de..e6583e9428e12a3aaddf1e4b99a0a80e2f49922c) the stability risks of defragmentation. :+1: 

> If thats not ok as it is, please make an edit to the text how your preferences would be for that section.

My only strong preference is really just to include *why* periodic defragmentation [^1] may be a good idea for cluster robustness and hence belong in this document about "K8s cluster robustness features".
If no reason is strong enough to be included in the document itself, I'd opt to remove the paragraphs about it completely (and maybe add a paragraph about guaranteeing disk space to etcd).

[^1]: As opposed to e. g. manual defragmentation as part of an operational handbook in case of an out-of-space condition

cah-hbaum: I removed most of the defragmentation parts for now (except a very small mentioning in the `etcd maintenance` section).

I also decided to not add a paragraph about disk space guarantees for now.

Standards/scs-0215-v1-robustness-features.md
```diff
@@ -0,0 +1,250 @@
...
+## Motivation
+
+A typical productive Kubernetes cluster could be hardened in many different ways, whereas probably many of these actions
+would overlap and target similar weaknesses of a cluster.
+For this version of the standard, the following points should be addressed:
+
+* Kube-API rate limiting
+* etcd compaction/defragmentation
+* etcd backup
+* CA expiration avoidance
```
anjastrunk, line 25: What does "CS" means?

Standards/scs-0215-v1-robustness-features.md
```diff
@@ -0,0 +1,250 @@
...
+
+These robustness features should mainly increase the up-time of the Kubernetes cluster by avoiding downtimes either
+because of internal problems or external threads like "Denial of Service" attacks.
+Additionally, the ETCD database should be strengthened with these features in order to provide a secure and robust
```
anjastrunk, line 29: In line 23 and 24 "etcd" is written in lower case and here in upper We should be consistent throughout the whole document.

Standards/scs-0215-v1-robustness-features.md
```diff
@@ -0,0 +1,250 @@
...
+
+A clusters certificates can either be rotated by updating the cluster, which according to the Kubernetes documentation
+automatically renews the certificates, or by running the kubeadm certs renew command for the certificates that need to
+updated, which can be seen in the following inline code
+
+```bash
+kubeadm certs renew all
+```
+
+Since clusters conformant with the SCS standards would probably be updated within a 14 month time period, this
```
anjastrunk, line 131: I would refer and link the standard, which dictates an update period of 14 months, here.

Standards/scs-0215-v1-robustness-features.md
```diff
@@ -0,0 +1,250 @@
...
+rotation can probably be assumed to happen. Nevertheless, the alternative can still be mentioned in the standard.
+Additionally, the CSR needs to be approved manually due to security reasons with the commands
+
+```bash
+kubectl get csr
+kubectl certificate approve <CSR>
+```
+
+Another option to approve the CSRs would be to use a third-party controller that automates the process. One example for
```
anjastrunk, line 140: Please do not use abbreviations. without introducing them. Additionally, I recommend to add a glossary at the beginning to the document, like in [scs-0302-v1-domain-manager-role.md](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0302-v1-domain-manager-role.md)

Standards/scs-0215-v1-robustness-features.md
```diff
@@ -0,0 +1,250 @@
...
+  type: Server
+```
+
+It is also RECOMMENDED to activate the Kubernetes API priority and fairness feature,
+which also uses the aforementioned cluster parameters to better queue, schedule and
+prioritize incoming requests.
+
+### etcd compaction
+
+etcd needs to be cleaned up regularly, so that it functions correctly and doesn't take
```
anjastrunk, line 193: I guess, etcd cleanup is a must and we should phrase it like that.

```suggestion
etcd must to be cleaned up regularly, so that it functions correctly and doesn't take
```

Standards/scs-0215-v1-robustness-features.md
```diff
@@ -0,0 +1,250 @@
...
+up too much space, which happens because of its increase of the keyspace.
+
+To compact the etcd keyspace, the following flags/parameters MUST be set for etcd:
+
+* auto-compaction-mode = periodic
+* auto-compaction-retention = 8h
+
+### etcd backup
+
+An etcd cluster MUST be backed up regularly. It is RECOMMENDED to adapt
```
anjastrunk, line 203: etcd MUST be back up regularly, I agree. However, this requirement is senseless, without defining a backup strategy. IMO it is a huge difference in robustness, if a provider backups monthly or daily. I recommend to add a minimal backup cycle in this standard, a customer can expect for k8s cluster, that are SCS certified.

Standards/scs-0215-v1-robustness-features.md
```diff
@@ -0,0 +1,250 @@
...
+In order to do so, at least the following parameters SHOULD be set on a Kubernetes cluster:
+
+* max-requests-inflight
+* max-mutating-requests-inflight
+* min-request-timeout
+
+Values for these flags/parameters SHOULD be adapted to the needs of the environment and
+the expected load.
+
+A cluster MUST also activate and configure a Ratelimit admission controller.
```
anjastrunk, line 173: Why is limitation of number of requests a SHOULD but Ratelimit admission controller a MUST. Isn't Ratelimit admission controller a special case for the limitation of request as Kube API? If so, we should set number of request limitation to MUST. Otherwise line 161ff and link 173ff are contradictory.

Standards/scs-0215-v1-robustness-features.md
cah-hbaum: CA = Certificate Authority

Standards/scs-0215-v1-robustness-features.md
cah-hbaum: Made it consistent across the document.

Standards/scs-0215-v1-robustness-features.md
cah-hbaum: Referred to the standard with a link here, but the referred standard isn't in the main branch yet. I nevertheless made the reference link use the main branch, since the standard will be merged soon (hopefully).

Standards/scs-0215-v1-robustness-features.md
cah-hbaum: There is a recommendation for the backup strategy after that which refers to the previous section.
I wouldn't further define (or make this subsection a MUST), since not all providers can do a backup (or multiple backups) every day.
I would maybe put in a weekly backup requirement, but more would be too fine-grained and defined for some providers.

Standards/scs-0215-v1-robustness-features.md
cah-hbaum: Set both lines to MUST.

Standards/scs-0215-v1-robustness-features.md
cah-hbaum: Added a glossary and put the abbreviations at the first occurrence of their respective meanings.

cah-hbaum:
I think I fixed everything in here for now. Please review again and then we can merge.

Standards/scs-0215-v1-robustness-features.md
```diff
@@ -0,0 +1,260 @@
...
+Additionally, the etcd database should be strengthened with these features in order to provide a secure and robust
+backend for the Kubernetes cluster.
+
+## Design Considerations
+
+In order to provide a conclusive standard, some design considerations need to be set beforehand:
+
+### Kube-API rate limiting
+
+Rate limiting is the practice of preventing too many requests to the same server in some time frame. This can help prevent
```
anjastrunk, line 47:
```suggestion
Rate limiting is the practice of preventing too many requests to the same server in some time frame. This can help to prevent
```

anjastrunk:
There are three open threads left. After they are solved, I will give my approval.

Standards/scs-0215-v1-robustness-features.md
cah-hbaum: Both sentences are correct and I would like to keep the original one, since it sounds better in this context (but it doesn't make a difference meaning wise)

Standards/scs-0215-v1-robustness-features.md
```diff
@@ -0,0 +1,252 @@
...
+```
+
+After this, new CSRs MUST be approved manually, normally done with
+
+```bash
+kubectl get csr
+kubectl certificate approve <CSR>
+```
+
+or be approved with a third-party controller, e.g. the [kubelet-csr-approver](https://github.com/postfinance/kubelet-csr-approver).
```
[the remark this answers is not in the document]
cah-hbaum: I changed a few sentences in the document and especially in the `Decision` section in order to better reflect a few of the concerns you had. You're right that `kubeadm` was a bit to specific, so I only mentioned it as an example in the `Design Considerations`.

About the CSR part, I'm not really sure. Yes they don't HAVE to be used, but https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#kubelet-serving-certs reads like its a recommended way, especially if some external tools should be used with the cluster (like monitoring solutions).
But I'm really not an expert on this.

jschoone:
@SovereignCloudStack/vp05 please review as we want to achieve the standards in our reference implementation

anjastrunk (approved):
LGTM

Standards/scs-0215-v1-robustness-features.md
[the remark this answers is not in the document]
cah-hbaum: @joshmue Is this fine for now or are there any other problems with this?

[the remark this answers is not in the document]
joshmue: > You're right that kubeadm was a bit to specific, so I only mentioned it as an example in the Design Considerations.

:+1: 

> About the CSR part, I'm not really sure. [...] But I'm really not an expert on this.

I did not dig into that to much right now, so I'm not totally sure whether this should be mandated here, either (I'd tend not to).
If this is to be included and merged, someone should be sure about that, though.

> Yes they don't HAVE to be used, but https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#kubelet-serving-certs reads like its a recommended way, especially if some external tools should be used with the cluster (like monitoring solutions).

Again, that is kubeadm-centric documentation, so not necessarily relevant for other distributions.

Other than that, the current status has some SHOULD/can/MUST ambiguity problems:

> To avoid this scenario, certificates SHOULD be rotated regularly; in the
> case of SCS, we REQUIRE at least a yearly certificate rotation.

That could be understood two ways: Either someone could decide to be non-compliant with that SHOULD, hence making the REQUIRE-ment ineffective (That again would mandate *how* something is done, while the *what* is optional - pretty much the opposite of what standards usually do). Or someone could read the REQUIRE as actual requirement, making the SHOULD an effective MUST.

> The certificates can be rotated by either updating the Kubernetes cluster, which automatically
> renews certificates, or by manually renewing them. How this is done is dependent on the used K8s cluster.
> 
> After this, new CSRs MUST be approved manually or with a third-party controller, e.g. the [kubelet-csr-approver](https://github.com/postfinance/kubelet-csr-approver).

Kind of the same problem as above.

> It is also RECOMMENDED to renew the CA regularly to avoid an expiration of the CA.

:+1: 

> This standard doesn't set a timeline for this, since it is dependent on the CA.

Is this intended to say that the CA may be out of the cluster operator's control? That's not really clear.

matofeder (approved):
LGTM 
I've read this updated version again, and I like this format, i.e. I as a consumer of SCS standard I can see what exactly I should do to have an "SCS-compatible" k8s cluster   :+1:

Standards/scs-0215-v1-robustness-features.md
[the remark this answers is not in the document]
cah-hbaum: > > Yes they don't HAVE to be used, but https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#kubelet-serving-certs reads like its a recommended way, especially if some external tools should be used with the cluster (like monitoring solutions).
>
> Again, that is kubeadm-centric documentation, so not necessarily relevant for other distributions.

Yes this documentation uses `kubeadm`, but I would argue that the CSR part is not-centered especially on `kubeadm` and can be applied to other K8s cluster "types" aswell.

> > To avoid this scenario, certificates SHOULD be rotated regularly; in the
> > case of SCS, we REQUIRE at least a yearly certificate rotation.
>
> That could be understood two ways: Either someone could decide to be non-compliant with that SHOULD, 
> hence making the REQUIRE-ment ineffective (That again would mandate *how* something is done, while the 
> *what* is optional - pretty much the opposite of what standards usually do). Or someone could read the 
> REQUIRE as actual requirement, making the SHOULD an effective MUST.
>
> > The certificates can be rotated by either updating the Kubernetes cluster, which automatically
> > renews certificates, or by manually renewing them. How this is done is dependent on the used K8s cluster.
> > 
> > After this, new CSRs MUST be approved manually or with a third-party controller, e.g. the [kubelet-csr-approver](https://github.com/postfinance/kubelet-csr-approver).
>
> Kind of the same problem as above.

I agree with the first part (and I fixed that), but the second part doesn't have the same problem.
The whole part is about certificate rotation (meaning its continuous), so the part with the `can` is refering to the certificate rotation part, which is written as a `MUST` now (`SHOULD` before, so that would be contradictive).
Also `can` isn't an RFC keyword (https://www.rfc-editor.org/rfc/rfc2119), so it's not defining anything, just explaining a process from this.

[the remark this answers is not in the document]
cah-hbaum: Updated with https://github.com/SovereignCloudStack/standards/pull/361/commits/84fcef3ff4670d82ac28ab5cbf85caf515678a01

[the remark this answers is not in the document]
joshmue: > > > Yes they don't HAVE to be used, but https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#kubelet-serving-certs reads like its a recommended way, especially if some external tools should be used with the cluster (like monitoring solutions).
> >
> > Again, that is kubeadm-centric documentation, so not necessarily relevant for other distributions.
> 
> Yes this documentation uses `kubeadm`, but I would argue that the CSR part is not-centered especially on `kubeadm` and can be applied to other K8s cluster "types" aswell.

I did not really use `kubeadm` all that much for a few years now. But looking at its [docs](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#renew-certificates-with-the-kubernetes-certificates-api) (mind the `Caution` paragraph), handling certificates this way seems to be the exception even for `kubeadm` setups, nevermind other distributions.

> I agree with the first part (and I fixed that)

:+1: 

> The whole part is about certificate rotation (meaning its continuous), so the part with the can is refering to the certificate rotation part, which is written as a MUST now (SHOULD before, so that would be contradictive).
> Also can isn't an RFC keyword (https://www.rfc-editor.org/rfc/rfc2119), so it's not defining anything, just explaining a process from this.

Still, this MUST refers to `how` certificates should be approved. In some setups, there probably is not even a concept for approval present.

And if there is, it is buried so low that it is unpractical to check for.
For example: Could [k3s certificate rotation](https://docs.k3s.io/cli/certificate) ever be compliant to this standard's MUST? That is cumbersome to (dis-)prove and does not add much value over the simple requirement of "Certificates should be rotated regularly".

Standards/scs-0215-v1-robustness-features.md
```diff
@@ -0,0 +1,257 @@
...
+this rotation can probably be assumed to happen, because of the cluster update functionality.
+
+It is also important to mention, that due to security reasons, the CSR needs to be approved manually with the commands
+
+```bash
+kubectl get csr
+kubectl certificate approve <CSR>
+```
+
+in order to complete a certificate rotation.
```
joshmue, line 153: Related to https://github.com/SovereignCloudStack/standards/pull/361#discussion_r1448822622

> It is also important to mention, that due to security reasons, the CSR needs to be approved manually with the commands

Sure? https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#manual-certificate-renewal does not seem to mention it.

Standards/scs-0215-v1-robustness-features.md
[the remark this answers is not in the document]
cah-hbaum: > > The whole part is about certificate rotation (meaning its continuous), so the part with the can is refering to the > > certificate rotation part, which is written as a MUST now (SHOULD before, so that would be contradictive).
> >Also can isn't an RFC keyword (https://www.rfc-editor.org/rfc/rfc2119), so it's not defining anything, just explaining a process from this.
>
> Still, this MUST refers to how certificates should be approved. In some setups, there probably is not even a
> concept for approval present.
>
> And if there is, it is buried so low that it is unpractical to check for.
> For example: Could [k3s certificate rotation](https://docs.k3s.io/cli/certificate) ever be compliant to this 
> standard's MUST? That is cumbersome to (dis-)prove and does not add much value over the simple requirement
> of "Certificates should be rotated regularly".

Removed the sentence in question.

Standards/scs-0215-v1-robustness-features.md
cah-hbaum: Rewrote the section to hopefully fit better.

cah-hbaum: Summoning @joshmue again to resolve (and maybe approve)

joshmue: The rewording from hard "needs to" etc. to "may need to" reduces the risk of strictly false statements, indeed.

My point (from https://github.com/SovereignCloudStack/standards/pull/361#discussion_r1447407033) still is that information should be verified with a high degree of certainty, which I do not feel there is:

> > About the CSR part, I'm not really sure. [...] But I'm really not an expert on this.
> >
> I did not dig into that to much right now, so I'm not totally sure whether this should be mandated here, either (I'd tend not to).
> If this is to be included and merged, someone should be sure about that, though.

cah-hbaum: So I read up on most of the stuff again (https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs, https://kubernetes.io/docs/reference/access-authn-authz/kubelet-tls-bootstrapping, https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/) and came to the conclusion, to rewrite the `Certificate rotation` section, since there were quite a few errors and mixups in this part. IMO its now better structured and correct.

I also shortened the section related to this under `Decision` to the relevant stuff.

joshmue: Looks way better to me, as well! :+1:

jschoone (approved):

Standards/scs-0215-v1-robustness-features.md
```diff
@@ -0,0 +1,268 @@
...
+
+Some tools or clusters provide possibilities to rotate certificates manually.
+For example, `kubeadm` and `k3s` provides the following commands
+
+```bash
+# kubeadm
+kubeadm certs renew all
+
+# k3s
+k3s certificate rotate
```
jschoone, line 139:
```suggestion
For example, `kubeadm` provides the following command

```bash
# kubeadm
kubeadm certs renew all
```
Remove k3s to avoid confusion since we don't use this in KaaS

jschoone (changes requested):

Standards/scs-0215-v1-robustness-features.md
cah-hbaum: I can do that, but I thought more about this a general example. Maybe "we" don't use it, but what if someone external wants to make his k3s cluster SCS conform?
Or do we generally say, that this isn't a thing?

jschoone: @cah-hbaum ok let's keep it. Since there is a k3s providing Kubernetes on IaaS, it's probably not too far away from us.

cah-hbaum: Dont know how to "resolve" your changes if we don't integrate them, maybe you need to approve or something similar.

jschoone (approved):

PR #364, opened 2023-10-20
Forbid using CPU generation without CPU vendor

### LINKED ISSUE #363: [BUG] Extension syntax in flavor naming standard misleading
SCS standard for [Server Flavors](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0100-v3-flavor-naming.md#generation-and-vendor) does not dictate to use suffix for CPU generation **always** in conjunction with suffix for CPU vendor. As it doesn't makes sense to use suffix for CPU generation without suffix of CPU vendor, SCS Flavor Standard should forbid this explicitly.

anjastrunk:
closes #363


anjastrunk:
> This way it is very difficult to see what has changed. Can you please add a diff (v3 vs v4) as a comment?

```
git diff --no-index scs-0100-v3-flavor-naming.md scs-0100-v4-flavor-naming.md 
diff --git a/scs-0100-v3-flavor-naming.md b/scs-0100-v4-flavor-naming.md
index 4d4447a..b1823b0 100644
--- a/scs-0100-v3-flavor-naming.md
+++ b/scs-0100-v4-flavor-naming.md
@@ -1,10 +1,10 @@
 ---
 title: SCS Flavor Naming Standard
 type: Standard
-status: Stable
-stabilized_at: 2023-06-14
+status: Draft
+stabilized_at: 
 track: IaaS
-replaces: scs-0100-v2-flavor-naming.md
+replaces: scs-0100-v3-flavor-naming.md
 ---
 
 ## Introduction
@@ -360,6 +360,7 @@ The generations are vendor specific and can be left out.
 Not specifying arch means that we have a generic CPU (**x86-64**).
 The letters `i`, `z`, `a` and `r` specify the vendors Intel,
 AMD (`z` like in Zen), ARM v8+, RISC-V.
+Each generation number MUST be prefixed by a letter for CPU vendor. It is forbidden to just use a generation number.
 
 | Generation | i (Intel x86-64) | z (AMD x86-64) |  a (AArch64)       | r (RISC-V) |
 | ---------- | ---------------- | -------------- | ------------------ | ---------- |
@@ -394,6 +395,7 @@ out when generating the name for comparison. In other words: 0 has a meaning of
 - SCS-2C-4-10n_bms_**z3**
 - SCS-2C-4-10n_bms_**z3h**
 - SCS-2C-4-10n_bms_**z3hh** <- Bare Metal, Intel Ice Lake with > 3.25GHz all core freq
+- ~~SCS-2C-4-10n\_**3**~~ <- illegal, missing letter for CPU vendor
 
 ### [OPTIONAL] GPU support
```

garloff:
You are right, we are missing a closing bracket, as you say. (You can compare the python tool output if you like.)
And yes, clarifications and fixed typos should not make us change the version number. (Unless the typo resulted in wrong names out in the wild.)

anjastrunk:
> You are right, we are missing a closing bracket, as you say. (You can compare the python tool output if you like.) And yes, clarifications and fixed typos should not make us change the version number. (Unless the typo resulted in wrong names out in the wild.)

All right, I will update PR accordingly.

garloff (approved):
Looks good. Though reading, I spotted a mistake in line 410 (which is unchanged by you). I'll submit a separate PR for it unless you want to address it (and change a `z` to an `i` or adjust the description to read AMD Milan instead of Intel Ice Lake.

PR #370, opened 2023-11-01
Add proposal OS API and OS internal Traffic TLS encryption and PKI

fdobrovolny:
https://github.com/SovereignCloudStack/issues/issues/457

### DIFF
--- a/Standards/scs-XXXX-v1-encrypted-internal-openstack-traffic-and-pki-infra.md
+++ b/Standards/scs-XXXX-v1-encrypted-internal-openstack-traffic-and-pki-infra.md
@@ -1,0 +1,260 @@
+---
+title: OpenStack API and OpenStack internal Traffic TLS encryption and PKI Infrastructure
+type: Decision Record
+status: Draft
+track: IaaS
+---
+
+## Introduction
+
+OpenStack is a platform used for the SCS IaaS layer (Infrastructure as a Service).
+Currently, the deployment is mostly done using [Kolla](https://opendev.org/openstack/kolla-ansible).
+This ADR document current status of Kolla in regard to end-to-end TLS encryption of selected OpenStack services shown below.
+As implementation of most of the following measures is quite straight forward the outcome of this document is divided into two parts.
+1. Proposal for implementation of configuration changes in Kolla
+2. More common ADR considering adoption of PKI infrastructure.
+
+### Services
+
+This document will focus on the following services as they are the default ones in our IaaS layer.
+
+
+1. **Keystone** - Keystone is the identity service in OpenStack, responsible for user authentication and authorization and managing domains, projects, and roles.
+
+2. **Horizon** - Horizon is the dashboard of OpenStack, providing a web-based user interface to other services, allowing administrators and users to manage resources and services.
+
+3. **Glance** - Glance is the image service in OpenStack, responsible for discovering, registering, and retrieving virtual machine images.
+
+4. **Cinder** - Cinder is the block storage service, providing persistent block storage resources for virtual machine instances.
+
+5. **Placement** - Placement is a service for tracking resource provider inventories and usages, helping in the efficient allocation of resources across the cluster.
+
+6. **Nova** - Nova is the compute service in OpenStack, responsible for creating and managing virtual machines and compute instances.
+
+7. **Neutron** - Neutron is the networking service providing network connectivity between interface devices managed by other OpenStack services.
+
+8. **Heat** - Heat is the orchestration service, allowing developers to define and deploy composite cloud applications using templates.
+
+9. **Memcached** - Memcached is a caching layer in OpenStack, improving performance by alleviating database load.
+
+10. **MariaDB** - MariaDB is the database backend for OpenStack services, storing and managing data efficiently.
+
+11. **RabbitMQ** - RabbitMQ is the messaging service used in OpenStack for communication between the different components and services.
+
+12. **Redis** - Redis is a database, cache, and message broker within OpenStack, ensuring high-performance data management.
+
+13. **OpenvSwitch** - OpenvSwitch is a multilayer virtual switch implemented to ensure effective network automation in an OpenStack environment.
+
+
+### Current state
+
+#### Non-internal OpenStack endpoints
+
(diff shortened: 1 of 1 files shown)


Standards/scs-XXXX-v1-encrypted-internal-openstack-traffic-and-pki-infra.md
```diff
@@ -0,0 +1,260 @@
...
+title: OpenStack API and OpenStack internal Traffic TLS encryption and PKI Infrastructure
+type: Decision Record
+status: Draft
+track: IaaS
+---
+
+## Introduction
+
+OpenStack is a platform used for the SCS IaaS layer (Infrastructure as a Service).
+Currently, the deployment is mostly done using [Kolla](https://opendev.org/openstack/kolla-ansible).
```
fkr, line 11:
```suggestion
Currently, the deployment is done via [OSISM](https://github.com/osism/) which makes heavy use of [Kolla](https://opendev.org/openstack/kolla-ansible).
```

Standards/scs-XXXX-v1-encrypted-internal-openstack-traffic-and-pki-infra.md
```diff
@@ -0,0 +1,260 @@
...
+This ADR document current status of Kolla in regard to end-to-end TLS encryption of selected OpenStack services shown below.
```
fkr, line 12:
```suggestion
This ADR documents the current status of Kolla in regard to end-to-end TLS encryption of selected OpenStack services shown below.
```

fkr:
@artificial-intelligence can you read this and share your thoughts?

Standards/scs-XXXX-v1-encrypted-internal-openstack-traffic-and-pki-infra.md
```diff
@@ -0,0 +1,260 @@
...
+2. Horizon
+3. Glance
+4. Cinder
+5. Placement
+6. Nova
+7. Netron
+8. Heat
+13. OpenVSwitch
+
+Only RabbitMQ is not currently supported.
```
artificial-intelligence, line 102: there is interest in adding an internal CA for rabbitmq and libvirt TLS support, see:

https://etherpad.opendev.org/p/kolla-caracal-ptg#L283

the problematic part currently is, choosing the appropriate CA software, which has many challenges, as can be seen in the linked document where we evaluated various alternatives.

Standards/scs-XXXX-v1-encrypted-internal-openstack-traffic-and-pki-infra.md
```diff
@@ -0,0 +1,260 @@
...
+
+[Details](https://docs.openstack.org/kolla-ansible/latest/admin/tls.html#back-end-tls-configuration)
```
artificial-intelligence, line 104:
```suggestion
[Details](https://docs.openstack.org/kolla-ansible/2023.1/admin/tls.html#back-end-tls-configuration)
```

Standards/scs-XXXX-v1-encrypted-internal-openstack-traffic-and-pki-infra.md
```diff
@@ -0,0 +1,260 @@
...
+(External)   (public internet) (HAProxy) (OpenStack Networking)   (OpenStack Services)  (OpenStack Networking)  (Data Service)
+   |                                                                  |      |      |                            |         |
+ Client                                                           Keystone  Glance  Placement ...               MariaDB   Redis ...
+```
+
+To provide truly end-to-end traffic encryption within our infrastructure, all communications between services and their respective databases and brokers must be conducted via TLS (Transport Layer Security). This measure not only encrypts the data in transit but also ensures the communication channels are authenticated and integrity-protected.
+
+##### Memcached
+
+Memcached does support TLS; however, this support is only [experimental](https://github.com/memcached/memcached/commit/ee1cfe3bf9384d1a93545fc942e25bed6437d910), and custom build binaries are needed. For our use case, this is a no-go.
```
artificial-intelligence, line 119: so what do we do with this?

Standards/scs-XXXX-v1-encrypted-internal-openstack-traffic-and-pki-infra.md
```diff
@@ -0,0 +1,260 @@
...
+
+##### Redis
+
+Redis supports TLS encryption for client and inter-node communication; however, this has to be enabled during the build and is not feature-enabled by default.
+[Details](https://redis.io/docs/management/security/encryption/)
+
+Kolla currently does not support TLS encryption.
+
+Note: We had issues with `oslo.cache` connecting via TLS to redis, minor upstream change might be needed.
+
```
artificial-intelligence, line 151: I miss a paragraph about secure libvirt live migration:

https://docs.openstack.org/nova/latest/admin/secure-live-migration-with-qemu-native-tls.html

artificial-intelligence (changes requested):
I'll add more remarks later.

Libvirt TLS is not discussed at all, so I would really at least like to see a paragraph about it.
I'll also need to check if we are missing other important services.

One thing coming to my mind is e.g. nonvnc transport encryption via vencrypt and binding of the communication channel to the novnc proxyserver, as is discussed here:

https://bugs.launchpad.net/kolla-ansible/+bug/2043709

artificial-intelligence:
another service - which is used in SCS - which is not covered here is cloudkitty, mostly used for billing purposes afaik.

see e.g. this commit to enable internal TLS usage, but the cert is not trusted during bootstrap:

https://review.opendev.org/c/openstack/kolla-ansible/+/866598

in general I would advice to go through each openstack service one-by-one and make a detailed review if the service uses any networking and what is transmitted via which network channels and which of these are already encrypted and if not why not.

So I don't know if this was already done and the above mentioned services where just missed - this would bring up the interesting question why they where missed - or if this has just not happened.

I would strongly encourage to do this, if it hasn't happened yet, because otherwise we have no complete picture and the work done here might be incomplete and ultimately lead to a false sense of security by suggesting that everything is encrypted when it really isn't.

I hope this helps, thank you all for all the work put into this so far!

PR #372, opened 2023-11-03
Fix/fix cpuvendor example

garloff:
Also: Add note on Bergamo chips (Zen-4c Siena cores) not getting their own generation number.

### DIFF
--- a/Standards/scs-0100-v3-flavor-naming.md
+++ b/Standards/scs-0100-v3-flavor-naming.md
@@ -407,7 +407,7 @@ out when generating the name for comparison. In other words: 0 has a meaning of
 - SCS-2C-4-10n_bms_**z3**
 - SCS-2C-4-10n_bms_**z3**
 - SCS-2C-4-10n_bms_**z3h**
-- SCS-2C-4-10n_bms_**z3hh** <- Bare Metal, Intel Ice Lake with > 3.25GHz all core freq
+- SCS-2C-4-10n_bms_**z3hh** <- Bare Metal, AMD Milan with > 3.25GHz all core freq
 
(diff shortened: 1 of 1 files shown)


mbuechse (approved):

Standards/scs-0100-v3-flavor-naming.md
```diff
@@ -388,6 +388,12 @@ help the parser tool, which assumes 0 for an unspecified value and does leave it
 out when generating the name for comparison. In other words: 0 has a meaning of
 "rather old or unspecified".
 
+Note: We don't differentiate between Zen-4 (Genoa) and Zen-4c (Bergamo); L3 cache per
```
berendt, line 391: Should we use Docusaurus admonitions here?

https://docusaurus.io/docs/markdown-features/admonitions

:::note

...

:::

garloff: Will probably look nicer in the docusaurus rendering, so I followed your suggestion.

mbuechse: The problem is it doesn't work with the Github rendering. Is this the right tradeoff?

garloff: Fair question.
I started consuming SCS standards primarily via <https://docs.scs.community/standards/standards/overview> and think we should point people there. So the formatting there is very relevant.
But maybe we should strive to use formatting the both docusaurus and github MD renderer can use to produce nice-looking results.

garloff: I see that the examples are not rendered correctly on docusaurus and that the latest merge is not available there yet either. So this will need some more work ....

Issue #371, opened 2023-11-02
[Other] add pluscloud open prod3 to healthmonitor + standards compliance check


frosty-geek:
Hi,

I've just spawned the necessary projects + users for our next productive environment aka. pluscloud open "prod3".

@garloff credentials should be in your inbox @maxwolfs mentioned that the current compliance still uses the github actions (-> I have no access to the secrets in this repo).

Ralf

garloff:
https://health.prod1.plusserver.sovereignit.cloud:3000/d/9ltTEmlnk/openstack-health-monitor2?orgId=1&refresh=5m&var-mycmd=All&var-mymethod=All&var-mywait=All&var-mybench=All&var-mycloud=plus-prod3
has been added.
See https://github.com/SovereignCloudStack/openstack-health-monitor/pull/143.
The compliance check (SCS-compatible IaaS) will be added next ...

mbuechse:
I think it's been dealt with.

PR #373, opened 2023-11-03
Add compliance workflow for PCO Prod3.

garloff:
Using an application credential, because it's more elegant.


garloff:
The second half to address #371 ...

.github/scs-compliance-check/openstack/clouds.yaml
```diff
@@ -29,6 +29,14 @@ clouds:
     region_name: "prod2"
     interface: "public"
     identity_api_version: 3
+  pco-prod3:
+    auth_type: "v3applicationcredential"
+    auth:
+      auth_url: https://prod3.api.pco.get-cloud.io:5000
+      application_credential_id: "fe66c4c8cd3b4ea08262424783e1c58f"
+    region_name: "prod2"
```
mbuechse, line 37: Is that correct?

mbuechse (commented):
One question that should be answered, apart from that, LGTM

.github/scs-compliance-check/openstack/clouds.yaml
garloff: No.
Thanks for spotting.

mbuechse (approved):

PR #374, opened 2023-11-06
Clarify SMs/CUs/EUs belonging to nVidia/AMD/Intel.

garloff:
Also add a line clarifying the order of extensions being fixed.

### DIFF
--- a/Standards/scs-0100-v3-flavor-naming.md
+++ b/Standards/scs-0100-v3-flavor-naming.md
@@ -207,19 +207,19 @@ so users can expect some level of parallelism and independence.
 
 - SCS-2C-4-**10n**
 - SCS-2C-4-**10s**
-- SCS-2C-4-**10s**\_bms_z3
+- SCS-2C-4-**10s**\_bms\_z3
 - SCS-2C-4-**3x10s** <- Cloud creates three 10GB SSDs
-- SCS-2C-4-**3x10s**\_bms_z3
+- SCS-2C-4-**3x10s**\_bms\_z3
 - SCS-2C-4-**10** <- Cloud decides disk type
-- SCS-2C-4-**10**\_bms_z3
-- SCS-2C-4-**n** <- Cloud decides disk size (min_disk from image or larger)
-- SCS-2C-4-**n**\_bms_3
+- SCS-2C-4-**10**\_bms\_z3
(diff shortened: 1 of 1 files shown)


Standards/scs-0100-v3-flavor-naming.md
```diff
@@ -428,16 +431,16 @@ This extension provides more details on the specific GPU:
 - pass-through (`G`) vs. virtual GPU (`g`)
 - vendor (X)
 - generation (N)
-- no. of compute units / SMs / EUs (M) that are exposed (for pass-through) or assigned
+- no. (M) of compute units (CUs) / streaming multiprocessors (SMs) / execution units (EUs) - that are exposed (for pass-through) or assigned
```
mbuechse, line 434: I think I would prefer to use a generic term (i.e., one that is not vendor-specific). Then we can mention that the vendors use other terms for the "same" thing (the thing that we want to distinguish, however "same").

mbuechse (approved):
I approve, but I think it could be improved by using a term that is not vendor specific.

Standards/scs-0100-v3-flavor-naming.md
anjastrunk: I agree, I will ask our GPU expert for such a name.

anjastrunk: Let's choose "processing units".

```suggestion
- no. (M) of processing units that are exposed (for pass-through) or assigned. A processing unit performs computational work. Vendors may have specific names for processing units, such as execution units (EUs) by Intel or "streaming multiprocessors (SMs)" by NVIDIA calls.
```

anjastrunk (approved):

Standards/scs-0100-v3-flavor-naming.md
garloff: As there are three major GPU vendors, shouldn't we have examples for all three in the text to treat them equally?

anjastrunk: I agree and will figure out, how AMD calls processing units.

mbuechse: Anja, I have already incorporated this point. This whole PR has been merged. It's done!

PR #376, opened 2023-11-07
[Standardization] Baseline K8s cluster security (#415)

### LINKED ISSUE #415: [Other] Adapt C&H test cluster to SCS Flavor Naming Standard (scs-0100)
Adapt C&H test cluster to [SCS Flavor Naming Standard (scs-0100)](https://docs-staging.scs.community/standards/scs-0100-v3-flavor-naming) and evaluate its costs. 

Todos:

- [x]  Adapt C&H test cluster to SCS standard 
- [x]  Update standard documentation with costs/cost estimation
- [x]  Write how to and/or adaption script

cah-hbaum:
~This PR adds a standard for high security setups, which should increase the overall of a Kubernetes cluster.~
Edit: After communicating with @joshmue, this is being renamed into "Baseline K8s cluster security", since we only put up a few points which should be important for a starting security setup.

Closes https://github.com/SovereignCloudStack/issues/issues/415

### DIFF
--- a/Standards/scs-0217-v1-baseline-cluster-security.md
+++ b/Standards/scs-0217-v1-baseline-cluster-security.md
@@ -7,10 +7,7 @@ track: KaaS
 
 ## Introduction
 
-A Kubernetes instance is provided as a cluster, which consists of a set of machines,
-so-called nodes. These clusters are configured depending on the needs of its users, which
-is possible due to the flexibility of Kubernetes.
-Due to the regular changes, there are always new security features to deploy and use in Kubernetes.
+Due to the regular changes and updates, there are always new security features to deploy and use in Kubernetes.
 Nevertheless, a provider (or even a customer) needs to take action in order to achieve a
 hardened, secure cluster due to the myriad of configurations possible. This is especially
 the case since Kubernetes ships with insecure features and configurations out of the box,
@@ -18,14 +15,22 @@ which will need to be mitigated by an administrator with the proper knowledge.
 Hardened, secure Kubernetes clusters are desirable regardless of the possible threat model,
 since higher security doesn't necessarily mean higher complexity in this case.
 
+## Terminology
+
+| Term | Meaning                     |
+|------|-----------------------------|
+| TLS  | Transport Layer Security    |
+| CA   | Certificate Authority       |
+| CSR  | Certificate Signing Request |
+
 ## Motivation
 
 Kubernetes clusters are highly configurable, which also gives rise to different security
 problems, if the configuration isn't done properly.
 These security risks can potentially be exposed in many different parts of a cluster, e.g.
 different APIs, authorization and authentication procedures or even Pod privilege mechanisms.
-In order to mitigate these problems, different hardening and prevention steps and mechanisms
-could be used to increase the security of a Kubernetes setup.
+In order to mitigate these problems, different steps and mechanisms could be used to increase
+the security of a Kubernetes setup.
 
 ## Design Considerations
 
@@ -53,7 +58,7 @@ a reference to the CA keypair, which was used in the previous example to sign a 
 
 ### Protected Kubernetes endpoints
 
-In order to secure a Kubernetes cluster, the protection and hardening of endpoints is important.
+In order to secure a Kubernetes cluster, the protection of endpoints is important.
 To do this, different approaches can be taken.
 
 #### TLS for all internal/API traffic
@@ -70,8 +75,9 @@ A list of the default service endpoints can be seen in the following table
 | TCP      | 10250       | Kubelet API             | -                                                                                       |
 | TCP      | 10251/10259 | kube-scheduler          | 10251 could be insecure before 1.13, after that only the secure port 10259 is available |
 | TCP      | 10252/10257 | kube-controller-manager | 10252 could be insecure before 1.13, after that only the secure port 10257 is available |
-| TCP      | 10255       | Read-Only Kubelet API   | HTTP port for GET requests                                                              |
 | TCP      | 30000-32767 | NodePort Services       | Service endpoints, could be HTTP                                                        |
+
+The usage of `readOnlyPort` (enabling a read-only Kubelet API port on 10255) by design neither provides authentication nor authorization. Its usage is strongly discouraged!
 
 #### Authentication and Authorization
 
@@ -83,7 +89,7 @@ x509 certificates, which will normally be created automatically during the setup
 External users can authenticate via an access pattern of choice, which is typically decided by
 the cluster provider.
 
-Authorization is (normally) done by he Role-Based Access Control (RBAC), which matches a request
+Authorization is (normally) done by the Role-Based Access Control (RBAC), which matches a request
 by a user with a set of permissions, also called a role. Kubernetes deploys some roles out-of-the-box;
 additional roles need to be carefully checked, since some permissions for specific resources allow
 modification of other resources.
@@ -107,34 +113,30 @@ After that, the Kubelet calls the `SubjectAccessReview` API in order to determin
 
 ## Decision
 
-This standard tries to increase security for a Kubernetes cluster and harden it in order
-to provide a high security setup. For this to work, multiple measures need to be undertaken.
-It is important to note that this standard is not REQUIRED for all clusters,
-but instead gives best practices for increasing security. Nevertheless, if a cluster is
-provided on the basis of high security, this standard MUST be applied.
+This standard tries to increase security for a Kubernetes cluster in order to provide a
+solid baseline setup with regard to security. For this to work, multiple measures need to be undertaken.
 
 A self-controlled CA SHOULD be used in order to be in control of the TLS certificates, which
-enables the operator to provide and revoke certificates according to the requirements.
+enables operators to provide and revoke certificates according to their own requirements.
 
-All internal endpoints found in the section [TLS for all internal/API traffic] should be
+All internal endpoints found in the section [TLS for all internal/API traffic] MUST be
 encrypted with TLS in order to secure internal traffic.
 
 The Kubernetes API (kubeAPI) MUST be secured by authenticating and authorizing the users
 trying to access its endpoints. How a user is authenticated is up to the provider of the
-cluster and/or the wishes of the customer. Authorization MUST be done by providing
-fine-grained RBAC.
-This whole process also MUST be applied to the Kubelet, which in its default configuration
-doesn't enable authentication or authorization. A way to do this can be found in the
-section [Authentication and Authorization].
+cluster and/or the wishes of the customer. Authorization MUST be done by providing fine-grained RBAC.
+The authentication and authorization steps MUST also be applied to the Kubelet, which in its default configuration
+doesn't enable them. A way to do this can be found in the section [Authentication and Authorization].
 
 ## Related Documents
 
-[Managing TLS in a cluster](https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/)
-[Securing a cluster](https://kubernetes.io/docs/tasks/administer-cluster/securing-a-cluster/)
-[Controlling access](https://kubernetes.io/docs/concepts/security/controlling-access/)
-[Kubelet Authentication and Authorization](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-authn-authz/)
-[Authentication](https://kubernetes.io/docs/reference/access-authn-authz/authentication/)
-[Kubernetes Security](https://cheatsheetseries.owasp.org/cheatsheets/Kubernetes_Security_Cheat_Sheet.html)
+- [Managing TLS in a cluster](https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/)
+- [Securing a cluster](https://kubernetes.io/docs/tasks/administer-cluster/securing-a-cluster/)
+- [Controlling access](https://kubernetes.io/docs/concepts/security/controlling-access/)
+- [Kubernetes Security Checklist](https://kubernetes.io/docs/concepts/security/security-checklist/)
+- [Kubelet Authentication and Authorization](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-authn-authz/)
+- [Authentication](https://kubernetes.io/docs/reference/access-authn-authz/authentication/)
+- [OWASP Kubernetes Security Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Kubernetes_Security_Cheat_Sheet.html)
 
 ## Conformance Tests
 
--- a/Standards/scs-0214-v1-k8s-node-distribution.md
+++ b/Standards/scs-0214-v1-k8s-node-distribution.md
@@ -8,7 +8,7 @@ track: KaaS
 ## Introduction
 
 A Kubernetes instance is provided as a cluster, which consists of a set of machines,
-so called nodes. A cluster is composed of a control plane and at least one worker node.
+so-called nodes. A cluster is composed of a control plane and at least one worker node.
 The control plane manages the worker nodes and therefore the pods in the cluster by making
 decisions about scheduling, event detection and rights management. Inside the control plane,
 multiple components exist, which can be duplicated and distributed over multiple nodes
@@ -19,11 +19,11 @@ separate the controller component from user workloads, which could pose a securi
 
 The following terms are used throughout this document:
 
-| Term | Meaning |
-|---|---|
-| Worker | Virtual or bare-metal machine, which hosts workloads of customers |
+| Term          | Meaning                                                                                                                                                                     |
+|---------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Worker        | Virtual or bare-metal machine, which hosts workloads of customers                                                                                                           |
 | Control Plane | Virtual or bare-metal machine, which hosts the container orchestration layer that exposes the API and interfaces to define, deploy, and manage the lifecycle of containers. |
-| Machine | Virtual or bare-metal entity with computational capabilities |
+| Machine       | Virtual or bare-metal entity with computational capabilities                                                                                                                |
 
 ## Motivation
 
--- a/Standards/scs-0216-v1-cluster-baseline-security.md
+++ b/Standards/scs-0216-v1-cluster-baseline-security.md
@@ -15,7 +15,7 @@ Nevertheless, a provider (or even a customer) needs to take action in order to a
 hardened, secure cluster due to the myriad of configurations possible. This is especially
 the case since Kubernetes ships with insecure features and configurations out of the box,
 which will need to be mitigated by an administrator with the proper knowledge.
-Hardened Kubernetes clusters are desirable regardless of the possible thread model, since
+Hardened Kubernetes clusters are desirable regardless of the possible threat model, since
 higher security doesn't necessarily mean higher complexity in this case.
 
 ## Motivation
@@ -60,7 +60,7 @@ To do this, different approaches can be taken.
 
 #### TLS for all internal/API traffic
 
-It is already expected by Kubernetes, that all API communication internally is encrypted with TLS.
+It is already expected by Kubernetes that all API communication internally is encrypted with TLS.
 Nevertheless, some endpoints of internal components could be/will be exposed without the necessary
 encryption, which could lead to weak points in the system.
 A list of the default service endpoints can be seen in the following table
@@ -92,7 +92,7 @@ the cluster provider.
 
 Authorization is (normally) done by he Role-Based Access Control (RBAC), which matches a request
 by a user with a set of permissions, also called a role. Kubernetes deploys some roles out-of-the-box;
-additional roles needs to be carefully checked, since some permissions for specific resources allow
+additional roles need to be carefully checked, since some permissions for specific resources allow
 modification of other resources.
 
 This whole process is especially important for the Kubelet, which allows anonymous requests in its
--- a/Standards/scs-0216-v1-cluster-baseline-security.md
+++ b/Standards/scs-0216-v1-cluster-baseline-security.md
@@ -8,22 +8,20 @@ track: KaaS
 ## Introduction
 
 A Kubernetes instance is provided as a cluster, which consists of a set of machines,
-so called nodes. These clusters are configured depending on the needs of its users, which
+so-called nodes. These clusters are configured depending on the needs of its users, which
 is possible due to the flexibility of Kubernetes.
 Due to the regular changes, there are always new security features to deploy and use in Kubernetes.
 Nevertheless, a provider (or even a customer) needs to take action in order to achieve a
 hardened, secure cluster due to the myriad of configurations possible. This is especially
 the case since Kubernetes ships with insecure features and configurations out of the box,
 which will need to be mitigated by an administrator with the proper knowledge.
-Hardened Kubernetes clusters are desirable regardless of the possible threat model, since
-higher security doesn't necessarily mean higher complexity in this case.
+Hardened, secure Kubernetes clusters are desirable regardless of the possible threat model,
+since higher security doesn't necessarily mean higher complexity in this case.
 
 ## Motivation
 
-Kubernetes clusters are highly configurable to fit a wide array of use-cases and setups.
-This makes them flexible and scalable to a certain extent, which is probably also why they
-found such a wide adoption. But its also obvious, that security problems can arise if a
-cluster isn't configured properly to mitigate them.
+Kubernetes clusters are highly configurable, which also gives rise to different security
+problems, if the configuration isn't done properly.
 These security risks can potentially be exposed in many different parts of a cluster, e.g.
 different APIs, authorization and authentication procedures or even Pod privilege mechanisms.
 In order to mitigate these problems, different hardening and prevention steps and mechanisms
@@ -65,20 +63,15 @@ Nevertheless, some endpoints of internal components could be/will be exposed wit
 encryption, which could lead to weak points in the system.
 A list of the default service endpoints can be seen in the following table
 
-Protocol | Port Range | Purpose
-— | — | —
-TCP | 6443- | Kubernetes API Server
-TCP | 2379-2380 | etcd server client API
-TCP | 10250 | Kubelet API
-TCP | 10251 | kube-scheduler
-TCP | 10252 | kube-controller-manager
-TCP | 10255 | Read-Only Kubelet API
-
-Protocol | Port Range | Purpose
-— | — | —
-TCP | 10250 | Kubelet API
-TCP | 10255 | Read-Only Kubelet API
-TCP | 30000-32767 | NodePort Services
+| Protocol | Port Range  | Purpose                 | Notes                                                                                   |
+|----------|-------------|-------------------------|-----------------------------------------------------------------------------------------|
+| TCP      | 6443*       | Kubernetes API Server   | -                                                                                       |
+| TCP      | 2379-2380   | etcd server client API  | -                                                                                       |
+| TCP      | 10250       | Kubelet API             | -                                                                                       |
+| TCP      | 10251/10259 | kube-scheduler          | 10251 could be insecure before 1.13, after that only the secure port 10259 is available |
+| TCP      | 10252/10257 | kube-controller-manager | 10252 could be insecure before 1.13, after that only the secure port 10257 is available |
+| TCP      | 10255       | Read-Only Kubelet API   | HTTP port for GET requests                                                              |
+| TCP      | 30000-32767 | NodePort Services       | Service endpoints, could be HTTP                                                        |
 
 #### Authentication and Authorization
 
@@ -101,7 +94,7 @@ could manipulate resources that are managed with the Kubelet.
 
 To disable anonymous requests, the Kubelet should be started with `--anonymous-auth=false`.
 Authentication can be provided either through x509 client certificates or API bearer tokens.
-How to setup both approaches can be found in the [Kubelet Authentication and Authorization](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-authn-authz/).
+How to set up both approaches can be found in the [Kubelet Authentication and Authorization](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-authn-authz/).
 
 Kubelet authorization is set to `AlwaysAllow` as a default mode. This can be quite problematic,
 since all authenticated users can do all actions. To mitigate this, it is possible to delegate
--- a/Standards/scs-0216-v1-cluster-baseline-security.md
+++ b/Standards/scs-0216-v1-cluster-baseline-security.md
@@ -1,0 +1,148 @@
+---
+title: Kubernetes cluster baseline security
+type: Standard
+status: Draft
+track: KaaS
+---
+
+## Introduction
+
+A Kubernetes instance is provided as a cluster, which consists of a set of machines,
+so called nodes. These clusters are configured depending on the needs of its users, which
+is possible due to the flexibility of Kubernetes.
+Due to the regular changes, there are always new security features to deploy and use in Kubernetes.
+Nevertheless, a provider (or even a customer) needs to take action in order to achieve a
+hardened, secure cluster due to the myriad of configurations possible. This is especially
+the case since Kubernetes ships with insecure features and configurations out of the box,
+which will need to be mitigated by an administrator with the proper knowledge.
+Hardened Kubernetes clusters are desirable regardless of the possible thread model, since
+higher security doesn't necessarily mean higher complexity in this case.
+
+## Motivation
+
+Kubernetes clusters are highly configurable to fit a wide array of use-cases and setups.
+This makes them flexible and scalable to a certain extent, which is probably also why they
+found such a wide adoption. But its also obvious, that security problems can arise if a
+cluster isn't configured properly to mitigate them.
+These security risks can potentially be exposed in many different parts of a cluster, e.g.
+different APIs, authorization and authentication procedures or even Pod privilege mechanisms.
+In order to mitigate these problems, different hardening and prevention steps and mechanisms
+could be used to increase the security of a Kubernetes setup.
+
+## Design Considerations
+
+### External CA
+
+Kubernetes provides an API to provision TLS certificates that can be signed by a CA.
+This CA can be controlled by the cluster provider, which enables much more tight control
+over the clusters communication and therefore also better controllable security.
+
+In order to do this, the CA certificate bundle needs to be added to the trusted certificates
+of the server.
+To provide a certificate, the following steps need to be undertaken:
+
+1. Create a CSR
+2. Send the CSR manifest to the k8s API
+3. Approve the CSR
+4. Sign CSR with your CA
+5. Upload the signed certificate to the server
+
+This certificate could now be used by a user in a pod in order to provide a trusted certificate.
+
+It is also possible for the Kubernetes controller manager to provide the signing functionality.
+To enable this, `--cluster-signing-cert-file` and `--cluster-signing-key-file` need to be set with
+a reference to the CA keypair, which was used in the previous example to sign a CSR.
+
+### Protected Kubernetes endpoints
+
+In order to secure a Kubernetes cluster, the protection and hardening of endpoints is important.
+To do this, different approaches can be taken.
+
+#### TLS for all internal/API traffic
+
+It is already expected by Kubernetes, that all API communication internally is encrypted with TLS.
+Nevertheless, some endpoints of internal components could be/will be exposed without the necessary
+encryption, which could lead to weak points in the system.
+A list of the default service endpoints can be seen in the following table
+
+Protocol | Port Range | Purpose
+— | — | —
+TCP | 6443- | Kubernetes API Server
+TCP | 2379-2380 | etcd server client API
+TCP | 10250 | Kubelet API
+TCP | 10251 | kube-scheduler
+TCP | 10252 | kube-controller-manager
+TCP | 10255 | Read-Only Kubelet API
+
+Protocol | Port Range | Purpose
+— | — | —
+TCP | 10250 | Kubelet API
+TCP | 10255 | Read-Only Kubelet API
+TCP | 30000-32767 | NodePort Services
+
+#### Authentication and Authorization
+
+All API clients should authenticate and authorize in order to be able to access an API or even
+specific functions of this API. This is the case for users as well as internal components.
+
+Most internal clients (like proxies or nodes) are typically authenticated via service accounts or
+x509 certificates, which will normally be created automatically during the setup of a cluster.
+External users can authenticate via an access pattern of choice, which is typically decided by
+the cluster provider.
+
+Authorization is (normally) done by he Role-Based Access Control (RBAC), which matches a request
+by a user with a set of permissions, also called a role. Kubernetes deploys some roles out-of-the-box;
+additional roles needs to be carefully checked, since some permissions for specific resources allow
+modification of other resources.
+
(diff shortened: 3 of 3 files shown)


cah-hbaum:
This standard is written for now but probably needs a rework.
I needed to read a bit about Kubernetes security and hardening. IMO, it all makes sense, but the standard itself isn't really nice to read through, since it only says what to do very general. If someone needs more details for this, they need to read the `Design considerations` section or go more in-depth in the K8s documents.

Please make some improvements to the text or some proposals about additional hardening steps if they seem useful to you.

What I also found problematic is that this standard is one of those that can't be made MANDATORY (so I couldn't write MUST for the entire standard), since not all clusters need to be "secure" or it's just not possible for specific setups.
I think this makes more sense for clusters that are sold as "High Security", which should adhere to this standard.

Standards/scs-0216-v1-high-security-setup.md
```diff
@@ -0,0 +1,174 @@
...
+
+A Kubernetes instance is provided as a cluster, which consists of a set of machines,
+so called nodes. These clusters are configured depending on the needs of its users, which
+is possible due to the flexibility of Kubernetes.
+Due to the regular changes, there are always new security features to deploy and use in Kubernetes.
+Nevertheless, a provider (or even a customer) needs to take action in order to achieve a
+hardened, secure cluster due to the myriad of configurations possible. This is especially
+the case since Kubernetes ships with insecure features and configurations out of the box,
+which will need to be mitigated by an administrator with the proper knowledge.
+Hardened Kubernetes clusters are desirable regardless of the possible thread model, since
```
garloff, line 18:
```suggest
Hardened Kubernetes clusters are desirable regardless of the possible threat model, since
```

Standards/scs-0216-v1-high-security-setup.md
```diff
@@ -0,0 +1,174 @@
...
+## Motivation
+
+Kubernetes clusters are highly configurable to fit a wide array of use-cases and setups.
+This makes them flexible and scalable to a certain extent, which is probably also why they
+found such a wide adoption. But its also obvious, that security problems can arise if a
+cluster isn't configured properly to mitigate them.
+These security risks can potentially be exposed in many different parts of a cluster, e.g.
+different APIs, authorization and authentication procedures or even Pod privilege mechanisms.
+In order to mitigate these problems, different hardening and prevention steps and mechanisms
+could be used to increase the security of a Kubernetes setup.
```
garloff, line 30: The Motivation section duplicates the introduction to a certain degree, which means we have an opportunity to make the text shorter. We should use such opportunities ...

Standards/scs-0216-v1-high-security-setup.md
```diff
@@ -0,0 +1,174 @@
...
+TCP | 10251 | kube-scheduler
+TCP | 10252 | kube-controller-manager
+TCP | 10255 | Read-Only Kubelet API
+
+Protocol | Port Range | Purpose
+— | — | —
+TCP | 10250 | Kubelet API
+TCP | 10255 | Read-Only Kubelet API
+TCP | 30000-32767 | NodePort Services
+
```
garloff, line 82: Sidenote: The tables do not seem to render in github's MD preview.

Content: You talk about internal components being exposed without encryption and then you provide a table with all edpoints AFAICS. As a reader, I would expect you to point out which of these are unencrypted and this potentially a risk for security.

Standards/scs-0216-v1-high-security-setup.md
```diff
@@ -0,0 +1,174 @@
...
+specific functions of this API. This is the case for users as well as internal components.
+
+Most internal clients (like proxies or nodes) are typically authenticated via service accounts or
+x509 certificates, which will normally be created automatically during the setup of a cluster.
+External users can authenticate via an access pattern of choice, which is typically decided by
+the cluster provider.
+
+Authorization is (normally) done by he Role-Based Access Control (RBAC), which matches a request
+by a user with a set of permissions, also called a role. Kubernetes deploys some roles out-of-the-box;
+additional roles needs to be carefully checked, since some permissions for specific resources allow
```
garloff, line 95:
```suggestion
dditional roles need to be carefully checked, since some permissions for specific resources allow
```

Standards/scs-0216-v1-high-security-setup.md
```diff
@@ -0,0 +1,174 @@
...
+a reference to the CA keypair, which was used in the previous example to sign a CSR.
+
+### Protected Kubernetes endpoints
+
+In order to secure a Kubernetes cluster, the protection and hardening of endpoints is important.
+To do this, different approaches can be taken.
+
+#### TLS for all internal/API traffic
+
+It is already expected by Kubernetes, that all API communication internally is encrypted with TLS.
```
garloff, line 63:
```suggestion
It is already expected by Kubernetes that all API communication internally is encrypted with TLS.
```
(no comma to delimit relative clauses in English)

garloff (commented):
Nice work, Hannes!
What we need to do is to start having a discussion with our container team:
* Do we find consensus on what we recommend and what we require?
* Can we implement all of this in Cluster Stacks? And could we at least implement this in good old k8s-cluster-api-provider? (We need to decide how much of this we still want to add to old v1 KaaS which has only a limited life time left --things that are risky, very tedious for us or have the effect of breaking users or requiring significant changes on their side are probably not good for our old solution.) 

@jschoone - can you take it there?

Standards/scs-0216-v1-high-security-setup.md
cah-hbaum: Updated the `Motivation` section a little bit and shortened the text.

Standards/scs-0216-v1-high-security-setup.md
cah-hbaum: As far as I understood from the sources I read, all of these endpoints COULD potentially be unencrypted, if the wrong configuration is done.
But I will add another column to point to some especially endagered endpoints.

Standards/scs-0216-v1-high-security-setup.md
```diff
@@ -0,0 +1,166 @@
...
+A Kubernetes instance is provided as a cluster, which consists of a set of machines,
+so-called nodes. These clusters are configured depending on the needs of its users, which
+is possible due to the flexibility of Kubernetes.
+Due to the regular changes, there are always new security features to deploy and use in Kubernetes.
+Nevertheless, a provider (or even a customer) needs to take action in order to achieve a
+hardened, secure cluster due to the myriad of configurations possible. This is especially
+the case since Kubernetes ships with insecure features and configurations out of the box,
+which will need to be mitigated by an administrator with the proper knowledge.
+Hardened, secure Kubernetes clusters are desirable regardless of the possible threat model,
+since higher security doesn't necessarily mean higher complexity in this case.
```
joshmue, line 19: As far as recommendations in this document go, I strongly agree with that.

However, again, as far as recommendations in this document go, I'd probably call it rather a security baseline standard, while referencing primarily to CIS benchmarking.

Going further, a "High Security Kubernetes Cluster" would probably include points like...
- network protection of the API server using IP allow lists or using bastion hosts
- support for CRI runtimes supporting VM-based runtime isolation
- disabling generation of admin client certificates handed to users
- ...

Standards/scs-0216-v1-high-security-setup.md
```diff
@@ -0,0 +1,166 @@
...
+How to set up both approaches can be found in the [Kubelet Authentication and Authorization](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-authn-authz/).
+
+Kubelet authorization is set to `AlwaysAllow` as a default mode. This can be quite problematic,
+since all authenticated users can do all actions. To mitigate this, it is possible to delegate
+authorization to the API server by:
+
+- enabling the `authorization.k8s.io/v1beta1` API group
+- starting the Kubelet with the `--authorization-mode=Webhook` and the `--kubeconfig` flags
+
+After that, the Kubelet calls the `SubjectAccessReview` API in order to determine the authorization of a request.
```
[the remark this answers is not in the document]
cah-hbaum: Do you have a specific document of them in mind (TBF I'm not willing to give them my mail, even if its just my work mail). Then I can/will rewrite the text to mention these standards.

Standards/scs-0216-v1-high-security-setup.md
cah-hbaum: Since I only wrote this standard with the points given by the issue and I'm not that knowledgeable in this specific field (and also not with the points you bring up), I would probably rename the standard to "K8s Security Baseline" and put the points you mentioned in a new standard.

Also just to ask: Do we trust CIS as a standard enough to put it as a reference to such a big topic? They probably do good work, but we also would need to review their standards (versions) constantly to ensure consistency with our requirements.

joshmue: > Since I only wrote this standard with the points given by the issue and I'm not that knowledgeable in this specific field (and also not with the points you bring up), I would probably rename the standard to "K8s Security Baseline" and put the points you mentioned in a new standard.

:+1: 

[Your comment from above:](https://github.com/SovereignCloudStack/standards/pull/376#discussion_r1403370342)
> 
> Do you have a specific document of them in mind (TBF I'm not willing to give them my mail, even if its just my work mail). Then I can/will rewrite the text to mention these standards.

I found the access to the PDF to be gated through a registration to be odd, as well (https://www.cisecurity.org/benchmark/kubernetes). Also did not complete the form due to concerns that I might violate their TOS by getting their PDF to use it kind-of commercially without license.
Definitely, it appears to be an de-facto industry standard; [Azure](https://learn.microsoft.com/en-us/azure/aks/cis-kubernetes), [Google](https://cloud.google.com/kubernetes-engine/docs/concepts/cis-benchmarks#status), [VMware](https://docs.vmware.com/en/VMware-Tanzu-Mission-Control/services/tanzumc-concepts/GUID-1BB98612-A9BF-494C-8446-1DB2E80BF5F9.html) and probably others use it to benchmark their offerings. Of course, they also may have a business license.

While concerns around this topic are IMHO valid, I concluded that they are negligible, as AFAIK kube-bench is Apache-licensed and we AFAIK can read it with any (non-)commercial intent. That is, of course, given kube-bench maintainers had the rights to publish it like that to begin with :D

> Also just to ask: Do we trust CIS as a standard enough to put it as a reference to such a big topic? They probably do good work, but we also would need to review their standards (versions) constantly to ensure consistency with our requirements.

That is a good question indeed. Especially as the CIS benchmarks do not seem to be really free/open that it would be a great match with SCS.

But I do not think that there are many viable alternatives. E. g. if SCS-employed/contracted staff was tasked with writing up a similar security benchmark, they would...

- ...need a fair amount of time
- ...miss quite a few points OR mostly copy/paste from CIS benchmarks

None of these outcomes would prevent future efforts to keep them up-to-date OR make the benchmark both high-quality and completely free/open.

cah-hbaum: Makes a lot sense to me, thank you for the detailed explanation.
I think a good way to (at least) get on top of these standards would be to reference a specific version in one of our standards. If we agree with this version (if they version their standards, which they hopefully do), we would have a baseline, that we could later expand on if we find the time to read a newer version of it and agree on its contents.
I'm still not happy with the gate process and we should probably look into the license problem before we continue down this path, but at least that sounds good to me.

cah-hbaum: Did a bit more reading up. The CIS benchmarks themselves are available under a [CC BY-NC-SA 4.0 Deed ](https://creativecommons.org/licenses/by-nc-sa/4.0/deed.en) license, meaning that we could use it as long as it is for non-commercial use (which I think is the case).

The Gating process seems to be in place in order to enforce their ToS, which mainly say that they're not responsible for anything done with the help of/under their tests (so they're not liable).

While I agree with this, I'm still unhappy with the gating process.
For this reason and because I won't wait for a decision on this (and maybe someone else has more insight into this to write something useful), I would drop the parts mentioned in https://github.com/SovereignCloudStack/standards/pull/376#discussion_r1397322690 and publish the document without them. We can then create a Version 2 of this standard, which would include the CIS benchmark mentions.

I think thats the best solution for now, otherwise we would sit endlessly on this issue.

Standards/scs-0216-v1-high-security-setup.md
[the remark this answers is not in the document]
cah-hbaum: Took a decision for now at https://github.com/SovereignCloudStack/standards/pull/376#discussion_r1419013781

Standards/scs-0216-v1-high-security-setup.md
joshmue: > I would keep the parts mentioned in #376 (comment), since they still provide a good foundation, and publish the document in its current state. We can then create a Version 2 of this standard, which would include the CIS benchmark mentions and maybe drop the parts mentioned.

Alright!

> For this reason and because I won't wait for a decision on this

Going forward (v2 and beyond), it probably is a good idea to identify someone with a fitting role to make such decision. I guess someone from the core SCS project team.

Standards/scs-0216-v1-baseline-cluster-security.md
```diff
@@ -0,0 +1,141 @@
...
+
+After that, the Kubelet calls the `SubjectAccessReview` API in order to determine the authorization of a request.
+
+## Decision
+
+This standard tries to increase security for a Kubernetes cluster and harden it in order
+to provide a high security setup. For this to work, multiple measures need to be undertaken.
+It is important to note that this standard is not REQUIRED for all clusters,
+but instead gives best practices for increasing security. Nevertheless, if a cluster is
+provided on the basis of high security, this standard MUST be applied.
```
joshmue, line 114: This paragraph still contains the "high security setup" term instead of "baseline security" or similar.

(The `Introduction` and `Motivation` sections also contain the "hardening" term, which might be a little bit overselling as well)

joshmue (approved):
Given a perspective to ...
- have a decision for including OR discarding of CSI benchmarks in some v2 version
- have some "high security setup" standard at a later stage

([discussion](https://github.com/SovereignCloudStack/standards/pull/376#discussion_r1397363873))

... this LGTM generally.

Standards/scs-0216-v1-baseline-cluster-security.md
cah-hbaum: Removed `hardening` and `high security` from the mentioned parts (and the rest of the document) and especially rewrote the commented section a bit more.

garloff:
Notes on our reference implementation SCS KaaS v1 ([k8s-cluster-api-provider](https://github.com/SovereignCloudStack/k8s-cluster-api-provider)) which uses capi with kubeadm to configure the nodes with the kubelets:
* We are *not* using an external CA at this point (although the needed logic is available due to the work for supporting a custom CA #372)
* Control nodes and worker nodes are setup by kubeadm and have their kubelet config in `/var/lib/kubelet/config.yaml`.
  What I find there is:
  ```yaml
apiVersion: kubelet.config.k8s.io/v1beta1
authentication:
  anonymous:
    enabled: false
  webhook:
    cacheTTL: 0s
    enabled: true
  x509:
    clientCAFile: /etc/kubernetes/pki/ca.crt
authorization:
  mode: Webhook
  webhook:
    cacheAuthorizedTTL: 0s
    cacheUnauthorizedTTL: 0s
[...]
```
   If I am not misparsing this, we fulfill both the requirement to disable anonymous auth and to enable Webhook authorization.
* Trying to assess the TLS enforcement of any open network ports:
  - Node ports for services are service dependent; I guess we can only be responsible for services that are rolled out with the cluster itself, not the ones the user adds.
  - capo sets up security groups for us. We only allow what the CNI needs (TCP:179, proto 4 (IP-in-IP) for calico, UDP:8472, TCP:4240 and TCP:4244 for Cilium), Kubelet (TCP:10250). For the workers additionally the nodeports (30k-30767) and for the control-plane the kube-api (6443) and etcd (2379).
  - I have not studied how to get data from etcd or cilium hubble or cilium health. I can tell that kube-api and kubelet reject plain http.

File: Standards/scs-0217-v1-baseline-cluster-security.md [hunk 9]
```diff
@@ -0,0 +1,141 @@
...
+
+Kubernetes provides an API to provision TLS certificates that can be signed by a CA.
+This CA can be controlled by the cluster provider, which enables much more tight control
+over the clusters communication and therefore also better controllable security.
+
+In order to do this, the CA certificate bundle needs to be added to the trusted certificates
+of the server.
+To provide a certificate, the following steps need to be undertaken:
+
+1. Create a CSR
```
anjastrunk, line 42: abrivation "CSR" is not introduced.

```suggestion+
1. Create a CSR (Certificate Signing Request)
```

Standards/scs-0217-v1-baseline-cluster-security.md
```diff
@@ -0,0 +1,141 @@
...
+
+All API clients should authenticate and authorize in order to be able to access an API or even
+specific functions of this API. This is the case for users as well as internal components.
+
+Most internal clients (like proxies or nodes) are typically authenticated via service accounts or
+x509 certificates, which will normally be created automatically during the setup of a cluster.
+External users can authenticate via an access pattern of choice, which is typically decided by
+the cluster provider.
+
+Authorization is (normally) done by he Role-Based Access Control (RBAC), which matches a request
```
anjastrunk, line 86: Typo?

```suggestion
Authorization is (normally) done by Role-Based Access Control (RBAC), which matches a request
```

Standards/scs-0217-v1-baseline-cluster-security.md
```diff
@@ -0,0 +1,141 @@
...
+enables the operator to provide and revoke certificates according to the requirements.
+
+All internal endpoints found in the section [TLS for all internal/API traffic] should be
+encrypted with TLS in order to secure internal traffic.
+
+The Kubernetes API (kubeAPI) MUST be secured by authenticating and authorizing the users
+trying to access its endpoints. How a user is authenticated is up to the provider of the
+cluster and/or the wishes of the customer. Authorization MUST be done by providing
+fine-grained RBAC.
+This whole process also MUST be applied to the Kubelet, which in its default configuration
```
anjastrunk, line 126: > This whole process 
Which? The process of authenticating and authorizing?

anjastrunk (approved):
LGTM, except of three minor textual improvements.

Standards/scs-0217-v1-baseline-cluster-security.md
cah-hbaum: Added a `Terminology` section in https://github.com/SovereignCloudStack/standards/pull/376/commits/5610fe15c9cf2ad9ddf1ddf89f0d575a9002b5d2

Standards/scs-0217-v1-baseline-cluster-security.md
cah-hbaum: Fixed in https://github.com/SovereignCloudStack/standards/pull/376/commits/5610fe15c9cf2ad9ddf1ddf89f0d575a9002b5d2

Standards/scs-0217-v1-baseline-cluster-security.md
cah-hbaum: Edited the sentence in https://github.com/SovereignCloudStack/standards/pull/376/commits/5610fe15c9cf2ad9ddf1ddf89f0d575a9002b5d2

Standards/scs-0216-v1-high-security-setup.md
bitkeks: Thank you both for your work! 

The CIS benchmark covers a lot of security configuration options, for every component of Kubernetes. This would definitely be too much for a SCS standard document. Referring to it is the way to go. 

This also keeps the standard up-to-date when the K8s version changes and security needs to be adapted. New releases of the CIS benchmark then covers it automatically.

Standards/scs-0216-v1-high-security-setup.md
cah-hbaum: Resolving this, since its open for over 2 months now.

Standards/scs-0216-v1-baseline-cluster-security.md
cah-hbaum: Resolving this, since @joshmue gave his approval down below.

bitkeks (changes requested):

artificial-intelligence:
> What I also found problematic is that this standard is one of those that can't be made MANDATORY (so I couldn't write MUST for the entire standard), since not all clusters need to be "secure" or it's just not possible for specific setups. I think this makes more sense for clusters that are sold as "High Security", which should adhere to this standard.

I think I disagree with this.

First, could you specify which clusters don't need to be secure? I can't think of any.

Also I don't find much in the standard that is impossible to implement for specific setups?

I also don't think anything is mentioned which is only "high security"?

bitkeks:
As far as I read from the discussion, the "high security" proposal was replaced by now being "baseline security".
That said, as discussed in the meeting, I'd propose to make this standard mandatory, since it's the base line of security in K8s clusters. As a middle ground we could also discuss making this standard mandatory, but use it as a "container" for best practices and guidelines, so that CSPs acknowledge their existence when promoting their SCS-confirming k8s offerings.

Standards/scs-0217-v1-baseline-cluster-security.md
```diff
@@ -0,0 +1,148 @@
...
+
+| Protocol | Port Range  | Purpose                 | Notes                                                                                   |
+|----------|-------------|-------------------------|-----------------------------------------------------------------------------------------|
+| TCP      | 6443*       | Kubernetes API Server   | -                                                                                       |
+| TCP      | 2379-2380   | etcd server client API  | -                                                                                       |
+| TCP      | 10250       | Kubelet API             | -                                                                                       |
+| TCP      | 10251/10259 | kube-scheduler          | 10251 could be insecure before 1.13, after that only the secure port 10259 is available |
+| TCP      | 10252/10257 | kube-controller-manager | 10252 could be insecure before 1.13, after that only the secure port 10257 is available |
+| TCP      | 10255       | Read-Only Kubelet API   | HTTP port for GET requests                                                              |
```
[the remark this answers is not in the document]
cah-hbaum: Mhh I don't really know. As far as my understanding goes, this port is used to request metrics from the kubelet.
So if the port would be disabled, there could be problems when retrieving metric data. On the other hand, it should be possible to use the normal secured endpoint to retrieve these metrics.

Personally, I would declare it as insecure and advise against using it. Since it is normally set to `0`, I think most people won't even set it.

Standards/scs-0217-v1-baseline-cluster-security.md
```diff
@@ -0,0 +1,148 @@
...
+doesn't enable them. A way to do this can be found in the section [Authentication and Authorization].
+
+## Related Documents
+
+[Managing TLS in a cluster](https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/)
+[Securing a cluster](https://kubernetes.io/docs/tasks/administer-cluster/securing-a-cluster/)
+[Controlling access](https://kubernetes.io/docs/concepts/security/controlling-access/)
+[Kubelet Authentication and Authorization](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-authn-authz/)
+[Authentication](https://kubernetes.io/docs/reference/access-authn-authz/authentication/)
+[Kubernetes Security](https://cheatsheetseries.owasp.org/cheatsheets/Kubernetes_Security_Cheat_Sheet.html)
```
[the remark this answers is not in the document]
cah-hbaum: No this should just be a source list.

Standards/scs-0217-v1-baseline-cluster-security.md
```diff
@@ -0,0 +1,148 @@
...
+
+In order to secure a Kubernetes cluster, the protection of endpoints is important.
+To do this, different approaches can be taken.
+
+#### TLS for all internal/API traffic
+
+It is already expected by Kubernetes that all API communication internally is encrypted with TLS.
+Nevertheless, some endpoints of internal components could be/will be exposed without the necessary
+encryption, which could lead to weak points in the system.
+A list of the default service endpoints can be seen in the following table
```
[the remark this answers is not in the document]
cah-hbaum: I found multiple tables hard to handle and not really good to look at. I also didn't see a real need to split everything up, since the name of the service/server type makes it obvious what is meant.

But we can discuss this further if we want.

cah-hbaum:
> > What I also found problematic is that this standard is one of those that can't be made MANDATORY (so I couldn't write MUST for the entire standard), since not all clusters need to be "secure" or it's just not possible for specific setups. I think this makes more sense for clusters that are sold as "High Security", which should adhere to this standard.
> 
> I think I disagree with this.
> 
> First, could you specify which clusters don't need to be secure? I can't think of any.
> 
> Also I don't find much in the standard that is impossible to implement for specific setups?
> 
> I also don't think anything is mentioned which is only "high security"?

Yeah, that was written when the standard was still about high security. I think after renaming it to "Baseline security", we can safely assume that all features standardized here are very much doable and should be written as a `MUST`. I will adapt the standard to mirror this.

> As far as I read from the discussion, the "high security" proposal was replaced by now being "baseline security".
> That said, as discussed in the meeting, I'd propose to make this standard mandatory, since it's the base line of security in > K8s clusters.
> 
> As a middle ground we could also discuss making this standard mandatory, but use it as a "container" for best practices > and guidelines, so that CSPs acknowledge their existence when promoting their SCS-confirming k8s offerings. That
> would include sites like the k8s security guides, checklists, OWASP and CIS benchmarks.
>
> It's probably also a good idea to bring this to the SCS security team, as the implementation of security practices in the
> infrastructure might be done by tools provided by SCS (vuln scanners, traffic filtering, authentication etc.).

As mentioned above in response to the other comment, making it mandatory seems like a good idea, which will be done with the next commit.
I would also encourage bringing this into the SCS security team as long as this doesn't create a hold up again for another 2 months.

Standards/scs-0217-v1-baseline-cluster-security.md
```diff
@@ -0,0 +1,145 @@
+---
+title: Kubernetes cluster baseline security
+type: Standard
+status: Draft
+track: KaaS
+---
+
+## Introduction
+
+A Kubernetes instance is provided as a cluster, which consists of a set of machines,
```
artificial-intelligence, line 10: I'm not sure this first sentence is needed. I assume the target audience of this standard is people at least superficially familiar with kubernetes? It might be good to instead mention the target audience in the standard, but afaik we do not do this in our standards.

Nevertheless, I guess this sentence is redundant information for the target audience which doesn't really add anything to the standard.

Standards/scs-0217-v1-baseline-cluster-security.md
```diff
@@ -0,0 +1,145 @@
...
+a reference to the CA keypair, which was used in the previous example to sign a CSR.
+
+### Protected Kubernetes endpoints
+
+In order to secure a Kubernetes cluster, the protection of endpoints is important.
+To do this, different approaches can be taken.
+
+#### TLS for all internal/API traffic
+
+It is already expected by Kubernetes that all API communication internally is encrypted with TLS.
```
artificial-intelligence, line 69: where does this expectation come from? it's not mentioned as a requirement in this standard prior to this sentence.

I think we should add this.

Standards/scs-0217-v1-baseline-cluster-security.md
```diff
@@ -0,0 +1,145 @@
...
+## Motivation
+
+Kubernetes clusters are highly configurable, which also gives rise to different security
+problems, if the configuration isn't done properly.
+These security risks can potentially be exposed in many different parts of a cluster, e.g.
+different APIs, authorization and authentication procedures or even Pod privilege mechanisms.
+In order to mitigate these problems, different steps and mechanisms could be used to increase
+the security of a Kubernetes setup.
+
+## Design Considerations
```
artificial-intelligence, line 38: I think this section should maybe rather be called "Implementation" or something like this? Not sure though.

Standards/scs-0217-v1-baseline-cluster-security.md
```diff
@@ -0,0 +1,145 @@
...
+
+After that, the Kubelet calls the `SubjectAccessReview` API in order to determine the authorization of a request.
+
+## Decision
+
+This standard tries to increase security for a Kubernetes cluster in order to provide a
+solid baseline setup with regard to security. For this to work, multiple measures need to be undertaken.
+
+A self-controlled CA SHOULD be used in order to be in control of the TLS certificates, which
+enables the operator to provide and revoke certificates according to the requirements.
```
artificial-intelligence, line 122: according to which requirements? Requirements of this or another standard? Requirements of the cloud operator?

Please describe the requirements clearly. Also, if there are any clear requirements, please explain why these need a custom internal CA and not a standard browser CA.

Standards/scs-0217-v1-baseline-cluster-security.md
```diff
@@ -0,0 +1,145 @@
...
+
+All internal endpoints found in the section [TLS for all internal/API traffic] MUST be
+encrypted with TLS in order to secure internal traffic.
+
+The Kubernetes API (kubeAPI) MUST be secured by authenticating and authorizing the users
+trying to access its endpoints. How a user is authenticated is up to the provider of the
+cluster and/or the wishes of the customer. Authorization MUST be done by providing
+fine-grained RBAC.
+The authentication and authorization steps MUST also be applied to the Kubelet, which in its default configuration
+doesn't enable them. A way to do this can be found in the section [Authentication and Authorization].
```
artificial-intelligence, line 132: I find this very unspecific for a standards document.

Especially with regards to security this allows for many possible implementations which usually leads to - inadvertently - insecure implementations because there is no clear mandatory way to configure things in a secure fashion.

this is what opened many security holes in previous TLS standards, for example.

See this RFC for recommendations on how to write security related standards documents:

https://www.rfc-editor.org/rfc/rfc3552.html#page-26

I want to particularly highlight something relevant to the above:

> If the technology involves authentication, particularly user-host
   authentication, the security of the authentication method MUST be
   clearly specified.  That is, authors MUST document the assumptions
   that the security of this authentication method is predicated upon.

So, to sum it up:

If you don't enforce secure authentication and authorization by the standard it's a matter of time until someone implementing your standard will end up implementing it in an insecure manner.

It is also not sufficient that the standard relies on third party security-by-accident schemes, made up example:

"the supported authentication schemes in kubernetes are all secure, therefore we do not need to specify explicitly secure auth schemes".

A standard should not rely on implementation specific behaviour, which might change in the future, or might be version specific.

Be explicit what is secure and be strict, not liberal in what you accept, see this famous criticism of the robustness principle:

https://datatracker.ietf.org/doc/html/draft-iab-protocol-maintenance

HTH

artificial-intelligence (changes requested):

Standards/scs-0217-v1-baseline-cluster-security.md
cah-hbaum: I can understand that. I thought it was better for general text flow, but I don't need the sentence in, so I removed it.
But I don't have a "better" introduction, so we will keep with that.

Standards/scs-0217-v1-baseline-cluster-security.md
cah-hbaum: https://kubernetes.io/docs/tasks/administer-cluster/securing-a-cluster/#use-transport-layer-security-tls-for-all-api-traffic
From here, but I can mention this if that would make it more clear.

Standards/scs-0217-v1-baseline-cluster-security.md
cah-hbaum: I can understand that, but I think we would need to adapt the Template and "Standard" Standard to change that.
(Not sure if the template sections are that binding, but personally, I would like to stick to that).

Standards/scs-0217-v1-baseline-cluster-security.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Related Documents
+
+[Managing TLS in a cluster](https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/)
+[Securing a cluster](https://kubernetes.io/docs/tasks/administer-cluster/securing-a-cluster/)
+[Controlling access](https://kubernetes.io/docs/concepts/security/controlling-access/)
+[Kubelet Authentication and Authorization](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-authn-authz/)
+[Authentication](https://kubernetes.io/docs/reference/access-authn-authz/authentication/)
+[Kubernetes Security](https://cheatsheetseries.owasp.org/cheatsheets/Kubernetes_Security_Cheat_Sheet.html)
```
bitkeks, line 137:
```suggestion
* [Managing TLS in a cluster](https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/)
* [Securing a cluster](https://kubernetes.io/docs/tasks/administer-cluster/securing-a-cluster/)
* [Controlling access](https://kubernetes.io/docs/concepts/security/controlling-access/)
* [Kubernetes Security Checklist](https://kubernetes.io/docs/concepts/security/security-checklist/)
* [Kubelet Authentication and Authorization](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-authn-authz/)
* [Authentication](https://kubernetes.io/docs/reference/access-authn-authz/authentication/)
* [OWASP Kubernetes Security Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Kubernetes_Security_Cheat_Sheet.html)
```

Standards/scs-0217-v1-baseline-cluster-security.md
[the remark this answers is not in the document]
bitkeks: See my suggestion to format it as a list.

Standards/scs-0217-v1-baseline-cluster-security.md
```diff
@@ -0,0 +1,141 @@
...
+|----------|-------------|-------------------------|-----------------------------------------------------------------------------------------|
+| TCP      | 6443*       | Kubernetes API Server   | -                                                                                       |
+| TCP      | 2379-2380   | etcd server client API  | -                                                                                       |
+| TCP      | 10250       | Kubelet API             | -                                                                                       |
+| TCP      | 10251/10259 | kube-scheduler          | 10251 could be insecure before 1.13, after that only the secure port 10259 is available |
+| TCP      | 10252/10257 | kube-controller-manager | 10252 could be insecure before 1.13, after that only the secure port 10257 is available |
+| TCP      | 10255       | Read-Only Kubelet API   | HTTP port for GET requests                                                              |
+| TCP      | 30000-32767 | NodePort Services       | Service endpoints, could be HTTP                                                        |
+
```
bitkeks, line 80:
```suggestion
| Protocol | Port Range  | Purpose                 | Notes                                                                                   |
|----------|-------------|-------------------------|-----------------------------------------------------------------------------------------|
| TCP      | 6443*       | Kubernetes API Server   | -                                                                                       |
| TCP      | 2379-2380   | etcd server client API  | -                                                                                       |
| TCP      | 10250       | Kubelet API             | -                                                                                       |
| TCP      | 10251/10259 | kube-scheduler          | 10251 could be insecure before 1.13, after that only the secure port 10259 is available |
| TCP      | 10252/10257 | kube-controller-manager | 10252 could be insecure before 1.13, after that only the secure port 10257 is available |
| TCP      | 30000-32767 | NodePort Services       | Service endpoints, could be HTTP                                                        |

The usage of `readOnlyPort` (enabling a read-only Kubelet API port on 10255) by design neither provides authentication nor authorization. Its usage is strongly discouraged!

```

Standards/scs-0217-v1-baseline-cluster-security.md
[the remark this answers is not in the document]
bitkeks: I proposed a remark, that would work for me. Retaining the table as is would also work for me btw.

bitkeks (approved):

Standards/scs-0217-v1-baseline-cluster-security.md
cah-hbaum: Updated the table!

Standards/scs-0217-v1-baseline-cluster-security.md
cah-hbaum: Updated the list!

jschoone:
Hi @bitkeks, @cah-hbaum and @joshmue. What's the state here? It looks like we have enough approvals to merge before it gets more complicated.

bitkeks:
LGTM, already approved!

@artificial-intelligence has remaining changes that need to be resolved.

cah-hbaum:
TBF I don't know how to adapt the document to "appeal" more to the requested changes from @artificial-intelligence. Either he suggests some general changes (like he's written in his comment) or approves or we will be stuck here.

jschoone:
Hi @artificial-intelligence, could you please add suggestions to your change requests? Otherwise we would like to merge, since this is in Draft state we can add improvements later.

cah-hbaum:
Pinging @artificial-intelligence again, since the last time was two weeks ago. I still don't know how to change the paragraphs up in order to make it compliant with your vision of them.

Standards/scs-0217-v1-baseline-cluster-security.md
artificial-intelligence: There are neither operator requirements nor requirements about CAs in the linked issue, other than that an unspecified "external CA" should be used.

I guess we can move on and don't change this, as there don't actually seem to be any written down requirements for CAs anywhere here which we could explicitly mention, especially with regards to security.

Standards/scs-0217-v1-baseline-cluster-security.md
artificial-intelligence: Why was this marked as "resolved" @jschoone ? There wasn't even any effort done to reply to this review in any form.

If you disagree with the stated observations and opinions in a review please state that in written form. Why do you think this is not an issue or why it does not matter, or how this is already addressed in the standard or elsewhere?

Just marking stuff as "resolved" in order to unblock work devalues the work of the reviewer and is a really bad practice imho.

Please don't do that anymore, at least for my reviews.

artificial-intelligence:
> Hi @artificial-intelligence, could you please add suggestions to your change requests? Otherwise we would like to merge, since this is in Draft state we can add improvements later.

I don't find it reasonable to expect from reviewers to redraft completely missing necessary implementation details for a standards document.

This is work that is expected to be done by the authors of the document imho, because that is the main thing which needs to be done, when drafting such a document.

I'll thus remove myself from the reviewers of the document. I don't have currently the time to write the necessary detailed spec to make authn and authz reasonable secure. As this is completely left for the implementer to decide - and people will get this wrong, sooner or later - I guess there is no consensus here on how to do secure RBAC authn/autz, so even if I had the time, I can't write something down when there is no consensus, which is the point of standards.

I wish you good luck with the ongoing effort to secure Kubernetes.

Standards/scs-0217-v1-baseline-cluster-security.md
artificial-intelligence: marking this as "resolved" as I don't want to block you.

artificial-intelligence:
I don't seem to be able to remove myself from this review somehow. Two days ago I didn't show up when I removed myself. Now I'm again listed as a reviewer.

artificial-intelligence (commented):
It seems I at least need to leave a comment to get out of here, so here you are :slightly_smiling_face:

artificial-intelligence (approved):

artificial-intelligence:
It seems github review is totally broken. I need to either request changes (I didn't want to request further changes) or I need to approve the change (I honestly don't want to approve it).
I can't remove myself from the review, at this stage, through the Github UI.

But as we are stuck with this, I just "approved" this now.

cah-hbaum:
Merging this now, we will continue in https://github.com/SovereignCloudStack/standards/issues/475 to address issues brought up here.

PR #377, opened 2023-11-09
Add units for disk and ram size of images

### LINKED ISSUE #367: [BUG] Standard for image metadata omits unit for "min_disk_size" and "min_ram"
[Image metadata standard](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0100-v3-flavor-naming.md) refers to [useful-image-properties](https://docs.openstack.org/glance/latest//admin/useful-image-properties) for mandatory and recommended Image metadata. Two mandatory data are minimal size disk and ram- However, neither [useful-image-properties](https://docs.openstack.org/glance/latest//admin/useful-image-properties) nor [Image metadata standard](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0100-v3-flavor-naming.md) defines unit of ram and disk size. IMO this information is obligatory and MUST be added to Image property standard.

anjastrunk:
closes #367

### DIFF
--- a/Standards/scs-0102-v1-image-metadata.md
+++ b/Standards/scs-0102-v1-image-metadata.md
@@ -47,9 +47,11 @@ in the [OpenStack Image documentation](https://docs.openstack.org/glance/latest/
 The following properties are considered mandatory:
 
 - `architecture`, `hypervisor_type`
-- `min_disk_size`, `min_ram`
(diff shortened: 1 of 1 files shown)


mbuechse (approved):
I think this is very helpful. Particularly given the fact that the Glance docs mentioned in the section in question do not even mention those two properties!

PR #379, opened 2023-11-11
Display the results of the v2 compliance checks.

garloff:
v1 is out of business sinve Nov 1, 2023, so let's not display v1 information any longer. v2 is valid for another month, which is when we will require v3 (unless we decide on extending v2 validity). Main challenge is that v3 flavors require the two flavors with local SSDs (and that we dropped the requirement for all other flavors with local disks).


mbuechse:
@frosty-geek @fkr @garloff This seems important. How do we proceed?

mbuechse:
I'm pretty sure this is obsolete by now.

PR #384, opened 2023-11-18
WIP: Adjust Domain Manager policies for Horizon and Terraform compatibility

markus-hentsch:
Overhauls the Domain Manager policy configuration to improve compatibility with Horizon and Terraform.

Note the following section from the corresponding issue:

> FTR, we discussed[^1] that we do not want to jump the gun and implement full Horizon support for Domain Managers in a way that enables the Domain Manager workflow in the dashboard. Reasoning was that for Domain Managers limited API access (e.g. Keystone only) might be provided and full dashboard usage might not be intended to begin with. Furthermore, debugging and development effort would be required to make Horizon compatible (see comments above).
> 
> Due to this, the work on this issue and the linked PR **will address**:
> 
> - proper interaction for normal users (excluding Domain Managers) with the dashboard when the Domain Manager Standard is deployed
> - improved Terraform API compatibility for Domain Manager users
> 
> The PR **will not address**:
> 
> - Domain Manager functionality (CRUD for users, groups, projects, role assignments in domains) on the Horizon dashboard

[^1]: https://github.com/SovereignCloudStack/minutes/blob/main/iaas/20231122.md?plain=1#L107-L110


berendt:
This modification is also required to make the `/identity/users` dashboard usable.

```
identity:get_user": "(rule:is_domain_manager and token.domain.id:%(target.user.domain_id)s) or user_id:%(user_id)s or rule:admin_required"
```

berendt:
Project details are also not working. Probably `identity:get_project` has to be modified as well.

You can reproduce this by visiting a listed user in `/identity/users`. There click on the primary project of the users you are logged with.

markus-hentsch:
> This modification is also required to make the `/identity/users` dashboard usable.
> 
> ```
> identity:get_user": "(rule:is_domain_manager and token.domain.id:%(target.user.domain_id)s) or user_id:%(user_id)s or rule:admin_required"
> ```

FTR, this concerns non-domain-manager users, for which it is actually necessary. It does not make a difference for Domain Managers.


> Project details are also not working. Probably `identity:get_project` has to be modified as well.
> 
> You can reproduce this by visiting a listed user in `/identity/users`. There click on the primary project of the users you are logged with.

I can reproduce this with all policy overrides disabled (i.e. deleting `policy.yaml` and restarting Keystone) and using Keystone's default policies. Thus, I don't think this behavior is caused by the Domain Manager configuration.

josephineSei:
@berendt Could you test these changes again? I will take over from @markus-hentsch and work on any further bugs you might find.

berendt:
@josephineSei I cannot reproduce it right now. I have added the state from this PR in our testbed and it runs both the playbook for the test (https://github.com/osism/testbed/blob/main/environments/openstack/playbook-test-identity.yml) and Refstack (2022.11) completely.

berendt:
Fro my POV it can be merged.

josephineSei (approved):
I looked through the changes and it seems we got a whole policy file within this standard. I only have the question, whether we want this in the standard or maybe a separated extra file?

The last option has the positive side effect, that, when something in the upstream policy changes we do not have to adjust the standard. But maybe that is exactly what we want.

garloff:
When we split out implementation notes out of standards, we might move the policy file there.
This should not prevent us from merging the standard now, though.

PR #388, opened 2023-11-23
Add description field to front matter

mbuechse:


mbuechse:
The broken links are not due to me!

maxwolfs:
The dead links should be fixed in another PR. I will create Issues with the dead links in the documents and transform the link-checking into a cron job that checks daily and not only on pull request (which leads to this bad behaviour).

PR #389, opened 2023-11-25
Feat/add flavor page

garloff:
Code for PoC for https://github.com/SovereignCloudStack/issues/issues/483

Totally Web 1.0, that's my generation ...

### DIFF
--- a/Standards/scs-0100-v3-flavor-naming.md
+++ b/Standards/scs-0100-v3-flavor-naming.md
@@ -271,6 +271,9 @@ It goes beyond the above example in checking that the discoverable
 features of flavors (vCPUs, RAM, Disk) match what the flavor names claim.
 This is used for SCS-compatible compliance testing.
 
+The web page <a href="https://flavors.scs.community/">https://flavors.scs.community"</a>
+has a flavor name parser and (soon) generator.
+
 ## Extensions
 
 Extensions provide a possibility for providers that offer a very differentiated set
--- a/Standards/scs-0100-v3-flavor-naming.md
+++ b/Standards/scs-0100-v3-flavor-naming.md
@@ -271,7 +271,7 @@ It goes beyond the above example in checking that the discoverable
 features of flavors (vCPUs, RAM, Disk) match what the flavor names claim.
 This is used for SCS-compatible compliance testing.
 
-The web page <a href="https://flavors.scs.community/">https://flavors.scs.community"</a>
+The web page <https://flavors.scs.community/>
 has a flavor name parser and (soon) generator.
 
 ## Extensions
--- a/Standards/scs-0100-v3-flavor-naming.md
+++ b/Standards/scs-0100-v3-flavor-naming.md
@@ -271,8 +271,8 @@ It goes beyond the above example in checking that the discoverable
 features of flavors (vCPUs, RAM, Disk) match what the flavor names claim.
 This is used for SCS-compatible compliance testing.
 
-The web page <https://flavors.scs.community/>
-has a flavor name parser and (soon) generator.
+The functionality of the `flavor-name-check.py` script is also
+(partially) exposed via the web page <https://flavors.scs.community/>.
 
 ## Extensions
 
--- a/Tests/iaas/flavor-naming/flavor-form.py
+++ b/Tests/iaas/flavor-naming/flavor-form.py
@@ -4,13 +4,22 @@
 #
 # (c) Kurt Garloff <kurt@garloff.de>, 11/2023
 # SPDX-License-Identifier: CC-BY-SA-4.0
-import os
+"""
+flavor-form.py
+CGI script to get passed flavor from a html form (GET)
+and parses it according to SCS flavor naming.
+It returns an error (sometimes with a useful error message)
+or a human-readable description of the flavor.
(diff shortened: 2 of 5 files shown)


berendt:
Works for me. I already added it to the flavor manager docs:

https://osism.github.io/docs/guides/operations-guide/openstack/flavor-manager/#name-parser-and-generator

mbuechse (commented):
Nice! From my POV, two things should be improved though:

- the script should output valid HTML; to that end, it could read the index.html and insert its output into some conveniently marked (easy to find) position
- the error messages should also be improved; all I got at first try was `TypeError: cannot unpack non-iterable NoneType object` -- it could at least say "The cause is probably that the name given is not a valid SCS flavor name"

As always, I volunteer to implement these or similar improvements.

garloff:
> Nice! From my POV, two things should be improved though:
> 
>     * the script should output valid HTML; to that end, it could read the index.html and insert its output into some conveniently marked (easy to find) position

I had this on my list as well.
I would actually generate the main page from the script and embed the flavor parser (and later flavor generator) output in it. Maybe using some simple templating with plone or so would make this easy.
I also need to replace the `cgi` module with `urllib`.

>     * the error messages should also be improved; all I got at first try was `TypeError: cannot unpack non-iterable NoneType object` -- it could at least say "The cause is probably that the name given is not a valid SCS flavor name"

For some wrong flavors you do get useful hints where the parser failed, for some you don't.
My guess is that there are a few things that you can easily improve in the existing implementation, but I would be surprised if you are able to output meaningful error messages in all cases without adding a lot of error handling code.

An option is of course to use some open source tooling that is meant to support syntax parsing rather than coding it yourself. I had a quick check back then and decided that coming up with my own is easier. But that may have been due to insufficient research.

If you go there, we'd have a second independent implementation of the flavor spec, which may help with testing.
 
> As always, I volunteer to implement these or similar improvements.

I would want to take the easy steps (using templating to generate good HTML) quickly but then do a first merge.
The more tedious error message improvements can always come in a second step.
Oh, and of course if someone wants to use less antique Web1.0 technology ...

garloff:
So, this should be producing good HTML now.
It also outputs "non-SCS flavor" for flavors not starting with `SCS-`.
Next step will be to support flavor name generation, but that needs a somewhat complex form, which I can't do in a few minutes ... so I'd like this one to be merged. Reviews welcome!

Tests/iaas/flavor-naming/flavor-form.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env python3
+# Little wrapper to output flavor pretty description
+# from a CGI form with flavor
+#
+# (c) Kurt Garloff <kurt@garloff.de>, 11/2023
+# SPDX-License-Identifier: CC-BY-SA-4.0
```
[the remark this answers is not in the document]
garloff: All code in this directory is CC-BY-SA-4.0, which is a weak copyleft license.

CC is not a good license for compiled code, as it does not have any provisions for source code availability when distributing binary code. That is not a problem for scripting languages ...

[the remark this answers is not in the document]
mbuechse: I'm not sure if one could distribute pyc files. I guess it would probably be a hassle for everyone.

Standards/scs-0100-v3-flavor-naming.md
```diff
@@ -271,6 +271,9 @@ It goes beyond the above example in checking that the discoverable
 features of flavors (vCPUs, RAM, Disk) match what the flavor names claim.
 This is used for SCS-compatible compliance testing.
 
+The web page <https://flavors.scs.community/>
+has a flavor name parser and (soon) generator.
```
mbuechse, line 275: Maybe we should add that the functionality of the website is analogous to some of the scripts mentioned above (and in what way).

Tests/iaas/flavor-naming/flavor-form.py
```diff
@@ -0,0 +1,73 @@
...
+CGI script to get passed flavor from a html form (GET)
+and parses it according to SCS flavor naming.
+It returns an error (sometimes with a useful error message)
+or a human-readable description of the flavor.
+"""
+
+# import os
+import sys
+import re
+# import traceback
```
mbuechse, line 18: Shall we get rid of the imports that are commented out?

mbuechse (approved):
Overall fine, only very minor things in the comments.

Plus: could you add some snippet documenting how you integrated the cgi script into the website? Some config for nginx or Apache?

Standards/scs-0100-v3-flavor-naming.md
garloff: "The flavor name parsing and (soon) flavor name generation of these scripts is also exposed via the web page <https://flavors.scs.community/>."

Tests/iaas/flavor-naming/flavor-form.py
garloff: Those were required before but have been replaced ...
Feel free to push a change to remove them.

joshmue:
For future reference: There was a injection bug here that's fixed in #403.

Issue #386, opened 2023-11-23
[BUG] Incorrect 'type' header field in scs-0301-v1


mbuechse:
In [scs-0301-v1](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0301-v1-naming-conventions.md), the type header field reads `_Standard | Decision Record_`, which is inconsistent with the format laid down in [scs-0001-v1](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0001-v1-sovereign-cloud-standards.md#document-format); in particular, the vertical bar upsets [a certain Markdown table](https://docs-staging.scs.community/standards/iam/scs-0301). It could probably be escaped, but since it's against the aforementioned format, it should be changed.

NB: In the future, it might not be possible to set the header field to a 'wrong' value, see #385.

One other thing I noticed is that the title of the document is quite long. Maybe you can find a more succinct one? This would also help the rendering of [a certain table](https://docs-staging.scs.community/standards/standards/overview/) ;)

JuanPTM:
I have created a merge request to solve this issue. https://github.com/SovereignCloudStack/standards/pull/390
I assign you as reviewer.

mbuechse:
Closed because it's been resolved via #390.

PR #390, opened 2023-11-27
 [BUG] Incorrect 'type' header field in scs-0301-v1 #386

JuanPTM:
Fixed the 'type' and shorten the name of the naming ADR.


mbuechse (approved):
Thanks!

PR #351, opened 2023-09-29
Volume type standard

josephineSei:
This creates the new standard for volume types as mentioned in ticket #265

PR #395, opened 2023-12-01
Create scs-xxxx-volume-type-standard

josephineSei:
shortened version of the volume type standard without any optional parameters

as alternative to PR #351


josephineSei:
abandoned duplicate

PR #404, opened 2023-12-06
Use html.escape for all output that may be user-influenced.

garloff:
This is cleaner than filtering the input.
Suggested by @joshmue.


joshmue (approved):
LGTM!

Issue #385, opened 2023-11-23
Automatically check front matter of standards documents via a Zuul job


mbuechse:
The Procedural [scs-0001-v1](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0001-v1-sovereign-cloud-standards.md) lays down the format of the standards documents, including the front matter (therein called "Markdown header"). With the automated generation of the docs overview pages (see https://github.com/SovereignCloudStack/docs/issues/97), it became apparent that some files do not conform to this format; for instance:

- wrong track field for 0412
- missing track field, missing status field for 0100-v1
- wrong type field for 0301
- 'status' field erroneously called 'state' for 0002-v2

The goal of this issue is to have a Zuul job in the check pipeline that recognizes mistakes of these sorts and prevents them from being checked in.

**edit** in addition to the front-matter fields, also naming collisions shall be tested for

PR #406, opened 2023-12-07
Rename file to resolve collision

mbuechse:
0215 is already being worked on, so I opted for 0216

In the future, collisions shall be prevented by a gate job; see #385


maxwolfs (approved):
LGTM

PR #407, opened 2023-12-07
Reflect current state in the header of scs-0100-v1 and scs-0100-v2

mbuechse:


garloff:
Yes, this is correct.

garloff (approved):
/lgtm

PR #409, opened 2023-12-08
Add reference to the upstream bug about list_domains endpoint

markus-hentsch:
Links to the upstream bug reported as part of SovereignCloudStack/issues#446 via footnote in the appropriate section.


garloff (approved):
/lgtm
Thanks, @markus-hentsch

PR #410, opened 2023-12-18
Acc. to scs-0102-v1, os_hash_algo is recommended.

garloff:
So let's not error out, but just display a warning. Also don't gate using non-recommended names with verbose setting, but display these warnings as well. Capitalize displayed suggested name.

### DIFF
--- a/Tests/iaas/image-metadata/image-md-check.py
+++ b/Tests/iaas/image-metadata/image-md-check.py
@@ -142,6 +142,7 @@ def validate_imageMD(imgnm):
         return 1
     # Now the hard work: Look at properties ....
     errors = 0
+    warnings = 0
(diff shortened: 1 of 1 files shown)


garloff:
Note: Normally, upon registering an image, OpenStack does calculate an `os_hash_value` and sets it and `os_hash_algo` accordingly. However, according to the letter of [scs-0102-v1](https://docs.scs.community/standards/scs-0102-v1-image-metadata) we only recommend and do not require it. So we can't error out upon a missing `os_hash_algo`.
This caused a spurious PCO IaaS SCS-compatible check failure for the image `plusKI CUDA 12.1`.
@costelter - You might still want to look into why your image has no `os_hash_*` values set.

fkr (approved):

PR #411, opened 2023-12-18
Accept private image_source as set in the standard.

garloff:
We do produce an info if verbose mode -v is used, as this is of course not desirable, so it's worth a note.

### DIFF
--- a/Tests/iaas/image-metadata/image-md-check.py
+++ b/Tests/iaas/image-metadata/image-md-check.py
@@ -121,7 +121,7 @@ maint_props = (Property("replace_frequency", True, ("yearly", "quarterly", "mont
 
 
 def is_url(stg):
-    "Is string stg an URL?"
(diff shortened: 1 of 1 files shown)


garloff:
This caused trouble with the `VyOS 1.4 rolling 20221121` image which has the `image_source` value set to `private`.
While not preferred, it's legit, so we have to accept it.

fkr (approved):
I do think doing grammar / spelling fixes should be a separate commit from a change that changes the logic (since when backing out the functional change also spelling and grammar fixes are lost).

garloff:
Right.
I do see these things when I look at the code and can't stop myself from fixing.
You are absolutely right, they should be in a different commit.
For larger changes, I'm more disciplined, but I have to admit that I tend to do these smaller fixes along the way as it saves me some time. Sorry for that!

PR #413, opened 2023-12-18
Feat/add addtl img md checks

### LINKED ISSUE #412: [Feature Request] image-md-check.py: Missing checks
[image-md-check.py](https://github.com/SovereignCloudStack/standards/blob/e9590ed94c0489fa2d7662e3d6c724f9e54e1346/Tests/iaas/image-metadata/image-md-check.py#L179) has a few TODOs to add sanity checks:
```
    #  - uuid_validity has a distinct set of options (none, last-X, DATE, notice, forever)
    #  - hotfix hours (if set!) should be numeric
    # (5a) Sanity: Are we actually in violation of update_frequency?
    #  This is a bit tricky: We need to disregard images that have been rotated out
    #  - os_hidden = True is a safe sign for this
    #  - A name with a date stamp or old or prev (and a newer exists)
    # (2) sanity min_ram (>=64), min_disk (>= size)
    # (6) tags os:*, managed_by_*
```

garloff:
Closes #412

### DIFF
--- a/Tests/iaas/image-metadata/image-md-check.py
+++ b/Tests/iaas/image-metadata/image-md-check.py
@@ -14,6 +14,7 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 import os
 import sys
 import time
+import calendar
 import getopt
 import openstack
 
@@ -49,6 +50,22 @@ mand_images = ["Ubuntu 22.04", "Ubuntu 20.04", "Debian 11"]
 rec1_images = ["CentOS 8", "Rocky 8", "AlmaLinux 8", "Debian 10", "Fedora 36"]
 rec2_images = ["SLES 15SP4", "RHEL 9", "RHEL 8", "Windows Server 2022", "Windows Server 2019"]
 sugg_images = ["openSUSE Leap 15.4", "Cirros 0.5.2", "Alpine", "Arch"]
+
+# Just for nice formatting of image naming hints -- otherwise we capitalize the 1st letter
+os_list = ("CentOS", "AlmaLinux", "Windows Server", "RHEL", "SLES", "openSUSE")
+
+
+def recommended_name(nm):
+    "Return capitalized name"
+    ln = len(nm)
+    for osnm in os_list:
+        osln = len(osnm)
+        if ln >= osln and nm[:osln].casefold() == osnm.casefold():
+            rest = ""
+            if ln > osln:
+                rest = nm[osln:]
+            return osnm + rest
+    return nm[0].upper() + nm[1:]
 
 
 def get_imagelist(priv):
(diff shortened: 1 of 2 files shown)


Tests/iaas/image-metadata/image-md-check.py
```diff
@@ -121,17 +145,70 @@ def is_ok(self, props, warn = ""):
...
+    until_str = img.properties["provided_until"]
+    until = parse_date(img.properties["provided_until"])
+    if not until and not until_str == "none" and not until_str == "notice":
+        print(f'ERROR: Image "{img.name}" does not provide a valid provided until date',
+              file=sys.stderr)
+        return 3
+    if time.time() > until:
+        return 0
+    if img.is_hidden or img.name[-3:] == "old" or img.name[-4:] == "prev" or img.name[-8:].isdecimal():
+        return 1
```
mbuechse, line 201: The standard states

> It is recommended that the old image gets renamed (e.g. build date or patch level attached)

There's no mention of "old" nor "prev". On the other hand, a patch level is mentioned, and that will not be recognized by `is_outdated`, will it?

I think we need to work this out more, but I'm afraid this would need a more strict standard: either make the `os_hidden` mandatory, or mandate some other clear marker (such as a new property `original_name` that has to be set to the name of the current version, which must be a prefix of the name of the image in question).

mbuechse: A funny observation: if the patch level is "Gold", then it will be recognized ;)

garloff: You are right, the standard should define better defined rules for renaming (or hiding) old images, so users can rely on it and the test can have a less heuristic approach to checking it.
Patch level does not have a defined format either, so we can't use it unless we define a format.
I am afraid this will require a v2 of the image metadata standard.

garloff: Moving forward short-term without a v2 standard yet (and thus in absence of strict rules that make the test reliably avoid false positives), I see a few things to decide.
* We could drop the "prev" "old" "yyyymmdd" heuristic piece -- this will cause all outdated images to end up on a list where newer images are searched. (Or use this heuristic piece only to trigger a warning or not, but not decide whether the image ends up on the list.) Actually I think we should do this!
* When outdated images are found without a replacement being recognized by the script, we currently emit an error. We could downgrade this to a warning to avoid false positives. On the other hand, all existing SCS clouds pass this and we could probably live with the risk and wait for cloud providers to really trigger this ... and then react by either improving the algorithm further or then transforming the error into a warning. (I prefer this.)

Tests/iaas/image-metadata/image-md-check.py
```diff
@@ -217,11 +329,45 @@ def report_stdimage_coverage(imgs):
...
+        shortnm = outd
+        if last_spc != -1:
+            shortnm = outd[:last_spc]
+        for imgnm in images:
+            # Skip over other images
+            if imgnm != outd and imgnm != shortnm:
+                continue
+            # Skip over itself
+            if imgnm == outd:  # or success:
+                continue
```
mbuechse, line 347: I guess we can move this up, and then the check above becomes `if imgnm != shortnm:`
Also, if no space is found in the image name, then we can skip this whole for loop altogether.

Tests/iaas/image-metadata/image-md-check.py
```diff
@@ -217,11 +329,45 @@ def report_stdimage_coverage(imgs):
...
+            img = conn.image.find_image(imgnm)
```
mbuechse, line 348: I know this script can be invoked with a list of image names, but usually it will check every image, and so it will fetch the image info for every image. Wouldn't it be better to do this once, with a single API call (maybe multiple, but few calls if paging is involved)?

Tests/iaas/image-metadata/image-md-check.py
```diff
@@ -121,17 +145,70 @@ def is_ok(self, props, warn = ""):
...
+def freq2secs(stg):
+    """Convert frequency to seconds (round up a bit), return 0 if not applicable"""
+    secs = FREQ_TO_SEC.get(stg)
+    if secs is None:
+        print(f'ERROR: replace frequency {stg}?', file=sys.stderr)
+        secs = 0
+    return secs
+
+
+OUTDATED_IMAGES = []
```
mbuechse, line 179: Since I have an aversion to mutable globals: I would move this variable into `main` as a local, and then pass it to the relevant functions. Parameters are passed by value, but the value in this case would be the list object (the envelope, if you will) which is itself mutable.

mbuechse: I guess that was not quite right. The value is the reference to the list object, and the list object is mutable.

Tests/iaas/image-metadata/image-md-check.py
mbuechse: Sorry, I forgot to say that I agree.

mbuechse:
This has been lingering for so long. @garloff seems to be busy doing other things. I will try to move this forward.

mbuechse (approved):
Fine by me, even though it can still be improved. @garloff have a look, please.

PR #431, opened 2023-12-20
Refactor to reduce degree of nesting

mbuechse:


garloff (approved):
I like it.
With the one improvement I suggested.
I would prefer it to me merged after after #413 though to avoid merge conflicts.

PR #432, opened 2023-12-20
Feat/add artcodix

garloff:
Note: We need a separate PR to also expose this on docs.scs.community/standards/


garloff:
I believe the CI test is producing a false positive here ...

fkr (approved):

garloff:
> I believe the CI test is producing a false positive here ...

That was wrong. There was a typo in my link to the badge, see #433.

mbuechse:
@garloff it would have been nice to add artcodix to the zuul tests as well!

garloff:
Let's do that in the next step.

PR #434, opened 2023-12-21
Fix typos in scs-0102-v1-image-metadata.md

mbuechse:


garloff (approved):
Thanks for spotting these, @mbuechse !
/lgtm

PR #435, opened 2023-12-21
Automatically check front matter of standards documents via a Zuul job

### LINKED ISSUE #385: Automatically check front matter of standards documents via a Zuul job

mbuechse:
closes #385

### DIFF
--- a/.zuul.d/config.yaml
+++ b/.zuul.d/config.yaml
@@ -16,6 +16,13 @@
     post:
       jobs:
         - scs-check-gx-scs
+    check:
+      jobs:
+        - scs-check-adr-syntax
+- job:
+    name: scs-check-adr-syntax
+    parent: base
+    run: playbooks/adr_syntax.yaml
 - job:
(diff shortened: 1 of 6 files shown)


garloff:
Nice work!

PR #439, opened 2023-12-28
Fix OHM links for PCO prod2 & prod3

berendt:


berendt:
@garloff Looks like prod2 & prod3 OHMs are not yet there?

garloff:
> @garloff Looks like prod2 & prod3 OHMs are not yet there?

They are. Running on the same machine in Prod1. Just select cloud Prod2 and Prod3.

PR #440, opened 2024-01-03
Add draft for SONIC.

akafazov:
Issue https://github.com/SovereignCloudStack/issues/issues/508

ADR for SONiC support in SCS.

Comments for reviewers:
- which track is the correct for this ADR? Currently `Global` but `IaaS` may also apply.

Other resources:
- [Research SDN scalability](https://input.scs.community/VP04-issues-455-research-SDN-scalability)
- [Results SONIC Usage in SCS](https://input.scs.community/SCS-DR-SONIC-usage)
- [SONiC Community research](https://input.scs.community/oW_plPZ6RkuXs3k9mrRDdw#)

### DIFF
--- a/Standards/scs-0112-v1-sonic.md
+++ b/Standards/scs-0112-v1-sonic.md
@@ -82,9 +82,15 @@ Good initial adoption: Microsoft, Target. Adoption requires time and money
 
 Actually not, because of different use cases.
 
+- Sustainability of community SONIC for 2025+
+
+The SONiC community is healthy and growing, however progress is slower due to factors like investment of resources. The barrier of entry is much higher than other similar OSS projects.
+
 ## Decision
 
 IaaS team recommends to use Option 3: SCS develops SCS-specific modules as add-on for any SONiC (Community or Enterprise). It has the best tradeoff between time and resource investment and benefits for the community. Adopting this strategy would allow SCS to be agile and quickly adopt SONiC, by providing users with clear path while allowing the freedom to chose different hardware and software vendors. SCS code can be packaged independently of each SONiC distribution and installed as add-on. Also SCS contributions to core SONiC will be done directly upstream, so that the whole community can benefit from them.
+
+Work on hardware support in SONiC should be raised in upstream and SCS shouldn't make significant investments in this area.
 
 ## Related Documents
 
--- a/Standards/scs-0112-v1-sonic.md
+++ b/Standards/scs-0112-v1-sonic.md
@@ -38,7 +38,7 @@ There is no tooling in SCS or Openstack communities to facilitate the rollout an
 
 ### OVN for SONiC
 
-OVN and OVS are extensively reveraged in Neutron to SDN. In large scale deployments, OVN can become a bottleneck by exhausing resources on network nodes. SONiC can host the OVN control plane as a module (container) and spare the resources in network nodes. There is however a potential other bottleneck on SONiC hardware, as OVN control plane can be resource intensive. This is a potential area for further investigation.
+OVN and OVS are extensively leveraged in Neutron to SDN. In large scale deployments, OVN can become a bottleneck by exhausting resources on network nodes. SONiC can host the OVN control plane as a module (container) and spare the resources in network nodes. There is however a potential other bottleneck on SONiC hardware, as OVN control plane can be resource intensive. This is a potential area for further investigation.
 
 ## Design Considerations
 
--- a/Standards/scs-0112-v1-sonic.md
+++ b/Standards/scs-0112-v1-sonic.md
@@ -2,7 +2,7 @@
 title: SONiC Support in SCS
 type: Decision Record # Standard  | Decision Record | Procedural
 status: Draft
-track: Global  # Global | IaaS | Ops | KaaS | IAM
+track: Global # Global | IaaS | Ops | KaaS | IAM
 description: |
   SCSS-0112 outlines architecutral decisions in SCS in regards to SONi support and integration.
 ---
@@ -20,7 +20,7 @@ SONiC support in [SCS](https://scs.community) was considered within the context 
 
 ## Motivation
 
-In respect to SDN scalability improvements in Openstack and SCS, there are several ways SONiC can be leveraged. 
+In respect to SDN scalability improvements in Openstack and SCS, there are several ways SONiC can be leveraged.
 
 ### SONiC as a network OS where dynamic network configuration in Openstack is required
 
@@ -82,15 +82,12 @@ Good initial adoption: Microsoft, Target. Adoption requires time and money
 
 Actually not, because of different use cases.
 
-
 ## Decision
 
 IaaS team recommends to use Option 3: SCS develops SCS-specific modules as add-on for any SONiC (Community or Enterprise). It has the best tradeoff between time and resource investment and benefits for the community. Adopting this strategy would allow SCS to be agile and quickly adopt SONiC, by providing users with clear path while allowing the freedom to chose different hardware and software vendors. SCS code can be packaged independently of each SONiC distribution and installed as add-on. Also SCS contributions to core SONiC will be done directly upstream, so that the whole community can benefit from them.
 
 ## Related Documents
 
-- https://input.scs.community/VP04-issues-455-research-SDN-scalability#SDN-scalability-improvements
-- https://input.scs.community/SCS-DR-SONIC-usage?both#SCS-Decision-Record-about-Enterprise-vs-Custom-SONiC
-- https://input.scs.community/oW_plPZ6RkuXs3k9mrRDdw#
-
-
+- [Research SDN scalability](https://input.scs.community/VP04-issues-455-research-SDN-scalability)
+- [Results SONIC Usage in SCS](https://input.scs.community/SCS-DR-SONIC-usage)
+- [SONiC Community research](https://input.scs.community/oW_plPZ6RkuXs3k9mrRDdw#)
--- a/Standards/scs-0112-v1-sonic.md
+++ b/Standards/scs-0112-v1-sonic.md
@@ -1,18 +1,11 @@
 ---
 title: SONiC Support in SCS
-type: Decision Record # Standard  | Decision Record | Procedural
+type: Decision Record
 status: Draft
-track: Global # Global | IaaS | Ops | KaaS | IAM
+track: Global
 description: |
   SCSS-0112 outlines architectural decisions in SCS in regards to SONiC support and integration.
 ---
-
-<!---
-This is a template striving to provide a starting point for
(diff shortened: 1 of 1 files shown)


ignatov17 (approved):
I agree with the presented information and documents.

Standards/scs-0112-v1-sonic.md
```diff
@@ -0,0 +1,99 @@
+---
+title: SONiC Support in SCS
+type: Decision Record # Standard  | Decision Record | Procedural
+status: Draft
+track: Global # Global | IaaS | Ops | KaaS | IAM
+description: |
+  SCSS-0112 outlines architecutral decisions in SCS in regards to SONi support and integration.
```
artificial-intelligence, line 7:
```suggestion
  SCSS-0112 outlines architectural decisions in SCS in regards to SONiC support and integration.
```

Standards/scs-0112-v1-sonic.md
```diff
@@ -0,0 +1,99 @@
...
+
+In respect to SDN scalability improvements in Openstack and SCS, there are several ways SONiC can be leveraged.
+
+### SONiC as a network OS where dynamic network configuration in Openstack is required
+
+In many network designs for Openstack, configuration of the actual network hardware by Openstack Neutron service is required. The following network designs apply:
+
+- VLANs. Uisng VLANs to segment tenant networks requires the network switch to be configured. This can be manual or dynamic configuration via the ML2 Neutron driver.
+
+- EVPN/VXLAN on the switch. In this use case, SONiC runs on leaf switches. Leafs terminal VXLAN endpoints and run BGP/EVPN for the control plane. Again, the ML2 Neutron driver is used to dynamically configure the network switch. The link between the switch and the service is regular VLAN.
```
artificial-intelligence, line 31:
```suggestion
- EVPN/VXLAN on the switch. In this use case, SONiC runs on leaf switches. Leafs terminate VXLAN endpoints and run BGP/EVPN for the control plane. Again, the ML2 Neutron driver is used to dynamically configure the network switch. The link between the switch and the service is regular VLAN.
```

Standards/scs-0112-v1-sonic.md
```diff
@@ -0,0 +1,99 @@
...
+
+- VXLAN on servers and switches. This is a hybrid use case, where most of the SDN to pushed to the server, but the network is still involved where support for bare metal hosts is needed.
```
artificial-intelligence, line 33:
```suggestion
- VXLAN on servers and switches. This is a hybrid use case, where most of the SDN is pushed to the server, but the network is still involved where support for bare metal hosts is needed.
```

Standards/scs-0112-v1-sonic.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Automation and tooling for SONiC
+
+There is no tooling in SCS or Openstack communities to facilitate with the rollout and configuration of enterprise scale SONiC deployments. Netbox and OSISM can be integrated, so that Netbox becomes the source of truth for network configuration and OSISM supports the initial rollout and configuration for the switches.
```
artificial-intelligence, line 37:
```suggestion
There is no tooling in SCS or Openstack communities to facilitate the rollout and configuration of enterprise scale SONiC deployments. Netbox and OSISM can be integrated, so that Netbox becomes the source of truth for network configuration and OSISM supports the initial rollout and configuration for the switches.
```

Standards/scs-0112-v1-sonic.md
```diff
@@ -0,0 +1,99 @@
...
+
+### OVN for SONiC
+
+OVN and OVS are extensively reveraged in Neutron to SDN. In large scale deployments, OVN can become a bottleneck by exhausing resources on network nodes. SONiC can host the OVN control plane as a module (container) and spare the resources in network nodes. There is however a potential other bottleneck on SONiC hardware, as OVN control plane can be resource intensive. This is a potential area for further investigation.
```
artificial-intelligence, line 41:
```suggestion
OVN and OVS are extensively leveraged in Neutron to SDN. In large scale deployments, OVN can become a bottleneck by exhausting resources on network nodes. SONiC can host the OVN control plane as a module (container) and spare the resources in network nodes. There is however a potential other bottleneck on SONiC hardware, as OVN control plane can be resource intensive. This is a potential area for further investigation.
```

artificial-intelligence (commented):
I only have found some minor spelling mistakes, I think.

Thanks for the work put into this! This helped me to better understand the current proposals.

Standards/scs-0112-v1-sonic.md
```diff
@@ -0,0 +1,99 @@
...
+
+#### Option 3: SCS develops SCS-specific modules as add-on for any SONiC (Community or Enterprise)
+
+In option 3, SCS will change SONiC by releasing its own modules for it. Those module can be provided as add-ons and installed on top of any version, community or enterprise. While compatability between the modules the SONiC releases will need to be maintained, there will be much broader support for SONiC and users will be able to pick and chose distributions based on their existing relationships and experience and use SCS independent of this. In cases where SCS provides contributions to core SONiC, those can be made in upstream Community repositories, so that the whole community including the propriatory vendors can adopt them eventually.
+
```
horazont, line 60: Could you please add an "Option 4: SCS does not adopt SONiC at all" which outlines the advantages of adopting SONiC in the first place?

akafazov: Thanks @horazont !
Yes, this is definitely a valid option. I will add it to the ADR.

artificial-intelligence (approved):
LGTM

Standards/scs-0112-v1-sonic.md
```diff
@@ -0,0 +1,108 @@
...
+#### Option 4: SCS does not adopt SONiC at all
+
+This option entails no dedicated effort on SCS's part in supporting SONiC network equipement for it's users and software stack. Users can still use SONiC from what is available by other projects or if they invest the effort themselves. This has several disadvantages:
```
horazont, line 63: It also has an advantage: Less of a deterrent for people who are *not* leveraging SONiC and are not intending to approach a scale where SONiC would be useful.

(yes, it is a deterrent even if it's not a requirement to use SONiC with SCS, unless it's well documented how not needed it is)

Standards/scs-0112-v1-sonic.md
```diff
@@ -0,0 +1,108 @@
...
+
+- SCS is not contributing to the SONiC community
```
horazont, line 65: In which way is this a disadvantage to the SCS project or its participants?

Standards/scs-0112-v1-sonic.md
akafazov: I do not see it as clear advantage because I don't see it as "either/or" question about if we use SONiC or not. SCS users should always have the choice, regardless if SCS invest in wider support for it. Fact is, that there are many other network OSs and SONiC is a small piece of what can be used.

Standards/scs-0112-v1-sonic.md
akafazov: The idea was that SCS adopts OSS strategy and contributing and supporting other open-source project is seen as a good thing, also on non-technical level.

Standards/scs-0112-v1-sonic.md
horazont: Of course SCS users can still opt to *not* use SONiC. Still, without clear communication that that's the case, it may be deterring to people evaluating cloud solutions. If it looks as if they need to opt into SONiC to use SCS, it will be deterring.

Which is why I said that it needs clear documentation and I think that's a disadvantage of choosing to extensively support SONiC.

artificial-intelligence: I think this is a textbook example of FUD.

Just because you are supporting solution "a", this doesn't logically imply anything regarding other solutions "b" and "c", they might or might not be equally supported. This is a logical fallacy I would expect everyone who evaluates cloud solutions to recognize and expect them to avoid such fallacy.

akafazov: Hi @artificial-intelligence ,

If I understand you correctly, it suggest to not list any disadvantages in Option 4, because this should be out of scope of the ADR. I guess we should decide if to remove option 4 altogether, or only list advantages. I initially didn't include it, because it is a normal thing to consider by the reader, what will happen if we don't have a SONiC strategy.

joshmue: > If we don't want to support anything then we certainly don't need a draft document as a standard for that, because the list of things we don't support is arbitrary long and we don't write down everything we don't support (like running SCS on Amiga OS).

I agree that a standard saying "SCS does not adopt X at all" could be applied to basically every other existing technology. Hence, this **standard** would be completely superfluous.

However, AFAIK, the concept of a **decision record** (as this document is declared as) means that this document is not a standard one could comply to, but a mere record of what decision SCS made regarding a technology. That way, if someone wonders in the future why SONIC is (or is not) used in SCS, they can look at a document outlining the reasoning (at the time) which was accepted by a certain quorum.

So while I do not think it is practicable to do that for every technology anyone ever mentions, a decision record with the result "SCS does not adopt SONiC at all" is fine.

akafazov: > > If we don't want to support anything then we certainly don't need a draft document as a standard for that, because the list of things we don't support is arbitrary long and we don't write down everything we don't support (like running SCS on Amiga OS).
> 
> I agree that a standard saying "SCS does not adopt X at all" could be applied to basically every other existing technology. Hence, this **standard** would be completely superfluous.
> 
> However, AFAIK, the concept of a **decision record** (as this document is declared as) means that this document is not a standard one could comply to, but a mere record of what decision SCS made regarding a technology. That way, if someone wonders in the future why SONIC is (or is not) used in SCS, they can look at a document outlining the reasoning (at the time) which was accepted by a certain quorum.
> 
> So while I do not think it is practicable to do that for every technology anyone ever mentions, a decision record with the result "SCS does not adopt SONiC at all" is fine.

Well, the decision is Option 3. This has been discussed and decided within the SCS community. I think the question we are wondering here is if we should include Option 4 (No SONIC support) in the ADR and list all advantages/disadvantages for this option.

horazont: > A list of (dis)advantages when evaluating products should contain (dis)advantages special to this product, not (dis)advantages regarding any arbitrary solution.

That sounds like a decent middle ground to me I can get behind.

joshmue: Ah, was not aware of that.

(But, if there is no distinct decision record dealing with the question of "**if**", leaving the "**how**" to this document, including the reasoning for "**if**" here would probably help the reader to understand the trade-offs.)

PR #444, opened 2024-01-10
Clarify wording about flavor requirements

### LINKED ISSUE #295: You may offer additional `SCS-` flavors ...
Could this be confusing when moving between SCS clouds in the future?
I think it is important to see which flavors are SCS standard an which aren't. What do you think?

https://github.com/SovereignCloudStack/standards/blob/4aad551ed3c009d38c4ef7575200e7466de5cac9/Standards/scs-0100-v2-flavor-naming.md?plain=1#LL286-L286

mbuechse:
Resolves #295.


fkr (approved):
LGTM.

PR #455, opened 2024-01-22
Streamline syntax of flavor-name tooling

### LINKED ISSUE #454: Streamline command-line syntax of flavor-name-check.py
It is a bit confusing that the script has to be invoked with `-2` if we want to check for anything after v1; see [this comment](https://github.com/SovereignCloudStack/standards/pull/443#discussion_r1457799246).

The problem is that we usually want a command-line syntax that is backwards compatible, i.e., the semantics of a certain commandline shouldn't change over time. My solution for this problem would be to expect an argument thatt fully determines the semantics.

mbuechse:
Resolves #454

Following issues are addressed:

- provide a cli tool with a streamlined and stable syntax
- get rid of the importlib hack
- get rid of excessive globals
- fix a few bugs (and probably introduce new ones)
- refactor to make more Pythonic (and hopefully more accessible)


mbuechse:
Errors found so far (in the original code that's currently in main):

Flavor name input didn't work when CPU brand was given, but generation wasn't.

Flavor name parser accepted `_gN-h`, didn't accept `_gNh` (treated `h` as generation).

mbuechse:
See the [README](https://github.com/SovereignCloudStack/standards/tree/issue/454/Tests/iaas/flavor-naming) for details about the streamlined command-line syntax.

Tests/iaas/flavor-naming/cli.py
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/env python3
```
martinmo, line 1:
```suggestion
#!/usr/bin/env python3
"""Command-line interface for validation and interactive construction of flavor names"""
```

Tests/iaas/flavor-naming/README.md
```diff
@@ -0,0 +1,174 @@
...
+  h: High Perf
+  hh: Very High Perf
+  hhh: Very Very High Perf
+ Performance: 
+Infiniband
+ ?IB: y
+SCS-16T-64-3x10s_bms_i3h_GNa-64_ib
+```
+
+Where the final line is the constructed flavor name, and everything above is the interactive session.
```
martinmo, line 174: Only `cli.py` is covered. The other tools in the directory should also be (briefly) mentioned and, if necessary, described here.

Tests/iaas/flavor-naming/README.md
```diff
@@ -0,0 +1,174 @@
...
+  which to base the syntax validation. With 'v1/v2', flavor names of both
+  kinds are accepted, but warnings are emitted for v2, and similarly with
+  'v2/v1', where warnings are emitted for v1.
+
+Options:
+  -o, --output [none|prose|yaml]  select output format (default: none)
+  --help                          Show this message and exit.
+```
+
+Here is an example invokation:
```
martinmo, line 47:
```suggestion
Here is an example invocation:
```

Tests/iaas/flavor-naming/flavor-form-server.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python3
```
martinmo, line 1:
```suggestion
#!/usr/bin/env python3
"""Simple webserver that can be used to test the flavor-form.py CGI script."""
```

Tests/iaas/flavor-naming/flavor-form-server.py
```diff
@@ -0,0 +1,83 @@
...
+import sys
+
+
+class RequestHandler(CGIHTTPRequestHandler):
+
+    def is_cgi(self):
+        collapsed_path = _url_collapse_path(self.path)
+        if not collapsed_path.startswith('/cgi-bin/'):
+            return False
+        self.cgi_info = '', collapsed_path[8:]
```
martinmo, line 18:
```suggestion
        self.cgi_info = '', collapsed_path.removeprefix('/cgi-bin')
```
(that's [new in Python 3.9](https://docs.python.org/3/library/stdtypes.html#str.removeprefix))

Tests/iaas/flavor-naming/flavor-form-server.py
```diff
@@ -0,0 +1,83 @@
...
+    ServerClass.address_family, addr = _get_best_family(bind, port)
+    HandlerClass.protocol_version = protocol
+    with ServerClass(addr, HandlerClass) as httpd:
+        host, port = httpd.socket.getsockname()[:2]
+        url_host = f'[{host}]' if ':' in host else host
+        print(
+            f"Serving HTTP on {host} port {port} "
+            f"(http://{url_host}:{port}/) ..."
+        )
+        try:
```
martinmo, line 39:
```suggestion
        print(f"The form is accessible at http://{url_host}:{port}/cgi-bin/flavor-form.py")
        try:
```

Tests/iaas/flavor-naming/flavor-names-openstack.py
```diff
@@ -13,16 +13,17 @@
 provided by openstack, such as the number of vCPUs and memory.
```
martinmo, line 13: The module level comment is now outdated:

* the script doesn't use `flavor-name-check.py` anymore
* https://github.com/SovereignCloudStack/Docs/Design-Docs/tools is a dead link (but that's not caused by this PR)

martinmo (changes requested):
I left some review comments inline. The overall impression is very good and the goals that were set are achieved, in my opinion.

As you pointed out to me via chat there seems to be a regression compared to the code in `main`, when checking with `python3 Tests/iaas/flavor-naming/check-yaml.py Tests/testing`. For the record, this is the output difference:

```diff
--- main.txt    2024-02-20 16:51:54.066472519 +0100
+++ 455.txt     2024-02-20 16:52:01.326498563 +0100
@@ -1,8 +1,7 @@
 ERROR: flavor 'SCS-1V-4': field 'cpu-type' contradicting name-v2 'SCS-1V-4'; found 'crowded-core', expected 'shared-core'
 ERROR: flavor 'SCS-2V-8': field 'name-v1' contradicting name-v2 'SCS-2V-8'; found 'SCS-2V-8', expected 'SCS-2V:8'
 ERROR: flavor 'SCS-4V-16': field 'ram' contradicting name-v2 'SCS-4V-16'; found 12, expected 16.0
-ERROR: flavor 'SCS-8V-32': field 'disk' contradicting name-v2 'SCS-8V-32'; found 128, expected undefined
 ERROR: flavor 'SCS-1V-2': field 'cpus' contradicting name-v2 'SCS-1V-2'; found 2, expected 1
 ERROR: flavor 'SCS-2V-4-20s': field 'disk0-type' contradicting name-v2 'SCS-2V-4-20s'; found 'network', expected 'ssd'
 ERROR: flavor 'SCS-4V-16-100s': field 'disk' contradicting name-v2 'SCS-4V-16-100s'; found 10, expected 100
-ERROR: file 'scs-0103-v1-flavors-wrong.yaml': found 7 errors
+ERROR: file 'scs-0103-v1-flavors-wrong.yaml': found 6 errors
```

mbuechse:
@martinmo Thanks for the flowers! I have tried to incorporate everything.

Tests/iaas/flavor-naming/flavor-form-server.py
```diff
@@ -0,0 +1,81 @@
...
+        except KeyboardInterrupt:
+            print("\nKeyboard interrupt received, exiting.")
+            sys.exit(0)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-b', '--bind', metavar='ADDRESS',
+                        help='bind to this address '
+                             '(default: all interfaces)')
```
martinmo, line 48: One tiny suggestion that I missed to include yesterday (sorry!):

```suggestion
    parser.add_argument('-b', '--bind', metavar='ADDRESS',
                        default='127.0.0.1',
                        help='bind to this address '
                             '(default: %(default)s)')
```

This will make the string printed above actually usable in the default case:

```
The form is accessible at http://127.0.0.1:8000/cgi-bin/flavor-form.py
```
vs.

```
The form is accessible at http://0.0.0.0:8000/cgi-bin/flavor-form.py
```

martinmo (approved):
LGTM now and I confirm the regression is fixed :rocket: 

Maybe you can also incorporate the [above suggestion ](https://github.com/SovereignCloudStack/standards/pull/455#discussion_r1497049473) I forgot to include in my first review.

Tests/iaas/flavor-naming/flavor-form-server.py
mbuechse: @martinmo The variant with 0.0.0.0 does indeed work (at least on my machine).

martinmo: @mbuechse ok, I checked again and now it works here, too – don't know what I did wrong the first time :shrug: but nevertheless, I would recommend using a local interface because it is the more secure default.

PR #464, opened 2024-02-06
Stabilize scs-0104-v1 Standard Images

mbuechse:
To be voted upon in Team IaaS.


cah-hbaum:
Not stabilized for now, since a few things came up in the "Team IaaS Meeting" (copy from the meeting notes):
- Requirement to be able to upload 40+ images with 200GB+ space (per project) is not currently tested in our conformance test: Add this later, open issue
- deferred

mbuechse:
Excerpt from SIG Std/Cert 2024-02-08:

- reasoning:
    - In the standard there is a requirement to be able to upload 40 images per 100 GB, Operators were not aware of this limit, currently there are no quotas in place.
    - When we allow customers to upload their custom images we do not have to support a multitude of images.
    - We want to propose that custom images have to be possible however there should be some fair use limitations, as this should not be abused.
- Customers can snapshot running VMs therefore the limit has to be so it is possible
- Two workflows:
    - Bring your own image
    - Snapshot running VM
- @fkr Proposal: Remove hard limits from the standard and mention just fair use policy
- Solution for Snapshoting will be discussed later

garloff (approved):
Looks good to me.
We have had the discussion in the Team IaaS that for setting a fixed minimal quota for custom images, it would be difficult to set the right number.
We thus took it out.
With that, I had the feeling that everyone in the meeting was in support. We did not record this with a formal vote, as this is something that we don't normally do (outside of maybe standardization decisions).
For me this is a clear Approval.

artificial-intelligence (approved):

fkr (approved):

PR #467, opened 2024-02-06
Discussion about obsoleted_at and deprecated_at

cah-hbaum:
SCS-0001 defines deprecation (and deprecated_at) as the keywords for deprecated or obsolete standards, that are most of the time succeeded by newer versions.
This MR replaces the obsoleted_at strings with deprecated_at defined by the initial standard.

### DIFF
--- a/Tests/chk_adrs.py
+++ b/Tests/chk_adrs.py
@@ -18,7 +18,7 @@ import yaml
 # | `type`          | REQUIRED                                                                   | one of `Procedural`, `Standard`, or `Decision Record`                                 |
 # | `status`        | REQUIRED                                                                   | one of `Proposal`, `Draft`, `Stable`, `Deprecated`, or `Rejected`                     |
 # | `track`         | REQUIRED                                                                   | one of `Global`, `IaaS`, `KaaS`, `IAM`, `Ops`                                         |
-# | `deprecated_at`  | REQUIRED if `status` is `Deprecated`                                       | ISO formatted date indicating the date after which the deprecation is in effect       |
+# | `deprecated_at` | REQUIRED if `status` is `Deprecated`                                       | ISO formatted date indicating the date after which the deprecation is in effect       |
 # | `stabilized_at` | REQUIRED if `status` was ever `Stable`                                     | ISO formatted date indicating the date after which the document was considered stable |
 # | `rejected_at`   | REQUIRED if `status` is `Rejected`                                         | ISO formatted date indicating the date on which the document was rejected             |
 # | `replaced_by`   | RECOMMENDED if `status` is `Deprecated` or `Rejected`, FORBIDDEN otherwise | List of documents which replace this document.                                        |
--- a/Tests/chk_adrs.py
+++ b/Tests/chk_adrs.py
@@ -18,7 +18,7 @@ import yaml
 # | `type`          | REQUIRED                                                                   | one of `Procedural`, `Standard`, or `Decision Record`                                 |
 # | `status`        | REQUIRED                                                                   | one of `Proposal`, `Draft`, `Stable`, `Deprecated`, or `Rejected`                     |
 # | `track`         | REQUIRED                                                                   | one of `Global`, `IaaS`, `KaaS`, `IAM`, `Ops`                                         |
-# | `obsoleted_at`  | REQUIRED if `status` is `Deprecated`                                       | ISO formatted date indicating the date after which the deprecation is in effect       |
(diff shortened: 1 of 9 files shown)


garloff:
Thanks for raising this.
The current status is that we use the `obsoleted_at` keyword to record the transition date from `stable` to `deprecated` consistently.
People might expect the target state `deprecated` (or `obsolete`) to match with the tag that describes the transition (`deprecated_at` or `obsoleted_at`).
So we have three options:
(1) Leave as is and live with the words not being what most people would expect
(2) Change the date-recording keyword from `obsoleted_at` to `deprecated_at` (as you suggest in this PR)
(3) Change the state `deprecated` to `obsolete`

I can not get myself to have a clear opinion.
I guess, I'd like to hear from @horazont as the original author whether he has a clear opinion or reason.
We could spend a minute on the next standardization SIG on this ...

cah-hbaum:
@garloff Thx for the response.
Some time in the next SIG would probably be fine with me and maybe I can try to catch @horazont in the office and ask him about this (don't know if he reads Github mails here). Then we can maybe decide on this.

mbuechse:
At the end of scs-0001-v1, we find a remark that this document was inspired by xep-0001, and in the latter, we find both states:

https://xmpp.org/extensions/xep-0001.html#states-Deprecated
https://xmpp.org/extensions/xep-0001.html#states-Obsolete

So maybe we really need to decide what we mean, and then use the appropriate word.

cah-hbaum:
In accordance with the links and the statements they're providing, both

mbuechse:
@cah-hbaum One minor correction: the problem doesn't lie with the test, but within the standard itself. The test only checks that the standard is being adhered to.

cah-hbaum:
I now added both `obsoleted_at` and `deprecated_at`, since both server a purpose according to the inspiree document.
I also did some cosmetic work (and grammar fixes) in a separate commit and updated the test (also in a separate commit), so they can be checked individually.

cah-hbaum:
Like we discussed in the SIG Standardization/Certification meeting, I made `deprecated_at` and the `Deprecation` state the used state and removed `Obsolete` and `obsoleted_at` for now.

Tests/chk_adrs.py
```diff
@@ -18,7 +18,8 @@
 # | `type`          | REQUIRED                                                                   | one of `Procedural`, `Standard`, or `Decision Record`                                 |
 # | `status`        | REQUIRED                                                                   | one of `Proposal`, `Draft`, `Stable`, `Deprecated`, or `Rejected`                     |
 # | `track`         | REQUIRED                                                                   | one of `Global`, `IaaS`, `KaaS`, `IAM`, `Ops`                                         |
-# | `obsoleted_at`  | REQUIRED if `status` is `Deprecated`                                       | ISO formatted date indicating the date after which the deprecation is in effect       |
+# | `deprecated_at` | REQUIRED if `status` is `Deprecated`                                       | ISO formatted date indicating the date after which the deprecation is in effect       |
+# | `obsoleted_at`  | REQUIRED if `status` is `Obsoleted`                                        | ISO formatted date indicating the date after which the obsoletion is in effect        |
```
mbuechse, line 22:
```suggestion
```

Tests/chk_adrs.py
```diff
@@ -95,8 +97,10 @@ def check_front_matter(self, fn, front):
         status = front.get("status")
         if "replaced_by" in front and status not in ("Deprecated", "Rejected"):
             self.emit(f"in {fn}: replaced_by is set, but status does not match")
-        if status == "Deprecated" and "obsoleted_at" not in front:
+        if status == "Deprecated" and "deprecated_at" not in front:
             self.emit(f"in {fn}: status is Deprecated, but deprecated_at date is missing")
+        if status == "Obsoleted" and "obsoleted_at" not in front:
+            self.emit(f"in {fn}: status is Obsoleted, but obsoleted_at date is missing")
```
mbuechse, line 103:
```suggestion
```

mbuechse (approved):
Two minor items. BUT please, please take out the commit that changes all the line breaks in scs-0001 into a separate PR, because we have multiple PRs open that work on this file, and merging would become way harder if we changed the line breaks first.

Tests/chk_adrs.py
cah-hbaum: Thx for the catch, I thought I fixed that (or better I'm pretty sure about it), but somehow the changes ended up in limbo.

Tests/chk_adrs.py
cah-hbaum: Same here unfortunately.

PR #478, opened 2024-02-21
Pin testsuite dependencies with pip-compile

### LINKED ISSUE #477: Use pip-compile to pin Python dependencies
At the moment, the conformance tests list their dependencies in `requirements.txt` without specific versions. While at the moment this isn't problematic, this will lead to problems in the future if incompatible new versions of the listed packages are published on PyPI.

I would like to propose a relatively lightweight solution to this problem using `pip-compile` from the [pip-tools package](https://pypi.org/project/pip-tools/).

The idea is to use the current `requirements.txt` as the `requirements.in` file (this is a pip-tools convention). It lists our "abtract" requirements. The "concrete" requirements are compiled with `pip-compile` into `requirements.txt`. It will contain all dependencies, including transitive dependencies, along with specific versions ("pinned versions") and can be used to reproduce the same Python environment over and over again.

Of course, from time to time, the dependencies listed in the concrete `requirements.txt` need to be bumped.

martinmo:
Resolves #477

"Abstract" requirements, previously listed in requirements.txt, are now listed in requirements.in and get compiled to a list of "concrete" requirements in requirements.txt using `pip-compile`.

Furthermore I documented the changes as seen from a testsuite user and testsuite developer perspective.

I settled with Python 3.10 because it is used in `Tests/Dockerfile`. However there seems to be no consensus which version is used throughout the project (e.g., in `.github/workflows/lint-python.yml`, `3.x` is specified). The reasoning behind picking a Python version and sticking to it for `pip-compile` is that concrete requirements can change depending on the Python release that is used.


martinmo:
Rebased on `main` to include the addition of the `click` package.

Tests/README.md
```diff
@@ -48,8 +56,9 @@ docker build --tag scs-compliance-check .
 docker run -it --env OS_CLOUD=CLOUDNAME -v ~/.config/openstack:/root/.config/openstack:ro scs-compliance-check
 ```
 
-The Docker entrypoint uses [scs-compatible.yaml](scs-compatible.yaml) on the `iaas` layer by default. You can use an alternative spec file
-by simply appending it to the above call, e.g.
+The Docker entrypoint uses [scs-compatible-iaas.yaml](scs-compatible-iaas.yaml)
```
mbuechse, line 59: I'm afraid that this isn't true any more (not your fault). Could you update the Dockerfile as well?

mbuechse: Never mind, I did it myself. But while testing the Docker build (which worked okay), I got this message: 
```console
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/
```
:(

mbuechse (approved):

mbuechse (approved):
I really like this change. Thank you very much!

Tests/README.md
martinmo: > I'm afraid that this isn't true any more (not your fault). Could you update the Dockerfile as well?

Good catch, didn't see that – thx!

> But while testing the Docker build (which worked okay), I got this message […]

This is a local issue and not related to the `Dockerfile`: you need to install the docker buildx plugin which will automatically used for `docker build`. On Debian/Ubuntu with the Docker APT repo just `apt install docker-buildx-plugin`.

Issue #481, opened 2024-02-23
[Feature Request] Implement tests for scs-0210-v2, scs-0214-v1


mbuechse:
This is a high-priority crash operation: find a way to implement reasonable tests by 2024-02-28.

- scs-0210-v1 does have tests; what needs to be done for v2?
- scs-0214-v1 might be hard to test; what's the "MVP" that can be achieved within this timeframe?

Request assistance from whatever parties necessary!

cah-hbaum:
We integrated sonobuoy yesterday, but for the `scs-0210-v2` we should probably still use and update the old python test from v1. The problem here is, its not really useful to update the test, since `v2` adds a section talking about the support period for a K8s version. Testwise, this is contradicting to the stuff tested right now, since `v1` checks a given cluster and if its version is up-to-date with the release timeline. `v2` demands that old versions are still supported and available from the provider, which isn't really something you could test (or at least I don't know how).

`scs-0214-v1` is as @mbuechse said a bit more tricky. If it is doable, a test could be written in Go with Sonobuoy in mind. If there isn't time for that, just doing a Python test would be fine and we can convert that later.
The main thing to test here is the requirement for control plane nodes to be distributed over multiple nodes. The standard even calls for physical nodes, but that could be kinda hard to do.
The best thing to do would be to check the IPs of the control plane (and maybe worker nodes) and look for a difference there. This would at least guarantee multiple VMs used. Now to check if they're separated over different machines or even failure zones is harder and probably dependent on the infrastructure of the provider. If there would be access to OpenStack as well (or another IaaS solution) we could check that with something like Failure Zones, but this is maybe too much for now (?). Would be dependent on the available time overall.
Now if these failure zones or zones in general are strictly separated over e.g. multiple racks or buildings, we can't really check, since this is something the provider would know and the metadata necessary for that is probably not publicly available or can't be accessed for some reason or another.

martinmo:
For conformance testing of `scs-0210-v2` ("K8S Version Policy") I have the following proposal.

The check consists of two stages:

1. Getting the list of currently supported K8s versions for a given KaaS provider (the new and *maybe* tricky part). It can optionally be saved into a YAML file.
2. Deciding if this list, which can optionally be read from a YAML file, conforms to `scs-0210-v2`. For this, a lot of the the existing v1 test implementation can be reused, because the basic predicate "Is k8s version X.Y.Z recent enough?" of course still needs to be decided.

Stage 1 has multiple strategies:

[...]

PR #486, opened 2024-02-27
Create conformance test for scs-0214-v1

cah-hbaum:
As discussed in https://github.com/SovereignCloudStack/standards/issues/481, we need to create a conformance test for SCS-0214-v1. This PR does that and also includes a separate commit containing the necessary changes for the `Conformance Tests` section of the standard. If we don't want to include this, the commit will be split up into a separate PR.

Resolves https://github.com/SovereignCloudStack/issues/issues/477

### DIFF
--- a/Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
+++ b/Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
@@ -1,0 +1,252 @@
+#!/usr/bin/env python3
+# vim: set ts=4 sw=4 et:
+#
+"""
+K8s Node Distribution Check
+https://github.com/SovereignCloudStack/standards
+
+Return codes:
+0: Node distribution according to standard
+1: Help command used
+2: Node distribution either not according to standard or not detected due to missing labels
+
+One or more K8s clusters are checked by providing their kubeconfigs.
+scs-0214-v1 requires a K8s cluster to be distributed over multiple
+physical machines within different failure zones of an infrastructure.
+For now, this is hard to test, since we would need to know a multitude
+of things about the providers infrastructure, including things like
+circuitry and physical distribution of different machines.
+This test is for now only designed to work on the K8s layer, which is
+done by checking several well-known labels (https://kubernetes.io/docs/reference/labels-annotations-taints/)
+and comparing them. This doesn't always attest to a good distribution
+and does require these labels to be set, but should yield overall pretty
+good initial results.
+
+    kubernetes.io/hostname
+    topology.kubernetes.io/zone
+    topology.kubernetes.io/region
+    node-role.kubernetes.io/control-plane
+
+(c) Hannes Baum <hannes.baum@cloudandheat.com>, 6/2023
+License: CC-BY-SA 4.0
+"""
+
+import asyncio
+import getopt
+import kubernetes_asyncio
+import logging
+import logging.config
+import sys
+import yaml
+
+
+logging_config = {
+    "level": "INFO",
+    "version": 1,
+    "disable_existing_loggers": False,
+    "formatters": {
+        "k8s-node-distribution-check": {
+            "format": "%(levelname)s: %(message)s"
+        }
+    },
+    "handlers": {
+        "console": {
+            "class": "logging.StreamHandler",
+            "formatter": "k8s-node-distribution-check",
+            "stream": "ext://sys.stdout"
+        }
+    },
+    "root": {
+        "handlers": ["console"]
+    }
+}
+
+logger = logging.getLogger(__name__)
+
+
+class ConfigException(BaseException):
+    """Exception raised if a configuration error occurs"""
+
+
+class HelpException(BaseException):
+    """Exception raised if the help functionality is called"""
+
+
+class DistributionException(BaseException):
+    """Exception raised if the distribution seems to be not enough"""
+
+
+class Config:
+    config_path = "./config.yaml"
+    kubeconfig = None
+    logging = None
+
+
+def print_usage():
+    print("""
+K8s Node Distribution Compliance Check
+
+Usage: k8s-node-distribution-check.py [-h] [-c|--config PATH/TO/CONFIG] -k|--kubeconfig PATH/TO/KUBECONFIG
+
+The script is used to determine the compliance with the SCS standard SCS-0214-v1 describing the distribution
+and availability of K8s nodes.
+
+The following return values are possible:
+    0 - Distribution according to the standard could be verified.
+    1 - The help output was delivered.
+    2 - No distribution according to the standard could be detected for the nodes available.
+
+The following arguments can be set:
+    -c/--config PATH/TO/CONFIG         - Path to the config file of the test script
+    -k/--kubeconfig PATH/TO/KUBECONFIG - Path to the kubeconfig of the server we want to check
+    -h                                 - Output help
+""")
+
+
+def parse_arguments(argv):
+    """Parse cli arguments from the script call"""
+    config = Config()
+
+    try:
+        opts, args = getopt.gnu_getopt(argv, "c:k:h", ["config", "kubeconfig", "help"])
+    except getopt.GetoptError:
+        raise ConfigException
+
+    for opt in opts:
+        if opt[0] == "-h" or opt[0] == "--help":
+            raise HelpException
+        if opt[0] == "-c" or opt[0] == "--config":
+            config.config_path = opt[1]
+        if opt[0] == "-k" or opt[0] == "--kubeconfig":
+            config.kubeconfig = opt[1]
+
+    return config
+
+
+def setup_logging(config_log):
+
+    logging.config.dictConfig(config_log)
+    loggers = [
+        logging.getLogger(name)
+        for name in logging.root.manager.loggerDict
+        if not logging.getLogger(name).level
+    ]
+
+    for log in loggers:
+        log.setLevel(config_log['level'])
+
+
+def initialize_config(config):
+    """Initialize the configuration for the test script"""
+
+    try:
+        with open(config.config_path, "r") as f:
+            config.logging = yaml.safe_load(f)['logging']
+    except OSError:
+        logger.warning(f"The config file under {config.config_path} couldn't be found, "
+                       f"falling back to the default config.")
+    finally:
+        # Setup logging if the config file with the relevant information could be loaded before
+        # Otherwise, we initialize logging with the included literal
+        setup_logging(config.logging or logging_config)
+
+    if config.kubeconfig is None:
+        raise ConfigException("A kubeconfig needs to be set in order to test a k8s cluster version.")
+
+    return config
+
+
+async def get_k8s_cluster_labelled_nodes(kubeconfig, interesting_labels):
+    """Get the labels per node of the k8s cluster under test, ignoring labels we don't need."""
+    await kubernetes_asyncio.config.load_kube_config(kubeconfig)
+
+    nodes = list()
+
+    async with kubernetes_asyncio.client.ApiClient() as api:
+        core_api = kubernetes_asyncio.client.CoreV1Api(api)
+        result = await core_api.list_node()
+        for node in result.items:
+            filtered_labels = {
+                label: value
+                for label, value in node.metadata.labels.items()
+                if label in interesting_labels
+            }
+            nodes.append(filtered_labels)
+
+    return nodes
+
+
+def compare_labels(node_list, labels, node_type="master"):
+
+    label_data = {key: list() for key in labels}
+
+    for node in node_list:
+        for key in labels:
+            try:
+                label_data[key].append(node[key])
+            except KeyError:
+                logger.warning(f"The label for {key.split('/')[1]}s don't seem to be set for all nodes.")
+
+    for label in labels:
+        if len(label_data[label]) < len(node_list):
+            logger.warning(f"The label for {label.split('/')[1]}s doesn't seem to be set for all nodes.")
+        if len(set(label_data[label])) <= 1:
+            logger.warning(f"There seems to be no distribution across multiple {label.split('/')[1]}s "
+                           "or labels aren't set correctly across nodes.")
+        else:
+            logger.info(f"The nodes are distributed across {str(len(set(label_data[label])))} {label.split('/')[1]}s.")
+            return
+
+    if node_type == "master":
+        raise DistributionException("The distribution of nodes described in the standard couldn't be detected.")
+    elif node_type == "worker":
+        logger.warning("No node distribution could be detected for the worker nodes. "
+                       "This produces only a warning, since it is just a recommendation.")
+        return
+
+
+async def main(argv):
+    try:
+        config = initialize_config(parse_arguments(argv))
+    except (OSError, ConfigException, HelpException) as e:
+        if hasattr(e, 'message'):
+            logger.error(e.message)
+        print_usage()
+        return 1
+
+    # It is important to note, that the order of these labels matters for this test.
+    # Since we want to check if nodes are distributed, we want to do this from bigger
+    # infrastructure parts to smaller ones. So we first look if nodes are distributed
+    # across regions, then zones and then hosts. If one of these requirements is fulfilled,
+    # we don't need to check anymore, since a distribution was already detected.
+    labels = (
+        "topology.kubernetes.io/region",
+        "topology.kubernetes.io/zone",
+        "kubernetes.io/hostname",
+    )
+
+    nodes = await get_k8s_cluster_labelled_nodes(config.kubeconfig, labels + ("node-role.kubernetes.io/control-plane", ))
+
+    if len(nodes) < 2:
+        logger.error("The tested cluster only contains a single node, which can't comply with the standard.")
+        return 2
+
+    labelled_master_nodes = [node for node in nodes if "node-role.kubernetes.io/control-plane" in node]
+    try:
+        if len(labelled_master_nodes) >= 1:
+            worker_nodes = [node for node in nodes if "node-role.kubernetes.io/control-plane" not in node]
+            # Compare the labels of both types, since we have enough of them with labels
+            compare_labels(labelled_master_nodes, labels, "master")
+            compare_labels(worker_nodes, labels, "worker")
+        else:
+            compare_labels(nodes, labels)
+    except DistributionException as e:
+        logger.error(str(e))
+        return 2
+
+    return 0
+
+
+if __name__ == "__main__":
+    return_code = asyncio.run(main(sys.argv[1:]))
+    sys.exit(return_code)
--- a/Standards/scs-0214-v1-k8s-node-distribution.md
+++ b/Standards/scs-0214-v1-k8s-node-distribution.md
@@ -67,7 +67,7 @@ This standard formulates the requirement for the distribution of Kubernetes node
 to provide a fault-tolerant and available Kubernetes cluster infrastructure.
 Since some providers only have small environments to work with and therefore couldn't
 comply with this standard, it will be treated as a RECOMMENDED standard, where providers
-can OPT-OUT.
+can OPT OUT.
 
 If the standard is used by a provider, the following decisions are binding and valid:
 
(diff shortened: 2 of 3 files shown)


Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -0,0 +1,252 @@
...
+    for label in labels:
+        if len(labels[label]) <= 1:
+            if len(labels[label]) == 1:
+                logger.warning(f"There seems to be no distribution across multiple {label.split('/')[1]}s "
+                               "or labels aren't set correctly across nodes.")
+            elif len(labels[label]) == 0:
+                logger.warning(f"The label for {label.split('/')[1]}s don't seem to be set for all nodes.")
+        else:
+            logger.info(f"The nodes are distributed across {str(len(labels[label]))} {label.split('/')[1]}s.")
+            return
```
mbuechse, line 209: The `return` here comes as a little surprise to me. Are we happy if just one of the labels has more than 1 value? Maybe this can be documented a bit (preferably in the standard itself, but at least make a little remark here).

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -0,0 +1,252 @@
...
+
+    nodes = await get_k8s_cluster_labelled_nodes(config.kubeconfig)
+
+    if len(nodes) < 2:
+        logger.error("The tested cluster only contains a single node, which can't comply with the standard.")
+        return 2
+
+    labelled_master_nodes = [node for node in nodes if "node-role.kubernetes.io/control-plane" in node]
+    try:
+        if len(labelled_master_nodes) > 1 and len(nodes) > 2:
```
mbuechse, line 236: shouldn't this be `>= 1` and `=> 2`?

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -0,0 +1,252 @@
...
+            worker_nodes = [node for node in nodes if "node-role.kubernetes.io/control-plane" not in node]
+            # Compare the labels of both types, since we have enough of them with labels
+            compare_labels(labelled_master_nodes, "master")
+            compare_labels(worker_nodes, "worker")
+        else:
+            compare_labels(nodes)
+    except DistributionException:
+        logger.error("The distribution of nodes described in the standard couldn't be detected.")
```
mbuechse, line 244: Preferably pass this string literal as argument to the exception upon throwing it; here just use `str(exc)` or something similar.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -0,0 +1,252 @@
...
+        for key in labels:
+            labels[key].add(node[key])
+
+    for label in labels:
+        if len(labels[label]) <= 1:
+            if len(labels[label]) == 1:
+                logger.warning(f"There seems to be no distribution across multiple {label.split('/')[1]}s "
+                               "or labels aren't set correctly across nodes.")
+            elif len(labels[label]) == 0:
+                logger.warning(f"The label for {label.split('/')[1]}s don't seem to be set for all nodes.")
```
mbuechse, line 206: Should we produce this warning also when there are more than 1 nodes with the label, but fewer than total nodes?

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -0,0 +1,252 @@
...
+import asyncio
+import getopt
+import kubernetes_asyncio
+import logging
+import logging.config
+import sys
+import yaml
+
+
+logging_config = {
```
mbuechse, line 43: This seems kinda redundant given that the same info is in the yaml file?

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -0,0 +1,252 @@
...
+The following return values are possible:
+    0 - Distribution according to the standard could be verified.
+    1 - The help output was delivered.
+    2 - No distribution according to the standard could be detected for the nodes available.
+
+The following arguments can be set:
+    -c/--config PATH/TO/CONFIG         - Path to the config file of the test script
+    -k/--kubeconfig PATH/TO/KUBECONFIG - Path to the kubeconfig of the server we want to check
+    -h                                 - Output help
+    """)
```
mbuechse, line 103:
```suggestion
""")
```

mbuechse (commented):
A few things that I think should be addressed. If you are time-constrained, I offer to assist.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -0,0 +1,252 @@
...
+    return nodes
+
+
+def compare_labels(node_list, node_type="master"):
+
+    labels = {
+        "topology.kubernetes.io/region": set(),
+        "topology.kubernetes.io/zone": set(),
+        "kubernetes.io/hostname": set(),
+    }
```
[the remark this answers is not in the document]
cah-hbaum: Yes its intentional, we don't need it here.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
cah-hbaum: Will document it.

Just to clarify it here as well, we're fine with one of these labels.
Basically, if we see that nodes are in different regions, it should (if the labeling is done correctly) not matter, if the (availability) zone or hostname is the same, since the same zone name or hostname could be used in different regions. Same thing goes for availability zones and hostnames.

If we don't want/can't rely on that, what would our way to go be? Throwing an error would probably invalidate quite a few clusters that are configured correctly (and some that are just labelled wrong, but that could always be the case).

I think being a bit more generous instead of too strict (also with regard to our time limit on this) is the better way.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
cah-hbaum: No, since if there is only 1 master node, we can't compare it with another master node. And we can also only do this, if we have more than 2 nodes in general, since this would mean we have at least one worker node.

I could argue about `> 2` being `>= 2`, since there could be a cluster without worker nodes and we would do these edge cases in different `if else` paths right now.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
cah-hbaum: Understandable, done in the fixup.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
cah-hbaum: Yes seems like a good idea.
I will include that in the fixup.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
cah-hbaum: This is there if a `yaml` file isn't provided.

mbuechse: I know. Then it should just be provided (and the fallback can be way more minimal)

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
mbuechse: My understanding was that once there is at least one occurrence of "node-role.kubernetes.io/control-plane", then we know that these labels are set, and we can distinguish between master nodes and worker nodes; otherwise we just assume that every node is a master node. Can a cluster with only one master node pass the test at all?

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
mbuechse: If I understand your answer correctly, we should check the labels in a certain order, right? In this case, region first, then zone, then hostname.

mbuechse: This is not clear from the code. Yes, that is the order in which they are named when the dict `labels` is constructed, but you rely on an implementation detail, namely that this order is preserved when iterating over the dict. You have to document this assumption, or better yet, iterate over a tuple. At any rate, it should also be document why the order is important.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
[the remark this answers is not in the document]
mbuechse: I understood this now; the control-plane label is only for distinguishing node types. Good to know.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
cah-hbaum: > Can a cluster with only one master node pass the test at all?
I would personally say no, but not because the setup is wrong, but because such a cluster wouldn't be constructed for availability or long-lasting services, but its probably there for development, testing, etc.
So its probably not something you would test for that.

mbuechse: My question was meant with respect to the standard, not your opinion ;)

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
cah-hbaum: I will document it in the text as well (and the tuple is coming)

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
mbuechse: If you change the condition to `>= 1`, then such a setup would fail the test (because `compare_labels` wouldn't reach the return statement, but the raise statement), and if the standard agrees with that result, then I don't see why we need `> 1`.

cah-hbaum: The standard says that multiple control plane nodes should be distributed. So it would pass.

mbuechse: Got it.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -0,0 +1,252 @@
...
+    """Get the labels of the k8s cluster under test."""
+    _ = await kubernetes_asyncio.config.load_kube_config(kubeconfig)
+
+    nodes = list()
+    labels = [
+        "kubernetes.io/hostname",
+        "topology.kubernetes.io/zone",
+        "topology.kubernetes.io/region",
+        "node-role.kubernetes.io/control-plane"
+    ]
```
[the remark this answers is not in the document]
cah-hbaum: Added a tuple and put it into both functions as a parameter.

cah-hbaum:
Added a fixup containing the mentioned changes. I hope I didn't forget anything @mbuechse

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -0,0 +1,258 @@
...
+    # infrastructure parts to smaller ones. So we first look if nodes are distributed
+    # across regions, then zones and then hosts. If one of these requirements is fulfilled,
+    # we probably don't need to check anymore, since a distribution was already detected.
+    labels = (
+        "topology.kubernetes.io/region",
+        "topology.kubernetes.io/zone",
+        "kubernetes.io/hostname",
+    )
+
+    nodes = await get_k8s_cluster_labelled_nodes(config.kubeconfig, labels + tuple("node-role.kubernetes.io/control-plane"))
```
mbuechse, line 234:
```suggestion
    nodes = await get_k8s_cluster_labelled_nodes(config.kubeconfig, labels + ("node-role.kubernetes.io/control-plane", ))
```

Your version does strange things:

```console
$ ipython3
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 7.31.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: tuple("node-role.kubernetes.io/control-plane")
Out[1]: 
('n',
 'o',
 'd',
 'e',
 '-',
 'r',
 'o',
 'l',
 'e',
 '.',
 'k',
 'u',
 'b',
 'e',
 'r',
 'n',
 'e',
 't',
 'e',
 's',
 '.',
 'i',
 'o',
 '/',
 'c',
 'o',
 'n',
 't',
 'r',
 'o',
 'l',
 '-',
 'p',
 'l',
 'a',
 'n',
 'e')
```

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -0,0 +1,258 @@
...
+                    # Node is not labelled as a control plane node
+                    pass
+            nodes.append(node)
+
+    return nodes
+
+
+def compare_labels(node_list, labels, node_type="master"):
+
+    label_data = dict()
```
mbuechse, line 184:
```suggestion
    label_data = {key: set() for key in labels}
```

Then lines 188, 189 can be removed.

cah-hbaum: Updated.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -0,0 +1,258 @@
...
+
+    for node in node_list:
+        for key in labels:
+            if key not in label_data:
+                label_data[key] = set()
+            try:
+                label_data[key].add(node[key])
+            except KeyError:
+                logger.warning(f"The label for {labels[key].split('/')[1]}s don't seem to be set for all nodes.")
```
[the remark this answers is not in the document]
cah-hbaum: Updated.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
cah-hbaum: Interesting, thank you.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -0,0 +1,258 @@
...
+    for label in labels:
+        if len(labels[label]) <= 1:
+            if len(labels[label]) == 1:
+                logger.warning(f"There seems to be no distribution across multiple {label.split('/')[1]}s "
+                               "or labels aren't set correctly across nodes.")
+            elif len(labels[label]) == 0:
+                logger.warning(f"The label for {label.split('/')[1]}s don't seem to be set for all nodes.")
+        else:
+            logger.info(f"The nodes are distributed across {str(len(labels[label]))} {label.split('/')[1]}s.")
+            return
```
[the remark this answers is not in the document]
cah-hbaum: Yes the `labels` needed to be replaced with `label_data`, already did that now.

I'm just confused about the second part, which comment about these lines? Or did these already switch line?

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
mbuechse: This comment here. You are right, the line numbers have changed, it doesn't get shown any more in the files view.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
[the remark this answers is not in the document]
mbuechse: I replied to the comment in question.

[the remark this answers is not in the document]
mbuechse: Ah, you're right, of course! We could use a `collections.Counter` instead of a set; this should cover both use cases. But I think we can postpone this. I don't know how important this warning is anyway.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
cah-hbaum: Its already more minimal than the config, but maybe it can be minimized more(?) Not sure though.
I fixed some small artifacts I found in the config.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -0,0 +1,255 @@
...
+        # Otherwise, we initialize logging with the included literal
+        setup_logging(config.logging or logging_config)
+
+    if config.kubeconfig is None:
+        raise ConfigException("A kubeconfig needs to be set in order to test a k8s cluster version.")
+
+    return config
+
+
+async def get_k8s_cluster_labelled_nodes(kubeconfig, labels):
```
martinmo, line 159: To make the intent clearer that the second argument is used for filtering, it should have name such as `interesting_labels`.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -0,0 +1,255 @@
...
+        for item in result.items:
+            node = dict()
+            node_labels = item.metadata.labels
+            for label in labels:
+                try:
+                    node[label] = node_labels[label]
+                except KeyError:
+                    # Node is not labelled as a control plane node
+                    pass
+            nodes.append(node)
```
martinmo, line 177:
```suggestion
        for node in result.items:
            filtered_labels = {
                label: value
                for label, value in node.metadata.labels.items()
                if label in interesting_labels
            }
            nodes.append(filtered_labels)
```

I think this is easier to read. Also, that comment was misleading ("Node is not labelled as a control plane node").

martinmo:
> Additionally, with manual testing I could only verify that the following cases have the expected conformance test outcome

I forgot to mention that, to me, the testing logic seems to be ok. Still, it would be good to have an easy way to verify the different scenarios, without having access to different kinds of clusters.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -0,0 +1,255 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
cah-hbaum: Youre right, I will correct that.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -0,0 +1,255 @@
...
+        if hasattr(e, 'message'):
+            logger.error(e.message)
+        print_usage()
+        return 1
+
+    # It is important to note, that the order of these labels matters for this test.
+    # Since we want to check if nodes are distributed, we want to do this from bigger
+    # infrastructure parts to smaller ones. So we first look if nodes are distributed
+    # across regions, then zones and then hosts. If one of these requirements is fulfilled,
+    # we probably don't need to check anymore, since a distribution was already detected.
```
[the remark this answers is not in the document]
cah-hbaum: I wont do a Star Wars joke here and instead just remove that.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
cah-hbaum: Sure why not.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
cah-hbaum: Yeah the comment was ooolllddd.

And I'm not that "pythonic", thats why I never come up with that, but thank you for the suggestion.

cah-hbaum:
> > Additionally, with manual testing I could only verify that the following cases have the expected conformance test outcome
> 
> I forgot to mention that, to me, the testing logic seems to be ok. Still, it would be good to have an easy way to verify the different scenarios, without having access to different kinds of clusters.

I agree with you on this, we can probably look into this later on. (Sadly not until tomorrow).

I also fixed/updated/changed most of the things you mentioned in the other comments, so you can recheck tomorrow =)

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
mbuechse: Let's come back to this point later. It's not mission critical.

mbuechse (approved):

martinmo (approved):
LGTM & we agreed via chat to add tests for the tests later.

PR #488, opened 2024-02-28
Add conformance tests for scs-0210-v2 version policy

martinmo:
This adds the missing conformance tests for scs-0210-v2. It has still some rough edges and some of the code doesn't look pretty, but I think it is good enough to be used.

One thing that needs to be maintained from now on, until we find a better solution, is the new `k8s-eol-data.yml` file which contains the EOL data of the K8s release branches.

To test it, provide a kubeconfig that has the contexts `stable`, `oldstable` and `oldoldstable`. You can use `minikube` to create clusters of different versions, e.g., `minikube create -p stable --kubernetes-version=...` and provide `-k ~/.kube/config` as a script argument.


mbuechse:
I will have a look. At first glance, it looks really good!

Tests/kaas/k8s-version-policy/k8s_version_policy.py
```diff
@@ -147,127 +147,88 @@ def initialize_config(config):
...
+        """Return a tuple of only the major and minor version."""
+        return (self.major, self.minor)
 
     def __str__(self):
         return f"{self.major}.{self.minor}.{self.patch}"
 
 
-class CVEVersionInfo:
-    """Class that contains a CVE version info.
+def parse_version(version_str: str) -> K8sVersion:
```
mbuechse, line 164: I like this very much. I'm wondering though why you opted against a classmethod.

mbuechse (commented):
Again, it looks really good. Does it test though that the three context actually use different K8s versions? I missed that.

Tests/kaas/k8s-version-policy/k8s_version_policy.py
martinmo: Personal preference, there is no *real* technical reason to do so :innocent: I prefer it because it is shorter to write without the class prefix and you would either need to drop the type hints or add `from __future__ import annotations` to make the interpreter happy about the reference of `K8sVersion` return type before the class definition is finished.

martinmo:
> Again, it looks really good. Does it test though that the three contexts actually use different K8s versions? I missed that.

Yes, it does so by collecting the used "branches" in a set and checking them afterwards. (Release) Branch means tuple(major, minor) in K8s context. Remark: I used the word "branches" here for two reasons: the [K8s release docs](https://kubernetes.io/releases/patch-releases/#detailed-release-history-for-active-branches) use it and there is no confusion with the minor version.

martinmo:
I've finished the last few missing bits. The PR description is updated and the draft status removed. Please review. If someone other than me merges this, feel free to clean up the squash merge message.

cah-hbaum (approved):
LGTM!
Tested at least with the "stable" part of the cluster and that worked well.

Tests/kaas/k8s-version-policy/k8s_version_policy.py
```diff
@@ -0,0 +1,531 @@
...
+Usage: k8s_version_policy.py [-h] [-c|--config PATH/TO/CONFIG] -k|--kubeconfig PATH/TO/KUBECONFIG
+
+The K8s version policy check returns 0 if the versions of the tested clusters are still acceptable, otherwise
+it returns 2 for an out-of date version or 3 if the used version should be updated due to a highly critical CVE.
+It returns 4 if a supported upstream K8s release is missing.
+
+    -c/--config PATH/TO/CONFIG         - Path to the config file of the test script
+    -k/--kubeconfig PATH/TO/KUBECONFIG - Path to the kubeconfig of the server we want to check
+    -h                                 - Output help
+    """)
```
mbuechse, line 105:
```suggestion
""")
```

Tests/kaas/k8s-version-policy/k8s_version_policy.py
```diff
@@ -0,0 +1,531 @@
...
+            # FIXME: this is ugly
+            return self
+        return K8sBranch(self.major, self.minor - 1)
+
+    def __str__(self):
+        return f"{self.major}.{self.minor}"
+
+
+@dataclass(frozen=True)
+class K8sBranchInfo:
```
mbuechse, line 203: I'm wondering if this class is really necessary, or if the map from K8sBranch to datetime would suffice (the methods `is_supported` and `is_eol` could be inlined, or a subclass of datetime could be used).

Tests/kaas/k8s-version-policy/k8s_version_policy.py
```diff
@@ -0,0 +1,531 @@
...
+                cluster.version,
+                cluster.name,
+            )
+        else:
+            logger.error(
+                "The K8s cluster version %s of cluster '%s' is outdated according to the standard.",
+                cluster.version,
+                cluster.name,
+            )
+            return 2
```
mbuechse, line 491: It would be preferable to collect the problems for all contexts. Then the return code could be computed as the maximum, or we do away with these return codes and we just return the number of errors (like with the entropy check, for instance).

mbuechse (approved):
As I said, I like it very much. A few remarks here and there, which could be converted into an issue for later. Most importantly: try to find as many problems with the clusters as possible before returning to the caller.

mbuechse:
@martinmo And one question that just occurred to me: are these hard-coded context names practicable? Or could we allow for a more sophisticated scheme? If we knew that three versions are officially supported (which is not always the case), then the command-line argument could look like the following: `-k kubeconfig.yaml:stable -k kubeconfig.yaml:oldstable -k kubeconfig.yaml:oldoldstable`. What do you think about that?

mbuechse:
Please fix this bug that was taken over from the old program: https://github.com/SovereignCloudStack/standards/issues/492#issuecomment-1976125074

Tests/kaas/k8s-version-policy/k8s_version_policy.py
```diff
@@ -0,0 +1,531 @@
...
+    -h                                 - Output help
+    """)
+
+
+def parse_arguments(argv):
+    """Parse cli arguments from the script call"""
+    config = Config()
+
+    try:
+        opts, args = getopt.gnu_getopt(argv, "c:k:h", ["config", "kubeconfig", "help"])
```
cah-hbaum, line 113:
```suggestion
        opts, args = getopt.gnu_getopt(argv, "c:k:h", ["config=", "kubeconfig=", "help"])
```

Tests/kaas/k8s-version-policy/k8s_version_policy.py
martinmo: I agree :+1: working on it.

martinmo:
> @martinmo And one question that just occurred to me: are these hard-coded context names practicable? Or could we allow for a more sophisticated scheme? If we knew that three versions are officially supported (which is not always the case), then the command-line argument could look like the following: `-k kubeconfig.yaml:stable -k kubeconfig.yaml:oldstable -k kubeconfig.yaml:oldoldstable`. What do you think about that?

I like your idea. We then could simplify some of the code, especially I could remove that "skipped because next context references a cluster that is already EOL" step.

Do you intend to change the script call in Tests/scs-compatible-kaas.yaml on a regular basis, e.g., when there is a switch from 4 to 3 simultanous supported K8s upstream branches (or vice versa)?

Tests/kaas/k8s-version-policy/k8s_version_policy.py
martinmo: I would like to keep the class, because IMHO it is more readable that way :)

mbuechse:
> Do you intend to change the script call in `Tests/scs-compatible-kaas.yaml` on a regular basis, e.g., when there is a switch from 4 to 3 simultanous supported K8s upstream branches (or vice versa)?

@martinmo I think these specifics are best captured using a variable (just like the kubeconfig itself). We have one Zuul job per subject, and I would set the variable there (as a Zuul/Ansible variable that can then be passed on to the main check script, which in turn it passes it on to the individual script).

We shouldn't forget though that in the future we are supposed to create the clusters ourselves. So this could already be fashioned with plugins in mind. So the actual syntax could be `--cluster kubeconfig:kubeconfig.yaml:v1.29`, where `kubeconfig` is fixed, and later we could add new methods such as `--cluster gardener:XYZ` or `--cluster CAPI:xyz`. Not quite sure.

mbuechse:
@martinmo One more thing: Should we just ignore any cluster whose branch is no longer supported? Maybe this could facilitate the transition periods. Otherwise people have to change the invocation of the script on the day of EOL.

Tests/kaas/k8s-version-policy/k8s_version_policy.py
mbuechse: It's definitely very readable, but there are lots of these classes, and this one seems a bit redundant (even more so with the map, where the key is entirely contained in the value). But I'm not too passionate about this. It's a matter of taste.

martinmo:
> @martinmo One more thing: Should we just ignore any cluster whose branch is no longer supported? Maybe this could facilitate the transition periods. Otherwise people have to change the invocation of the script on the day of EOL.

This is already done. One important caveat though is that the contexts need to be checked from newest to oldest for this detection to work properly.

Tests/kaas/k8s-version-policy/k8s-eol-data.yml
```diff
@@ -0,0 +1,12 @@
...
+- branch: '1.29'
+  end-of-life: '2025-02-28'
+- branch: '1.28'
+  end-of-life: '2024-10-28'
+- branch: '1.27'
+  end-of-life: '2024-06-28'
+- branch: '1.26'
+  end-of-life: '2024-02-28'
+- branch: '1.25'
+  end-of-life: '2023-10-28'
```
martinmo, line 12: One thing that bothers me: this file needs to be maintained from now on. I couldn't find a machine readable version upstream, just the [Patch releases document](https://kubernetes.io/releases/patch-releases/#detailed-release-history-for-active-branches).

mbuechse:
@martinmo Yes, we could also use an URL syntax. I deliberately put v1.29 there instead of stable to indicate that providers could name their stuff whatever they like, and the meaning of `stable` would seem to me to drift over time, which I personally wouldn't like too much, but evyerone's mileage may vary.

Tests/kaas/k8s-version-policy/k8s-eol-data.yml
mbuechse: We could assume that this rhythm of Oct 28, Feb 28, June 28 goes on, and then we only need to change anything if this rhythm breaks. Then of course we also need to know when these versions come out initially.

martinmo: Okay. With this scheme, another assumption that can break is that only the minor branch increments. I would like to take that idea into a separate issue.

For this PR, what do you think of issuing a warning if the file seems to be out of date at the time of script execution. This is the case if the file lists less than 3 non-EOL releases. And bail out if there are less than 2. With that approach we would at least see it if we monitor the logs in the CI.

martinmo: > issuing a warning if the file seems to be out of date at the time of script execution. This is the case if the file lists less than 3 non-EOL releases. And bail out if there are less than 2.

I've implemented this in the last commit. If you do not agree with that change, I can easily revert it, no problem :-)

PR #490, opened 2024-02-28
Enable master check tool to run KaaS tests

mbuechse:
resolves https://github.com/SovereignCloudStack/standards/issues/481


cah-hbaum:
I looked a bit around with this topic. IMO there are two options to run kubernetes tests with the pipeline, which do some stuff different:
1) Use a pre-provided cluster  
If we go this way, I would probably do the testing similar to how it is done already with the clouds.
We would put the `certificate-authority-data`, `client-certificate-data` and `client-key-data` from the kubeconfig file in use into the secret of the cloud provider and use this in a `pre_k8s_cluster.yaml` (similar to the existing `pre_cloud.yaml`) to setup the kubeconfig of each cloud provider. This would also require a kubeconfig jinja file (or similar).

2) Setup a new cluster each time
This would be quite elegant, since we could use the cloud instance data already available and create a new cluster to test, BUT it would take quite some time to create such a cluster every time. To create a cluster, we could either use the `k8s-cluster-api-provider` or `ClusterStacks`, which would probably both be fine (but `ClusterStacks` is the way going forward I think).

IMO for now, the quicker, easier solution is option 1. If you want @mbuechse, I can provide a patch or make a commit directly.

mbuechse:
@cah-hbaum Yes, we use option one, because (a) we are pressed for time and (b) currently there isn't any universal interface that allows us to create a cluster -- many providers use Gardener, others want to use the Clusterstacks approach from the reference implementation, but that isn't quite done yet.

We have to put the necessary files (client cert etc.) into the secrets part of the Zuul config, but that isn't what this issue is about. This issue pertains only to the Python script.

cah-hbaum:
> @cah-hbaum Yes, we use option one, because (a) we are pressed for time and (b) currently there isn't any universal interface that allows us to create a cluster -- many providers use Gardener, others want to use the Clusterstacks approach from the reference implementation, but that isn't quite done yet.
> 
> We have to put the necessary files (client cert etc.) into the secrets part of the Zuul config, but that isn't what this issue is about. This issue pertains only to the Python script.

I will make a commit then.

martinmo (approved):
LGTM!

garloff (approved):
looks good, thanks!

PR #498, opened 2024-03-05
Expand status-page-openapi-spec-decision with severity levels

joshuai96:
Impacts might have different severities on a systems component. For the further development of the `status-page-web` project, these severities must be defined on the API service.

### DIFF
--- a/Standards/scs-0402-v1-status-page-openapi-spec-decision.md
+++ b/Standards/scs-0402-v1-status-page-openapi-spec-decision.md
@@ -90,11 +90,14 @@ Furthermore, a `SeverityValue` MUST be supplied to the `Impact` when referencing
 
 ### Severity
 
-A severity contains a name, that MUST be unique and WILL be used as identifier. The value marks the upper boundary of the severity.
+A severity contains a name, that MUST be unique and will be used as identifier. The `SeverityValue` marks the upper boundary of the severity.
 
 The severity's value range is calculated by taking the previous severity's (`SeverityA`) value and adding 1 to obtain the starting point and taking the current severity's (SeverityB) value as the end point. These limits are inclusive.
 
-`SeverityA.value + 1 <= Severity <= SeverityB.value`
+```acsii
+0, ... , SeverityA.value, SeverityA.value, + 1, ... , SeverityB.value - 1, SeverityB.value, SeverityB.value + 1, ... , 100
+                        |<------------range of severity values for SeverityB------------->|
+```
 
 Example:
 
--- a/Standards/scs-0402-v1-status-page-openapi-spec-decision.md
+++ b/Standards/scs-0402-v1-status-page-openapi-spec-decision.md
@@ -43,7 +43,7 @@ An `Incremental` is used in combination with other identifiers to identify a sub
 
 #### SeverityValue
 
-A `SeverityValue` is an unsiged integer ranging from 0 to 100 inclusively. It MUST be utilized by an `Impact` when referencing a `Component` to gauge the severity of the impact on that component. It MUST NOT be added to an `Impact` when referencing an `Incident`. While being described as an unsiged integer, implementing this value MAY not require it to be an uint data type in any form, because its range even fits in a signed int8 (byte) data type.
+A `SeverityValue` is an unsiged integer ranging from 0 to 100 inclusively. It MUST be utilized by an `Impact` when referenced by a `Component` to gauge the severity of the impact on that component. It MUST NOT be added to an `Impact` when referenced by an `Incident`. While being described as an unsiged integer, implementing this value MAY not require it to be an uint data type in any form, because its range even fits in a signed int8 (byte) data type.
 
 ### API objects
 
@@ -86,7 +86,7 @@ An impact defines the relation between an incident and a component. A component 
 
 To reflect this, each component and incident can have a list of impacts, stating the type of impact and a reference to the incident or component, it refers to.
 
-Furthermore, a `SeverityValue` MUST be supplied to the `Impact` when referencing a `Component`, to gauge the impact's severity on that component.
+Furthermore, a `SeverityValue` MUST be supplied to the `Impact` when referenced by a `Component`, to gauge the impact's severity on that component.
 
 ### Severity
 
--- a/Standards/scs-0402-v1-status-page-openapi-spec-decision.md
+++ b/Standards/scs-0402-v1-status-page-openapi-spec-decision.md
@@ -43,7 +43,7 @@ An `Incremental` is used in combination with other identifiers to identify a sub
 
 #### SeverityValue
 
-A `SeverityValue` is an unsiged integer ranging from 0 to 100 inclusively. It SHOULD be utilized by an `Impact` when referencing a `Component` to gauge the severity of the impact on that component. While being described as an unsiged integer implementing this value MAY not require it to be an uint data type in any form, because it range even fits in a signed int8 (byte) data type.
+A `SeverityValue` is an unsiged integer ranging from 0 to 100 inclusively. It MUST be utilized by an `Impact` when referencing a `Component` to gauge the severity of the impact on that component. It MUST NOT be added to an `Impact` when referencing an `Incident`. While being described as an unsiged integer, implementing this value MAY not require it to be an uint data type in any form, because its range even fits in a signed int8 (byte) data type.
 
 ### API objects
 
@@ -86,7 +86,7 @@ An impact defines the relation between an incident and a component. A component 
 
 To reflect this, each component and incident can have a list of impacts, stating the type of impact and a reference to the incident or component, it refers to.
 
-Furthermore, a `SeverityValue` SHOULD be added to the `Impact` when referencing a component, to gauge the impact's severity on that component.
+Furthermore, a `SeverityValue` MUST be supplied to the `Impact` when referencing a `Component`, to gauge the impact's severity on that component.
 
 ### Severity
 
--- a/Standards/scs-0402-v1-status-page-openapi-spec-decision.md
+++ b/Standards/scs-0402-v1-status-page-openapi-spec-decision.md
@@ -82,15 +82,15 @@ An impact defines the relation between an incident and a component. A component 
 
 To reflect this, each component and incident can have a list of impacts, stating the type of impact and a reference to the incident or component, it refers to.
 
-Furthermore a severity value is added to the impact when referencing an component, to gauge the impacts severity on the system.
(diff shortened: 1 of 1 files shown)


Standards/scs-0402-v1-status-page-openapi-spec-decision.md
```diff
@@ -82,6 +82,46 @@ An impact defines the relation between an incident and a component. A component
...
+
+This means:
+
+- Operational from 0 to 10
+- Restriced from 11 to 50
+- Broken from 51 to 100.
+
+A value of 100 is the maximum of the severity value.
+
+A severity with the value of 100 MUST always be supplied. This is the heighest severty for the system. If no severty with the 100 value exists, e.g. heighest severty value is 90, an `Impact` with a heigher severity
```
snkaupe, line 122: Couple typos. Fixed:

`A severity with the value of 100 MUST always be supplied. This is the highest severity for the system. If no severity with a value of 100 exists, e.g. the highest severity value is set at 90, an `Impact` with a higher severity WILL be considered to be an _unkown_ severity.`

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
```diff
@@ -82,6 +82,46 @@ An impact defines the relation between an incident and a component. A component
 
 To reflect this, each component and incident can have a list of impacts, stating the type of impact and a reference to the incident or component, it refers to.
 
+Furthermore a severity value is added to the impact when referencing an component, to gauge the impacts severity on the system.
```
snkaupe, line 85: Couple typos. Fixed: 

``Furthermore, a _severity value_ is added to the `Impact` when referencing a component, to gauge the impact's severity on that component.``

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
```diff
@@ -82,6 +82,46 @@ An impact defines the relation between an incident and a component. A component
...
+
+### Severity
+
+A severity contains a name, that MUST be unique and WILL be used as identifier. The value marks the upper boundary of the severity.
+
+The severity value range is calculated by the previous severitys (SeverityA) value + 1 to the current severitys (SeverityB) value. These limits are inclusive.
```
snkaupe, line 91: Typo. Fixed:

``The severity value range is calculated by taking the previous severity's (`SeverityA`) value and adding 1 to obtain the starting point and taking the current severity's (SeverityB) value as the end point. These limits are inclusive.``

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
```diff
@@ -82,6 +82,46 @@ An impact defines the relation between an incident and a component. A component
...
+
+`SeverityA.value + 1 <=Severity <=SeverityB.value`
```
snkaupe, line 93: This is a bit unclear at first glance because we have `SeverityA`, `SeverityB` and an additional `Severity` which stands for "all valid values falling into the range of `SeverityB`"without clearly stating so. Would using a range be clearer?

S<sub>B</sub>.severity ∈ (S<sub>A</sub>.value, S<sub>B</sub>.value)

Alternatively, it might be enough to just replace the un-annotated `Severity` in the middle with a clearer `severity range of SeverityB` or `SeverityB.severity`.

snkaupe (changes requested):
Text needs a bit of work due to some typos and the illustration of how a severity's range is calculated being a bit unclear, but the content is good.

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
```diff
@@ -82,6 +82,46 @@ An impact defines the relation between an incident and a component. A component
...
+  {
+    "displayName": "Broken",
+    "value": 100
+  }
+]
+```
+
+This means:
+
+- Operational from 0 to 10
```
joshmue, line 116: Does that mean that there are incidents affecting services in such little ways that they should not be noticed at all by users?
Like "We're doing maintenance, but you - dear customer - should not be impacted at all"?

joshuai96: AFAIK is an "maintenance mode" planned by @snkaupe

snkaupe: Indeed. I considered just using incidents with a special severity value (such as 0 or 100) to signify a maintenance event. Thinking about it now, it might be sensible to include this in the specification, so that we can depend upon it in the frontend an backend to detect maintenance incidents. There are a few open questions regarding maintenance events anyway, so we might leave that open right now and return to it at the appropriate time, i.e. when we are discussing maintenance events in depth.

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
```diff
@@ -82,6 +82,46 @@ An impact defines the relation between an incident and a component. A component
...
+- Restriced from 11 to 50
```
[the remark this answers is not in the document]
snkaupe: The frontend right now uses the term "limited" for such components.

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
joshuai96: The `Phase`s are referenced by their name too, in combination with their generation, to uniquely identify them. There is no UUID used either. 

Giving the severities an UUID is only required, if two or more severities with the same name should be possible.

Forcing uniqueness in the names, seems a lot more logical, then having severities with the same name. When names are unique, they can be used as identifier.

A possible future requirements is i18n, but the whole API isn't set up to handle different languages, e.g. Phases. Currently only one language at a time is supported.

I18n would probably be out of scope for an MVP anyways, but maybe we should check back with @fkr about this?

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
joshuai96: Thinking about this, the statement "... should not be impacted at all..." is not quite right either.

Maintenance can include disruptions of the normal usage, but these aren't on accident, like a normal incident, but are well planned.

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
snkaupe: Sounds good, we'll discuss this with him during the sprint review later today.

Aside from localization, the use-case I could imagine for non-unique severity names would be setting up several severities of the same name for public consumption, while the different severity ranges could be used for internal purposes, such as reporting. I am, however, uncertain how realistic this scenario is.

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
snkaupe: Good argument. Is a maintenance session that is taking down a component not basically the same as any other kind of incident that disrupts said component to a point where it becomes unavailable, just that we are aware of it before it happens? Maybe a maintenance event should either be indicated by a special flag inside an incident or be represented by an entirely different resource.

joshuai96: As mentioned by you in https://github.com/SovereignCloudStack/issues/issues/366#issuecomment-1959180141 a  maintenance is a kind of impact, that is known beforehand and currently the API has no way of managing this kind of behavior.

Either we allow for future impacts and handle their retrieval on the API like some kind of special case, or include a new resource for maintenance. I would prefer the later, to keep it more separated from the "normal" impacts. This may be clarified with @fkr today, too.

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
```diff
@@ -82,6 +86,45 @@ An impact defines the relation between an incident and a component. A component
 
 To reflect this, each component and incident can have a list of impacts, stating the type of impact and a reference to the incident or component, it refers to.
 
+Furthermore, a `SeverityValue` SHOULD be added to the `Impact` when referencing a component, to gauge the impact's severity on that component.
```
joshuai96, line 89: Is `SHOULD` the right wording here? 
Is it optional but recommended to add the value to the impact, or should this be required? @snkaupe

snkaupe: I'd prefer it being required.

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
joshuai96: As discussed in the sprint review the i18n feature is out of scope for the MVP.

Severities will use their unique name as identifier.

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
joshuai96: As discussed in the sprint review: A maintenance mode is out of scope for the MVP. So no further considerations about "future issues" at the moment.

joshuai96: To address @joshmue original question, an impact might be so insignificant that operation is basically not impacted e.g. Main router is down, but fail over activated a secondary one. An incident, but everything is still working fine.

Its an example after all, the definition of severities is part of the configuration done by the user who wants to run the status-page. What names and values they use is their choice.

Is this discussion resolved with it, or any change requests?

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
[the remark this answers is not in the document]
joshuai96: Changed to limited to have it a little more in line.

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
joshmue: Yes, having "Operational" as example severity seems LGTM in that regard.

Resolving this discussion...

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
snkaupe: Just noticed that my range is wrong, it should be [S<sub>A</sub>.value, S<sub>B</sub>.value), as the value of S<sub>A</sub> is not part of it.

Still, the following notation might be simpler and preferable:

``0, ... , SeverityA.value, SeverityA.value, + 1, ... , SeverityB.value - 1, SeverityB.value, SeverityB.value + 1, ... , 100
                        |<------------range of severity values for SeverityB------------->|
``

snkaupe (approved):
All fine now!

snkaupe (approved):
Looks good to me.

joshmue (approved):

PR #499, opened 2024-03-05
Minor editorial improvements to K8s version policy standards scs-0210-v1 and scs-0210-v2

martinmo:
While working on #488 I noticed several shortcomings in the text of the version policy standards scs-0210-v1 and scs-0210-v2, which this PR intends to fix.

I've split the commits to make it easier to review (hopefully), because some of them tackle only cosmetic issues and improve link sharing between the documents by making use of reference style Markdown links.

Most importantly, this PR intends to resolve the following ambiguities:

- What do we mean exactly with a critical CVE?
- What do we mean exactly with a supported K8s version?
- What are CNCF E2E tests or where can I find them? (my understanding is: https://github.com/cncf/k8s-conformance)

Furthermore, the conformance test section was outdated or missing in both standard versions and I fixed this.

### DIFF
--- a/Standards/scs-0210-v1-k8s-new-version-policy.md
+++ b/Standards/scs-0210-v1-k8s-new-version-policy.md
@@ -53,11 +53,10 @@ All important documents regarding versioning, releases, etc. for the official Ku
 
 ## Conformance Tests
 
-The conformance test is written in the 'k8s-version-recency-check.py' script. The script requires the path to a valid
-kubeconfig file, which should describe the k8s cluster under test. This can either be done by creating a config from
-the also provided 'config.yaml.template' or by calling the test script with its cli arguments.
+The conformance tests have been obsoleted by the tests for [scs-0210-v2][scs-0210-v2].
 
 [k8s-releases]: https://kubernetes.io/releases/
 [k8s-release-cycle]: https://kubernetes.io/releases/release/#the-release-cycle
 [k8s-release-cadence]: https://kubernetes.io/releases/patch-releases/#cadence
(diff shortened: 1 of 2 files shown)


cah-hbaum (approved):
LGTM, nice updates!

martinmo:
> Seems like minor editorial improvements, nothing to write home about ;)

Exactly =)

> Only one issue: You removed the parts about the now-obsolete test for v1. I think the sentence that this is now obsoleted (or rather **superseded**) by the new test is fine. But what about the old test? Did we remove it entirely?

Yes, I did in #488 and it was approved ;)

> I'm not sure that this was a good call. Well, if that's the case, then so be it. Out of curiosity: would it be very hard to add a v1-mode to the new test script?

If it is important, I would prefer to just restore the old script which also lived in a separate directory. And of course I would revert the changes wrt conformance testing in the scs-0210-v1 text.

mbuechse:
> If it is important, I would prefer to just restore the old script which also lived in a separate directory. And of course I would revert the changes wrt conformance testing in the scs-0210-v1 text.

Maybe we restore it. We don't delete the standard, so let's also keep the test.

martinmo:
> > If it is important, I would prefer to just restore the old script which also lived in a separate directory. And of course I would revert the changes wrt conformance testing in the scs-0210-v1 text.
> 
> Maybe we restore it. We don't delete the standard, so let's also keep the test.

I have created a separate PR #503 for restoring the test and reverted the change in scs-0210-v1 conformance test section here.

Issue #492, opened 2024-03-04
[BUG] Double dash CLI arguments do not work in k8s-version-recency-check


matofeder:
Double dash CLI arguments `--config` and `--kubeconfig` do not work as expected. Is seems that they are not recognized by the script k8s-version-recency-check.py.

See the STDOUT when `--config ` is used: 
```
$ python kaas/k8s-version-recency/k8s-version-recency-check.py --config config.yaml -k kubeconfig.yaml 
The config file under couldn't be found, falling back to the default config.
```

See the STDOUT when `--kubeconfig ` is used: 
```
$ python kaas/k8s-version-recency/k8s-version-recency-check.py -c config.yaml --kubeconfig kubeconfig.yaml
WARNING: Config not found: 
Traceback (most recent call last):
  File "/home/mato/Dev/ifne/scs/standards/Tests/kaas/k8s-version-recency/k8s-version-recency-check.py", line 467, in <module>
    return_code = asyncio.run(main(sys.argv[1:]))
  File "/usr/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "/home/mato/Dev/ifne/scs/standards/Tests/kaas/k8s-version-recency/k8s-version-recency-check.py", line 445, in main
    cluster_version, cluster_name = await get_k8s_cluster_version(config.kubeconfig)
  File "/home/mato/Dev/ifne/scs/standards/Tests/kaas/k8s-version-recency/k8s-version-recency-check.py", line 380, in get_k8s_cluster_version
    cluster_config = await kubernetes_asyncio.config.load_kube_config(kubeconfig)
  File "/home/mato/Dev/ifne/scs/standards/.venv/lib/python3.10/site-packages/kubernetes_asyncio/config/kube_config.py", line 598, in load_kube_config
    loader = _get_kube_config_loader_for_yaml_file(
  File "/home/mato/Dev/ifne/scs/standards/.venv/lib/python3.10/site-packages/kubernetes_asyncio/config/kube_config.py", line 562, in _get_kube_config_loader_for_yaml_file
    return KubeConfigLoader(
  File "/home/mato/Dev/ifne/scs/standards/.venv/lib/python3.10/site-packages/kubernetes_asyncio/config/kube_config.py", line 149, in __init__
    self.set_active_context(active_context)
[...]

PR #503, opened 2024-03-06
Restore scs-0210-v1 conformance tests

martinmo:
This restores the conformance check for scs-0210-v1 to exactly the state
as it was before https://github.com/SovereignCloudStack/standards/commit/55cdcb05f7c5a9187c8077a62620eebfe32ad956 and applies the bugfix for #492.


mbuechse (approved):
Great, thanks!

PR #508, opened 2024-03-08
Rework so only one cluster is checked

### LINKED ISSUE #507: Rework test for scs-0210-v2 to better integrate with e2e pipeline
In its current form, the test expects multiple k8s clusters, one for each active k8s branch, so we can check (among other things) that each active branch is being offered. This poses two problems:

- the end-to-end pipeline works with one cluster only;
- we need to check that each k8s version offered is compliant with all SCS standards (not just the k8s version policy), so the requirement that all active k8s versions be offered is more like a 'meta-requirement'.

In my view, we could (hypothetically) say that the scope `scs-compatible kaas` is parametrized with the k8s branch, so we have `scs-compatible kaas [v1.29]` etc., and then the plain `scs-compatible kaas` would be a (time-variant) scope that (today) comprised the three scopes `scs-compatible kaas [v1.29]`, `scs-compatible kaas [v1.28]`, `scs-compatible kaas [v1.27]`.

Short of these hypothetical considerations, I think we can make a pragmatic decision as follows: not every assertion from the standard has the same importance, and not every assertion from the standard needs to be tested automatically. For the time being, we could fall back to a test that merely checks that the k8s version of **one** given cluster

- is of an active branch
- is recent enough (with respect to patches, CVEs etc.)

Then the test can be integrated with tthe pipeline no problem. To this end, it should expect a kubeconfig (which it currently already does) and an optional context (currently, it looks at up to four predefined contexts).

mbuechse:
Resolves #507


mbuechse:
@martinmo I lack the kubeconfig necessary to test this... could you take over? Or maybe give me a quick Minikube rundown...

Tests/kaas/k8s-version-policy/k8s_version_policy.py
```diff
@@ -132,6 +134,8 @@ def parse_arguments(argv):
             config.config_path = opt[1]
```
martinmo, line 134: This can be removed.

Tests/kaas/k8s-version-policy/k8s_version_policy.py
```diff
@@ -132,6 +134,8 @@ def parse_arguments(argv):
...
         if opt[0] == "-k" or opt[0] == "--kubeconfig":
             config.kubeconfig = opt[1]
+        if opt[0] == "--context":
```
martinmo, line 137:
```suggestion
        if opt[0] == "-C" or opt[0] == "--context":
```

If you agree that a shorthand should be available.

Tests/kaas/k8s-version-policy/k8s_version_policy.py
```diff
@@ -479,63 +487,40 @@ async def main(argv):
...
-            else:
+        logger.info("Checking cluster of kubeconfig context '%s'.", config.context)
+        cluster = await get_k8s_cluster_info(config.kubeconfig, config.context)
+        cluster_branch = cluster.version.branch
+
+        if cluster_branch not in supported_branches:
+            logger.error("The K8s cluster version %s of cluster '%s' is already EOL.", cluster.version, cluster.name)
+
+        # XXX: When should we pass allow_older=True? I don't see it in the standard
+        if check_k8s_version_recency(cluster.version, cve_affected_ranges):
```
martinmo, line 499:
```suggestion
        if cluster_branch not in supported_branches:
            logger.error("The K8s cluster version %s of cluster '%s' is already EOL.", cluster.version, cluster.name)
        elif check_k8s_version_recency(cluster.version, cve_affected_ranges, allow_older=True):
```

**Issue 1**: `elif` is needed because otherwise confusing output can occur:

```
INFO: Checking cluster of kubeconfig context 'oldoldstable'.
ERROR: The K8s cluster version 1.26.14 of cluster 'oldoldstable' is already EOL.
INFO: The K8s cluster version 1.26.14 of cluster 'oldoldstable' is still in the recency time window.
```

**Issue 2**: The `allow_older` param must be set to `True` always now. Otherwise, for example, a v1.27 cluster would be detected as outdated by the script.

I introduced `allow_older` to able to reuse the already existing `check_k8s_version_recency(…)` function for clusters that do not use the latest K8s branch, which would be the case for `oldstable` and older but not for `stable` according to the abandoned context name scheme. (So it is an conformance test implementation detail and has no corresponding text in the standard.)

Admittedly this is confusing. Root cause is that this function does two things at a time which should be separate: a) checking if a recent patch is applied for an arbitrary K8s branch and b) checking if the newest K8s branch (plus/minus 4 month) cadence is used.

martinmo (changes requested):
I've tested it with a set of minikube profiles. There are just a few issues that need to be settled, see my suggestions above. If they are applied, everything works as expected (checked with cluster versions 1.26.x trough 1.29.x). Good work :+1: – and sorry for the confusing bits in my code! :grimacing:

Tests/kaas/k8s-version-policy/k8s_version_policy.py
```diff
@@ -395,44 +392,32 @@ async def get_k8s_cluster_info(kubeconfig, context=None) -> ClusterInfo:
...
+        # at this point `release` has the same major.minor, but higher patch than `my_version`
+        if release.age > PATCH_VERSION_CADENCE:
+            # whoops, the cluster should have been updated to his (or a higher version) already!
+            return False
+        if release.version in cve_version_list and release.age > CVE_VERSION_CADENCE:
+            # two FIXMEs:
+            # (a) can the `in` ever become true, when we have a version vs a set of ranges
+            # (b) if the release has a CVE, then there is no use if we updated to it?
+            # shouldn't we check for CVEs of my_version and then check whether the new one still has them?
+            return False
```
martinmo, line 420: Regarding (b): You are absolutely right, it should be like this:

```suggestion
        if my_version in cve_version_list and release.age > CVE_VERSION_CADENCE:
            # whoops, the cluster version is affected by a critical CVE and
            # should have already been updated before the usual patch cadence
            return False
```

I think we don't need to check if `release.version` is affected, because we don't tell the user to update to a specific version. We just say that the currently used version has a critical CVE. And if the user fails to update to a version that fixes the CVE and runs the script again it will still complain.

Regarding (a): yes, if the version range `cve_version_list` just represents one version (i.e., only lower_version is set), the `in` should work as expected.

Tests/kaas/k8s-version-policy/k8s_version_policy.py
```diff
@@ -395,44 +392,32 @@ async def get_k8s_cluster_info(kubeconfig, context=None) -> ClusterInfo:
...
-
-        if release.version.minor == (my_version.minor + 1) and release.version.patch == 0:
-            break
+        if my_version.branch != release.version.branch:
+            continue
+        if my_version.patch >= release.version.patch:
+            continue
+        # at this point `release` has the same major.minor, but higher patch than `my_version`
+        if release.age > PATCH_VERSION_CADENCE:
+            # whoops, the cluster should have been updated to his (or a higher version) already!
```
martinmo, line 413:
```suggestion
            # whoops, the cluster should have been updated to this (or a higher version) already!
```

martinmo (changes requested):
Only two little changes and this is ready to go in my opinion.

I like how you refactored the `check_k8s_version_recency(…)` function, it is now a lot easier to understand.

Tests/kaas/k8s-version-policy/k8s_version_policy.py
mbuechse: The standard expressly states (bold by me):

> This time period MUST be even shorter for patches that target critical CVEs (CVSS >= 8).
> It is RECOMMENDED to **provide a new patch version in a 2 day time period after their release**.

A patch version is a version that doesn't contain the CVE, right? If you agree with my interpretation here, I will gladly implement it.

mbuechse: (And, while I'm at it, I will also change `their release` to `its release`)

mbuechse: Oh, maybe `their release` refers to the release of the CVEs. Still, I assume the patch version should be one that fixes the CVE.

martinmo: > A patch version is a version that doesn't contain the CVE, right? If you agree with my interpretation here, I will gladly implement it.

No, "patch" just refers to the last number Z in the `X.Y.Z` version scheme. I would rather call it the "fixed version" to avoid confusion.

I think you are right: if `my_version` X.Y.Z has a critical CVE, the algorithm currently only checks if an at least 2 days old patch version X.Y.Z+1 is available and fails. But Z+1 doesn't necessarily need to be the fixed version. It could be Z+2 or later which is younger than 2 days.

Furthermore: the standard says *RECOMMENDED*. The script says *ERROR* in this case.

mbuechse: You are right about the terminology (patch). Rather, I should have pointed to this quote: `for patches that target critical CVEs` -- but our conclusions are in agreement anyway.

PR #517, opened 2024-03-14
Adapt certificate scope YAML and main check tool to facilitate evaluating test results

### LINKED ISSUE #510: Facilitate evaluating test results across resource-usage classifications
## The test script cannot produce a compliance judgment

The regulations for achieving certification make the following distinction:

> The tests MUST be run at given intervals, depending on their resource-usage classification:
> 
> - _light_: at least nightly,
> - _medium_: at least weekly,
> - _heavy_: at least monthly.

Note that the provider may follow a different scheme such as

- light+medium: nightly (the same run, the same report!)
- heavy: weekly

Whatever the case may be, the bottom line is that main test script cannot produce an ultimate judgment as to whether the test subject is compliant or not, except when all classifications are being tested together---and then, still, there may be manual tests that have to be considered.

## A puzzle of multiple pieces

In other words, this judgment requires multiple puzzle pieces from multiple sources to be considered. Therefore, the report must indicate which of the pieces it contains. For instance, the main test script could group results by this resource-usage classification and report them accordingly, like so (assume a fixed test date, subject, scope, and version):

- light: 1 error
- medium: 0 errors
- heavy: n/a
- manual: n/a

We would store all puzzle pieces in our database. In order to produce the ultimate judgment, we would have to fetch the most recent piece for each category (light/medium/heavy/manual). Then we discard all pieces that are too old (according to the quote at the top). Finally, the test subject passes if all necessary pieces are present and none of them is a "fail".

## More fine-grained jigsaw

The granularity of these puzzle pieces is a design choice, and the example above demonstrated the most coarse-grained possibility. On the other end of the spectrum, we could report more like (again, assume a fixed test date, subject, scope and version):

- standard X1, check script Y1: 1 error
- standard X2, check script Y2: 0 errors
- ...

**One major advantage would be that failing tests could be repeated in isolation.**

I did a bit of thinking about this possibility, but it seems to me that it's unnecessary. I will proceed with the coarse-grained puzzle.

mbuechse:
Resolves #510

### DIFF
--- a/Standards/scs-0003-v1-sovereign-cloud-standards-yaml.md
+++ b/Standards/scs-0003-v1-sovereign-cloud-standards-yaml.md
@@ -137,24 +137,38 @@ Every list of standards consists of several standards that – altogether – de
 | `name`                   | String | Full name of the particular standard                                                                   | _Flavor naming_                                                                                                                |
 | `url`                    | String | Valid URL to the latest raw version of the particular standard                                         | _[Flavor naming](https://raw.githubusercontent.com/SovereignCloudStack/standards/main/Standards/scs-0100-v2-flavor-naming.md)_ |
 | `condition`              | String | State of the particular standard, currently either `mandatory` or `optional`, default is `mandatory`   | _mandatory_                                                                                                                    |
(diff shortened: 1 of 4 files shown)


cah-hbaum (approved):
LGTM!

PR #522, opened 2024-03-14
standardizing IPv4 networking in SCS

cah-patrickthiem:
This is the initial draft document to standardize IPv4 networking in the context of SCS.

### DIFF
--- a/Standards/scs-xxxx-v1-ipv4-networking-standard.md
+++ b/Standards/scs-xxxx-v1-ipv4-networking-standard.md
@@ -23,40 +23,48 @@ ease of use for IaaS users and DevOps teams.
 
 ### Options considered
 
-#### _Option 1_
+#### _Neutron Routers_
 
-Usage of Neutron Routers: Neutron routers are mandated to manage traffic between
-internal and external networks. They shall act as the default gateway for VMs
-requiring access to external networks and the internet, thereby facilitating the
-routing of traffic and enhancing network security.
+Usage of Neutron Routers: To manage traffic between internal and external networks Neutron Routers **MUST** be used as the default gateway for VMs requiring access to external networks and the internet, thereby facilitating the routing of traffic and enhancing network security. 
 
-#### _Option 2_
+CSPs **SHOULD** use OVN or L3agent as High Availability (HA) service deployments.
+Standard external networks **MUST NOT** be made accessible as _shared networks_. It is advised that external networks are only reachable by the usage of routing and floating IPs.
+However, for special use cases like certain storage or VPN solutions it could be useful to allow _direct access networks_.
 
-Security Group Policies: Standardized security group policies shall be applied to all
-instances utilizing public IPv4 addresses. These policies must define and enforce access
+External networks and subnets **SHOULD** (very strong should) be configured with _--no-dhcp_ (DHCP - Dynamic Host Configuration Protocol). It is more secure to configure it like this, since it gives less space for reflection attacks, e.g. _Denial of Service_ (DOS) attacks. If _dhcp_ is configured, certain firewall configurations **MUST** be made to catch IPs from the _Neutron dhcp agent_ in the public network. 
+
+#### _Neutron Plugins_
+
+Neutron Plugins: A SCS conform CSP **MAY** use RBAC and VPNaaS plugins.
+(Neutron RBAC needs to be configured explicitly to be able to use it. If configured, Neutron configurations can be shared across OpenStack projects. It also can be beneficial for admins, since an admin could bind external networks only to certain projects.)
+
+#### _Security Groups_
+
+Security Group Policies: Standardized security group policies **SHOULD** be applied to all instances utilizing public IPv4 addresses. These policies must define and enforce access
 controls to ensure the security of the cloud environment.
+Security Groups **SHOULD** be enabled by default but **MUST** be capable of being switched off.
 
-#### _Option 3_
-IP Usage Monitoring: SCS clouds must implement monitoring solutions to track the
-utilization of IPv4 addresses. This facilitates efficient management of resources and
-supports capacity planning efforts.
+#### _Quota & Monitoring_
 
-#### _Option 4_
-External Network Naming: All SCS clouds must adopt the naming convention
-scs-external-net for external networks. This standardization facilitates easier
-identification and management of external network resources.
+Quota: The standard quota of floating IPs and routers **SHOULD** be rather small, e.g. 3-5 floating IPs. This ensures a more fair distribution of these resources for all cloud users. If a user wants to use more of these resources, the user **SHOULD** be able to pay for more.
 
-#### _Option 5_
-Floating IPs for Dynamic Allocation: Utilize Floating IPs to allow dynamic reassignment
-of public IPv4 addresses to different instances (VMs or Loadbalancers), facilitating
-high availability and fault tolerance.
+IP Usage Monitoring: SCS CSPs **SHOULD** implement monitoring solutions to track the utilization of IPv4 addresses. This facilitates efficient management of resources and supports capacity planning efforts.
+
+#### _External Network Naming_
+
+All SCS clouds **SHOULD** adopt the naming convention
+scs-external-net for external networks. This standardization facilitates easier identification and management of external network resources.
+
+#### _Floating IPs_
+
+Floating IPs for Dynamic Allocation: Utilization of Floating IPs to allow dynamic reassignment of public IPv4 addresses to different instances (VMs or Loadbalancers), facilitating high availability and fault tolerance.
+Floating IPs **MUST** be enabled.
 
 ## Open questions
 
 - Naming Convention Flexibility: How rigid should the naming convention for external
 networks be across various SCS clouds?
-- Security Policy Standardization: To what extent should security group policies be
-standardized without impeding individual cloud providers' autonomy?
+- Load Balancing: Do we want to dictate a Load Balancer or a set of Load Balancers or nothing at all? E.g. Octavia, Yawol
 
 ## Decision
 
--- a/Standards/scs-xxxx-v1-ipv4-networking-standard.md
+++ b/Standards/scs-xxxx-v1-ipv4-networking-standard.md
@@ -1,0 +1,71 @@
+---
+title: IPv4 Networking Standard
+type: Standard
+status: Proposal
+track: IaaS
+---
+
+## Introduction
+
+This document outlines the standardized approach for the management and allocation of
+public IPv4 addresses within Sovereign Cloud Stack (SCS) environments. Its aim is to
+ensure a consistent, secure, and efficient methodology for IP address provisioning
+across all SCS cloud services.
+
+## Motivation
+
+The motivation behind establishing this standard is to enhance interoperability, improve
+security measures, and streamline the operational processes across different SCS clouds.
+It addresses the need for a unified procedure in handling IPv4 networking to facilitate
+ease of use for IaaS users and DevOps teams.
+
+## Design Considerations
+
+### Options considered
+
+#### _Option 1_
+
+Usage of Neutron Routers: Neutron routers are mandated to manage traffic between
+internal and external networks. They shall act as the default gateway for VMs
+requiring access to external networks and the internet, thereby facilitating the
+routing of traffic and enhancing network security.
+
+#### _Option 2_
+
+Security Group Policies: Standardized security group policies shall be applied to all
+instances utilizing public IPv4 addresses. These policies must define and enforce access
+controls to ensure the security of the cloud environment.
+
+#### _Option 3_
+IP Usage Monitoring: SCS clouds must implement monitoring solutions to track the
+utilization of IPv4 addresses. This facilitates efficient management of resources and
+supports capacity planning efforts.
+
+#### _Option 4_
+External Network Naming: All SCS clouds must adopt the naming convention
+scs-external-net for external networks. This standardization facilitates easier
+identification and management of external network resources.
+
+#### _Option 5_
+Floating IPs for Dynamic Allocation: Utilize Floating IPs to allow dynamic reassignment
+of public IPv4 addresses to different instances (VMs or Loadbalancers), facilitating
+high availability and fault tolerance.
+
+## Open questions
+
+- Naming Convention Flexibility: How rigid should the naming convention for external
+networks be across various SCS clouds?
+- Security Policy Standardization: To what extent should security group policies be
+standardized without impeding individual cloud providers' autonomy?
+
+## Decision
+
+...
+
+## Related Documents
+
+Related Documents, OPTIONAL
+
+## Conformance Tests
+
+Conformance Tests, OPTIONAL
--- a/Standards/scs-xxxx-v1-ipv4-networking-standard.md
+++ b/Standards/scs-xxxx-v1-ipv4-networking-standard.md
@@ -25,13 +25,13 @@ ease of use for IaaS users and DevOps teams.
 
 #### _Neutron Routers_
 
-Usage of Neutron Routers: To manage traffic between internal and external networks Neutron Routers **MUST** be used as the default gateway for VMs requiring access to external networks and the internet, thereby facilitating the routing of traffic and enhancing network security. 
+Usage of Neutron Routers: To manage traffic between internal and external networks Neutron Routers **MUST** be used as the default gateway for VMs requiring access to external networks and the internet, thereby facilitating the routing of traffic and enhancing network security.
 
 CSPs **SHOULD** use OVN or L3agent as High Availability (HA) service deployments.
 Standard external networks **MUST NOT** be made accessible as _shared networks_. It is advised that external networks are only reachable by the usage of routing and floating IPs.
 However, for special use cases like certain storage or VPN solutions it could be useful to allow _direct access networks_.
 
-External networks and subnets **SHOULD** (very strong should) be configured with _--no-dhcp_ (DHCP - Dynamic Host Configuration Protocol). It is more secure to configure it like this, since it gives less space for reflection attacks, e.g. _Denial of Service_ (DOS) attacks. If _dhcp_ is configured, certain firewall configurations **MUST** be made to catch IPs from the _Neutron dhcp agent_ in the public network. 
+External networks and subnets **SHOULD** (very strong should) be configured with _--no-dhcp_ (DHCP - Dynamic Host Configuration Protocol). It is more secure to configure it like this, since it gives less space for reflection attacks, e.g. _Denial of Service_ (DOS) attacks. If _dhcp_ is configured, certain firewall configurations **MUST** be made to catch IPs from the _Neutron dhcp agent_ in the public network.
 
 #### _Neutron Plugins_
 


Standards/scs-xxxx-v1-ipv4-networking-standard.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Design Considerations
+
+### Options considered
+
+#### _Neutron Routers_
+
+Usage of Neutron Routers: To manage traffic between internal and external networks Neutron Routers **MUST** be used as the default gateway for VMs requiring access to external networks and the internet, thereby facilitating the routing of traffic and enhancing network security.
+
+CSPs **SHOULD** use OVN or L3agent as High Availability (HA) service deployments.
```
artificial-intelligence, line 30: I miss some reasoning why exactly I need to use OVN or L3 agent, also notice that due to some intricacies L3 agent is not really HA currently in some edge cases, like upgrades/reboots, depending on your exact setup - this technical discussion is probably out of scope for this document though.

to be clear I'm totally for using ovn, but we should write something down why we encourage it's use, what are advantages etc.

artificial-intelligence (changes requested):
I'll add more commentary later, thanks for providing the first draft of this!

Standards/scs-xxxx-v1-ipv4-networking-standard.md
```diff
@@ -0,0 +1,79 @@
...
+Standard external networks **MUST NOT** be made accessible as _shared networks_. It is advised that external networks are only reachable by the usage of routing and floating IPs.
+However, for special use cases like certain storage or VPN solutions it could be useful to allow _direct access networks_.
+
+External networks and subnets **SHOULD** (very strong should) be configured with _--no-dhcp_ (DHCP - Dynamic Host Configuration Protocol). It is more secure to configure it like this, since it gives less space for reflection attacks, e.g. _Denial of Service_ (DOS) attacks. If _dhcp_ is configured, certain firewall configurations **MUST** be made to catch IPs from the _Neutron dhcp agent_ in the public network.
+
+#### _Neutron Plugins_
+
+Neutron Plugins: A SCS conform CSP **MAY** use RBAC and VPNaaS plugins.
```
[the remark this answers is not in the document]
josephineSei: Does a current list of plugins exist? When I look into the code [^1] I only see the ML2 Plugin. Also the documentation does not mention a plugin list (anymore). It seems to me, that most of it was removed or moved to `extensions`. When looking into the openstack repository on github, there are some deprecated plugins, a lot of charms-repos, some repos that could be plugins or agents and some other stuff [^2].
From working in secustack I know it is possible to add custom made Plugins.

I looked into my devstack config and it does not specify any plugins, except the ML2:
```
[DEFAULT]
service_plugins = ovn-router
rpc_state_report_workers = 0
api_workers = 2
notify_nova_on_port_data_changes = True
notify_nova_on_port_status_changes = True
auth_strategy = keystone
debug = True
core_plugin = ml2
dhcp_agent_notification = False
transport_url = rabbit://stackrabbit:MuchSecretSuchW0W@192.168.23.238:5672/
logging_exception_prefix = ERROR %(name)s ^[[01;35m%(instance)s^[[00m
logging_default_format_string = %(color)s%(levelname)s %(name)s [^[[00;36m-%(color)s] ^[[01;35m%(instance)s%(color)s%(message)s^[[00m
logging_context_format_string = %(color)s%(levelname)s %(name)s [^[[01;36m%(global_request_id)s %(request_id)s ^[[00;36m%(project_name)s %(user_name)s%(color)s] ^[[01;35m%(instance)s%(color>
logging_debug_format_suffix = ^[[00;33m{{(pid=%(process)d) %(funcName)s %(pathname)s:%(lineno)d}}^[[00m
bind_host = 0.0.0.0
use_syslog = False
state_path = /opt/stack/data/neutron
```

So all extensions seem to be usable all the time - because I was able to test the `network rbac`.

My conclusion: We should not state anything about plugins here, as they are poorly documented, not well maintained or even completely customized. We could discuss about letting CSPs add customized plugins. But all of these plugins while touching networking issues should not interfere with the scope of this standard. So exclude this, add a new issue and maybe we will find out enough for a new standard or maybe we just have a short note about Plugins overall on the docs-page.


[^1]: https://github.com/openstack/neutron/tree/master/neutron/plugins
[^2]: https://github.com/orgs/openstack/repositories?language=&q=neutron&sort=&type=all

Standards/scs-xxxx-v1-ipv4-networking-standard.md
```diff
@@ -0,0 +1,79 @@
...
+(Neutron RBAC needs to be configured explicitly to be able to use it. If configured, Neutron configurations can be shared across OpenStack projects. It also can be beneficial for admins, since an admin could bind external networks only to certain projects.)
```
josephineSei, line 39: I still wonder which RBAC do you mean? Can you point me towards this please?

Because it is already confusing to differ between the keystone roles and scopes RBAC that has to be implemented by every OpenStack service and this `network rbac` extension of Neutron, which allows sharing of resources - that I don't recommend. Maybe we need a standard that forbids the usage and adjusts the policy?

Standards/scs-xxxx-v1-ipv4-networking-standard.md
```diff
@@ -0,0 +1,79 @@
...
+
+#### _Security Groups_
+
+Security Group Policies: Standardized security group policies **SHOULD** be applied to all instances utilizing public IPv4 addresses. These policies must define and enforce access
+controls to ensure the security of the cloud environment.
+Security Groups **SHOULD** be enabled by default but **MUST** be capable of being switched off.
```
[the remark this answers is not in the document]
josephineSei: @markus-hentsch is correct Security Groups are part of Neutron extensions and thus cannot be switched of: https://github.com/openstack/neutron/blob/master/neutron/extensions/securitygroup.py

Security Groups are always there and being used in VMs. Even if a user does not specify anything - in that case the default VM is being used.

Nevertheless this topic does not interfere with the scope of this standard and should be included. But as we maybe want to include an architecture definition here, it would be good to reference all the work we do for security groups.

And I wonder, what do you mean with ` security group policies ` ?  Do you mean the default rules for security groups? -> In that case you can just link the DR from me and maybe later on the guide and the standard:

DR: https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0113-v1-security-groups-decision-record.md
guide and standard for the rules are still in progress.

Standards/scs-xxxx-v1-ipv4-networking-standard.md
```diff
@@ -0,0 +1,79 @@
...
+
+#### _Quota & Monitoring_
+
+Quota: The standard quota of floating IPs and routers **SHOULD** be rather small, e.g. 3-5 floating IPs. This ensures a more fair distribution of these resources for all cloud users. If a user wants to use more of these resources, the user **SHOULD** be able to pay for more.
```
josephineSei, line 49: Yes the Quota is important and belongs into this standard. We can still argue about the number :)

Standards/scs-xxxx-v1-ipv4-networking-standard.md
```diff
@@ -0,0 +1,79 @@
...
+
+IP Usage Monitoring: SCS CSPs **SHOULD** implement monitoring solutions to track the utilization of IPv4 addresses. This facilitates efficient management of resources and supports capacity planning efforts.
+
+#### _External Network Naming_
+
+All SCS clouds **SHOULD** adopt the naming convention
+scs-external-net for external networks. This standardization facilitates easier identification and management of external network resources.
```
josephineSei, line 56: We have to clarify why we would need a Naming convention at all. Renaming networks without good reasons is not really helpful.

Pre-statement: external networks can be listed with `openstack network list --external`.

1. How many CSPs have more than 1 external network (for IPv4) ?
2. How many CSPs have multiple different subnets for one external network == more than one floating-IP-pool ?
3. Are there external networks that are only for one specific customer (i have seen something like this)?

Pro:
- distinguish between IPv4 and IPv6 external networks

Con:
- how to deal with multiple IPv4 external networks?
- use other options like tags:
```
stack@devstack:~/devstack$ openstack network show public
+---------------------------+----------------------------------------------------------------------------+
| Field                     | Value                                                                      |
+---------------------------+----------------------------------------------------------------------------+
| admin_state_up            | UP                                                                         |
| availability_zone_hints   |                                                                            |
| availability_zones        |                                                                            |
| created_at                | 2024-01-24T16:12:31Z                                                       |
| description               |                                                                            |
| dns_domain                | None                                                                       |
| id                        | 73edb86b-d7ab-4db3-82b7-25fa8b012e40                                       |
| ipv4_address_scope        | None                                                                       |
| ipv6_address_scope        | None                                                                       |
| is_default                | True                                                                       |
| is_vlan_transparent       | None                                                                       |
| mtu                       | 1500                                                                       |
| name                      | public                                                                     |
| port_security_enabled     | True                                                                       |
| project_id                | 15f2ab0eaa5b4372b759bde609e86224                                           |
| provider:network_type     | flat                                                                       |
| provider:physical_network | public                                                                     |
| provider:segmentation_id  | None                                                                       |
| qos_policy_id             | None                                                                       |
| revision_number           | 4                                                                          |
| router:external           | External                                                                   |
| segments                  | None                                                                       |
| shared                    | False                                                                      |
| status                    | ACTIVE                                                                     |
| subnets                   | 3e0206bc-53c8-44ca-a0f1-2c2548bba766, 84dffd43-6d7f-4c2f-9180-8f0f0b83c9d4 |
| tags                      | IPv4                                                                       |
| tenant_id                 | 15f2ab0eaa5b4372b759bde609e86224                                           |
| updated_at                | 2024-03-28T09:39:03Z                                                       |
+---------------------------+----------------------------------------------------------------------------+
stack@devstack:~/devstack$ openstack network list --external --long --tag IPv4
+----------------------------+--------+--------+----------------------------+-------+--------+----------------------------+--------------+-------------+--------------------+------+
| ID                         | Name   | Status | Project                    | State | Shared | Subnets                    | Network Type | Router Type | Availability Zones | Tags |
+----------------------------+--------+--------+----------------------------+-------+--------+----------------------------+--------------+-------------+--------------------+------+
| 73edb86b-d7ab-4db3-82b7-   | public | ACTIVE | 15f2ab0eaa5b4372b759bde609 | UP    | False  | 3e0206bc-53c8-44ca-a0f1-   | flat         | External    |                    | IPv4 |
| 25fa8b012e40               |        |        | e86224                     |       |        | 2c2548bba766, 84dffd43-    |              |             |                    |      |
|                            |        |        |                            |       |        | 6d7f-4c2f-9180-            |              |             |                    |      |
|                            |        |        |                            |       |        | 8f0f0b83c9d4               |              |             |                    |      |
+----------------------------+--------+--------+----------------------------+-------+--------+----------------------------+--------------+-------------+--------------------+------+
```

I am pretty much for investigating those tags! Help from CSPs is wanted (we don't want to accidentally render a network not working anymore :D )

Standards/scs-xxxx-v1-ipv4-networking-standard.md
```diff
@@ -0,0 +1,79 @@
...
+
+#### _Floating IPs_
+
+Floating IPs for Dynamic Allocation: Utilization of Floating IPs to allow dynamic reassignment of public IPv4 addresses to different instances (VMs or Loadbalancers), facilitating high availability and fault tolerance.
+Floating IPs **MUST** be enabled.
```
josephineSei, line 61: Again the question: Can they be disabled? OR just not set in pools?

Standards/scs-xxxx-v1-ipv4-networking-standard.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Open questions
+
+- Naming Convention Flexibility: How rigid should the naming convention for external
+networks be across various SCS clouds?
+- Load Balancing: Do we want to dictate a Load Balancer or a set of Load Balancers or nothing at all? E.g. Octavia, Yawol
```
josephineSei, line 67: We should keep a focus and discuss this important question in another issue.

josephineSei (commented):
There are many things discussed in thins standard. As it will hopefully lead to an architecture we standardize, mentioning all networking things is appropiate. But we should keep a clear focus and delegate anything that has nothing to do with the architecture and the workflow of getting a floating IPv4 (or can be a simple separate topic) to other issues.

markus-hentsch:
## Port Security and Shared Networks

@cah-patrickthiem, @josephineSei in the context of our discussion today about port security and security groups I had a closer look at this on a DevStack and found some interesting things.

Context recap: we consider disabled port security on ports within a network shared by multiple tenants a threat because then malicious tenants can spoof their addresses to be that of other tenants in the network. With port security enabled, OpenStack makes sure that only traffic from/to known assigned addresses is allowed. Also port security controls whether security group rules are applied or not (enabled port security = security groups enabled).

### Changing port security on admin side:

- an admin can create a shared or public/external[^1] network with `--disable-port-security` which disables port security for all ports created in this network per default
  - if tenants create VMs within such network while *not* specifying security groups, **no default security group** will be assigned to the VM
  - if tenants create VMs within such network while specifying at least one security group, the server will reach ERROR state but the message about it not being able to assign the security group because of the network enforcing disabled port security seems to be only visible in Nova's logs (see the bottom of this post), not in the server status

We should add "Port security MUST NOT be disabled for networks created as `shared` (`--share` on CLI) or with `router:external` (`--external` on CLI)" or something similar to the standard. 
*Note: how we phrase this depends on how the glossary turns out, i.e. whether we state the API attributes here (`shared`, `router:external`) or if we reference well-defined network classification terminology established in the glossary.*

[^1]: public/external as in `--external` on CLI or `router:external` in the API.

### Changing port security on user side:

- if a tenant uses `openstack port set --disable-port-security ...` on a port in a shared or public/external[^1] network, the default Neutron API policies will deny this request:
    ```
    ForbiddenException: 403: Client Error for url: $NEUTRON:9696/networking/v2.0/ports/b4a5ea84-6084-43dd-a5bc-8b8bf40cba0f, (rule:update_port and rule:update_port:port_security_enabled) is disallowed by policy
    ```
    - the default policy[^2] restricts this to admins or users of the same project the network was created in

Since in the default policy creation of shared or external networks is an admin-only action, we should be safe here in most cases. However, since admins *can* create networks in specific projects (e.g. `openstack network create --project ...`) in conjunction with the external or shared flag, this might introduce loopholes.

Therefore, I think we should also add "Networks marked as shared or external (`router:external` attribute) MUST NOT be created within a tenant's project" or something. Again, phrasing should be in line with the terminology established.

[^2]: https://docs.openstack.org/neutron/latest/configuration/policy.html

---

Here is the nova-compute log entry when a tenant tries to create a VM with a security group assigned in a network where the admin has disabled port security:

```
nova.exception.SecurityGroupCannotBeApplied: Network requires port_security_enabled and subnet associated in order to apply security groups
```

... and this is what the user sees in `openstack server show`:

```
| fault                               | {'code': 500, 'created': '2024-04-09T11:45:40Z', 'message':        |
|                                     | 'Exceeded maximum number of retries. Exhausted all hosts available |
|                                     | for retrying build failures for instance                           |
|                                     | 4dd38355-7bdf-4c84-b07f-1516c3ae0925.'}                            |
| status                              | ERROR                                                              |
```

cah-patrickthiem:
Current status of this topic & some open questions:

- this standard will be a rather more general standard, with important topics directly associated with IPv4 handling and linking to other standards for more in depth explanations (e.g. Security Group standard)

Key suggestions:
- IPv6 MUST be available
- subnet pools SHOULD be used for IPv6
- dual stack MUST be available for IPv6 and SHOULD be for IPv4
- Border Gateway Protocol (BGP): IPv6 MUST - IPv4 MAY - SHOULD
- Neutron Routers MUST be used
- OVN or L3agent SHOULD be used as high availability service deployments
- Standard external networks MUST NOT be made accessible as shared networks
   - needs further investigation regarding "port security"
- External networks and subnets MUST be configured with --no-dhcp
- SCS conform CSPs MAY use RBAC and VPNaaS Neutron plugins
   - need to investigate further on what plugins & extensions are really available and up to date
   - RBAC MUST set to admin only mode
- security Groups SHOULD be enabled by default
- standard quota of floating IPs and routers MAY be rather small
   - min quota SHOULD be > 1 floating IP
   - same for router usage
- SCS CSPs SHOULD implement monitoring solutions to track the utilization of IPv4 floating IPs
   - probably makes more sense to specify the tool from the SCS side which should be used
      - we would need to build it ourselves probably
- SCS clouds SHOULD adopt the naming convention scs-external-net for external 
      - **we decided to drop that entirely**
- Floating IPs MUST be enabled
- multiple subnets MAY be used in the standard external network

Open Questions:
- Should IPv4 and IPv6 networks be standardized in one document?

Todos:
- check port security
- investigate on Neutron router plugins and extensions
- create architectural overview regarding IPv4 networking in SCS

Standards/scs-xxxx-v1-ipv4-networking-standard.md
kgube: Just to add some background on this point: this is about [high availability of virtual routers](https://access.redhat.com/documentation/de-de/red_hat_openstack_platform/9/html/networking_guide/sec-l3-ha), i.e. replicating virtual routers across multiple network nodes and then allowing failover via VRRP in case of a node failure, a feature that is supported by both virtual router implementations included in neutron (OVN and L3agent), but may be mutually exclusive with distributed virtual routing (DVR, the implementation of virtual routers on the compute nodes).

So this is not really an endorsement of OVN or L3agent, it is just that those are the two available implementations. There might be proprietary service plugins to replace them, but every driver that is not OVN seems to just use L3agent.
 
This feature is also invisible to tenants, and I'm not sure if it should be part of this specific standard. We should probably have a different discussion about where and how to mandate HA features, maybe in the context of #527.

Standards/scs-xxxx-v1-ipv4-networking-standard.md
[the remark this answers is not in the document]
kgube: Some background for this as well: Neutron used to include a number of monolithic core plugins, but they have all been converted to drivers for the the ML2 core plugin, as well as service plugins that build on top of that.

The available service plugins are defined in neutrons [setup.cfg](https://github.com/openstack/neutron/blob/master/setup.cfg#L71), and CSPs can also configure additional external plugins there. Each plugin implementation has to declare the API extensions that it provides, most of which are defined in [neutron-lib](https://github.com/openstack/neutron-lib/tree/master/neutron_lib/api/definitions). The most prominent ones are `router` and `router-ovn` which implement virtual routers and floating IPs for l3agent and ovn, respectively.

A number of API extensions are also implemented by the ML2 core plugin itself, such as subnetpools, security groups, and the different rbac extensions. As such there is no separate plugin for RBAC.
There used to be a VPNaaS plugin in neutron but it has been removed at some point, though the definition for the API extension still exists in neutron-lib.

Standards/scs-xxxx-v1-ipv4-networking-standard.md
kgube: Neutrons RBAC extension allows sharing of various resources among projects and is the mechanism that neutron uses to implement provider networks, so it's not really an optional feature. Rules can be created with the actions `access_as_shared` and `access_as_external` and a target_project_id that can be either an actual project id or `*` for sharing with all projects.
To my knowledge it does not require any explicit extra configuration, as it is part of the ML2 core plugin.

Neutrons default policies allow only the owner of a resource to share it, and allows only admin users to create rules with the wildcard `*` target.
This would still allow a malicius non-admin user to create a faux external network and offer it to selected target projects for whom it might not be immediately obvious that this network was not provided by the CSP (`openstack network list` does not even show the project id of the networks per default).
This may give the attacker the opportunity to intercept external traffic of the target project.
Preventing non-admin users from sharing networks is at least something we should consider, especially if we do not have a naming convention.

cah-patrickthiem:
For the record, that topic was discussed recently in a SCS IaaS meeting, initiated by kgube.
One of the questions was, if we want to define all network topics in one single network standard and the answer was clearly a no.
Next we will figure out what network aspect will get its own standard document and what aspect might still be missing.

PR #524, opened 2024-03-15
Create v2 of node distribution standard (issues/#494)

cah-hbaum:
Adds the new label topology.scs.openstack.org/host-id to the standard and extend the standard to require providers to set the labels on their managed k8s clusters.

### DIFF
--- a/Standards/scs-0214-v1-k8s-node-distribution.md
+++ b/Standards/scs-0214-v1-k8s-node-distribution.md
@@ -80,15 +80,45 @@ If the standard is used by a provider, the following decisions are binding and v
   can also be scaled vertically first before scaling horizontally.
 - Worker node distribution MUST be indicated to the user through some kind of labeling
   in order to enable (anti)-affinity for workloads over "failure zones".
+- To provide metadata about the node distribution, which also enables testing of this standard,
+  providers MUST label their K8s nodes with the labels listed below.
+  - "topology.kubernetes.io/zone"
+
+    Corresponds with the label described in [K8s labels documentation][k8s-labels-docs].
+    It provides a logical zone of failure on the side of the provider, e.g. a server rack
+    in the same electrical circuit or multiple machines bound to the internet through a
+    singular network structure. How this is defined exactly is up to the plans of the provider.
+    The field gets autopopulated most of the time by either the kubelet or external mechanisms
+    like the cloud controller.
+
+  - "topology.kubernetes.io/region"
+
+    Corresponds with the label described in [K8s labels documentation][k8s-labels-docs].
+    It describes the combination of one or more failure zones into a region or domain, therefore
+    showing a larger entity of logical failure zone. An example for this could be a building
+    containing racks that are put into such a zone, since they're all prone to failure, if e.g.
+    the power for the building is cut. How this is defined exactly is also up to the provider.
+    The field gets autopopulated most of the time by either the kubelet or external mechanisms
+    like the cloud controller.
+
+  - "topology.scs.community/host-id"
+
+    This is an SCS-specific label, which MUST contain the hostID of the physical machine running
+    the hypervisor and not the hostID of a virtual machine. The hostID is an arbitrary identifier,
+    which doesn't need to contain things like hostname, but it should nonetheless be unique to the host.
+    This helps identify the distribution over underlying physical machines,
+    which would be masked if VM hostIDs would be used.
 
 ## Conformance Tests
 
 The script `k8s-node-distribution-check.py` checks the nodes available with a user-provided
-kubeconfig file. It then determines based on the labels `kubernetes.io/hostname`, `topology.kubernetes.io/zone`,
-`topology.kubernetes.io/region` and `node-role.kubernetes.io/control-plane`, if a distribution
-of the available nodes is present. If this isn't the case, the script produces an error.
+kubeconfig file. Based on the labels `topology.scs.community/host-id`,
+`topology.kubernetes.io/zone`, `topology.kubernetes.io/region` and `node-role.kubernetes.io/control-plane`,
+the script then determines whether the nodes are distributed according to this standard.
+If this isn't the case, the script produces an error.
 If also produces warnings and informational outputs, if e.g. labels don't seem to be set.
 
 [k8s-ha]: https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/high-availability/
 [k8s-large-clusters]: https://kubernetes.io/docs/setup/best-practices/cluster-large/
 [scs-0213-v1]: https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
+[k8s-labels-docs]: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone
--- a/Standards/scs-0214-v1-k8s-node-distribution.md
+++ b/Standards/scs-0214-v1-k8s-node-distribution.md
@@ -80,6 +80,34 @@ If the standard is used by a provider, the following decisions are binding and v
   can also be scaled vertically first before scaling horizontally.
 - Worker node distribution MUST be indicated to the user through some kind of labeling
   in order to enable (anti)-affinity for workloads over "failure zones".
+- To provide metadata about the node distribution, which also enables testing of this standard,
+  providers MUST label their K8s nodes with the labels listed below.
+  - `topology.kubernetes.io/zone`
+
+    Corresponds with the label described in [K8s labels documentation][k8s-labels-docs].
+    It provides a logical zone of failure on the side of the provider, e.g. a server rack
+    in the same electrical circuit or multiple machines bound to the internet through a
+    singular network structure. How this is defined exactly is up to the plans of the provider.
+    The field gets autopopulated most of the time by either the kubelet or external mechanisms
+    like the cloud controller.
+
+  - `topology.kubernetes.io/region`
+
+    Corresponds with the label described in [K8s labels documentation][k8s-labels-docs].
+    It describes the combination of one or more failure zones into a region or domain, therefore
+    showing a larger entity of logical failure zone. An example for this could be a building
+    containing racks that are put into such a zone, since they're all prone to failure, if e.g.
+    the power for the building is cut. How this is defined exactly is also up to the provider.
+    The field gets autopopulated most of the time by either the kubelet or external mechanisms
+    like the cloud controller.
+
+  - `topology.scs.community/host-id`
+
+    This is an SCS-specific label, which MUST contain the hostID of the physical machine running
+    the hypervisor and not the hostID of a virtual machine. Here, the hostID is an arbitrary identifier,
+    which need not to contain the actual hostname, but it should nonetheless be unique to the host.
+    This helps identify the distribution over underlying physical machines,
+    which would be masked if VM hostIDs were used.
 
 ## Conformance Tests
 
--- a/Standards/scs-0214-v1-k8s-node-distribution.md
+++ b/Standards/scs-0214-v1-k8s-node-distribution.md
@@ -103,9 +103,9 @@ If the standard is used by a provider, the following decisions are binding and v
 
   - `topology.scs.community/host-id`
 
-    This is an SCS-specific label, which MUST contain the hostID of the physical machine running
-    the hypervisor and not the hostID of a virtual machine. Here, the hostID is an arbitrary identifier,
-    which need not to contain the actual hostname, but it should nonetheless be unique to the host.
+    This is an SCS-specific label; it MUST contain the hostID of the physical machine running
+    the hypervisor (NOT: the hostID of a virtual machine). Here, the hostID is an arbitrary identifier,
+    which need not contain the actual hostname, but it should nonetheless be unique to the host.
     This helps identify the distribution over underlying physical machines,
     which would be masked if VM hostIDs were used.
 
--- a/Standards/scs-0214-v1-k8s-node-distribution.md
+++ b/Standards/scs-0214-v1-k8s-node-distribution.md
@@ -120,3 +120,4 @@ If also produces warnings and informational outputs, if e.g. labels don't seem t
 [k8s-ha]: https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/high-availability/
 [k8s-large-clusters]: https://kubernetes.io/docs/setup/best-practices/cluster-large/
 [scs-0213-v1]: https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
+[k8s-labels-docs]: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone
--- a/Standards/scs-0214-v2-k8s-node-distribution.md
+++ b/Standards/scs-0214-v2-k8s-node-distribution.md
@@ -2,7 +2,8 @@
 title: Kubernetes Node Distribution and Availability
 type: Standard
 status: Stable
-stabilized_at: 2024-03-21
+stabilized_at: 2024-04-04
+replaces: scs-0214-v1-k8s-node-distribution.md
 track: KaaS
 ---
 
--- a/Standards/scs-0214-v2-k8s-node-distribution.md
+++ b/Standards/scs-0214-v2-k8s-node-distribution.md
@@ -1,8 +1,7 @@
 ---
 title: Kubernetes Node Distribution and Availability
 type: Standard
-status: Stable
-stabilized_at: 2024-04-04
+status: Draft
 replaces: scs-0214-v1-k8s-node-distribution.md
 track: KaaS
 ---
--- a/Standards/scs-0214-v2-k8s-node-distribution.md
+++ b/Standards/scs-0214-v2-k8s-node-distribution.md
@@ -50,7 +50,7 @@ like fault-tolerance and data redundancy. But it also understands the costs and 
 for the providers associated with this effort, since the infrastructure needs to have
 hardware which will just be used to provide fail-over safety or duplication.
 
-The document [Best practices for large clusters] describes the concept of a failure zone.
+The document [Best practices for large clusters][k8s-large-clusters] describes the concept of a failure zone.
 This term isn't defined any further, but can in this context be described as a number of
 physical (computing) machines in such a vicinity to each other (either through physical
 or logical interconnection in some way), that specific problems inside this zone would put
@@ -82,8 +82,7 @@ If the standard is used by a provider, the following decisions are binding and v
 - Worker node distribution MUST be indicated to the user through some kind of labeling
   in order to enable (anti)-affinity for workloads over "failure zones".
 - To provide metadata about the node distribution, which also enables testing of this standard,
-  providers MUST annotate their K8s nodes with the labels listed below. These labels MUST be kept
-  up to date with the current state of the deployment.
+  providers MUST annotate their K8s nodes with the labels listed below.
   - `topology.kubernetes.io/zone`
 
     Corresponds with the label described in [K8s labels documentation][k8s-labels-docs].
--- a/Standards/scs-0214-v2-k8s-node-distribution.md
+++ b/Standards/scs-0214-v2-k8s-node-distribution.md
@@ -117,14 +117,14 @@ kubeconfig file. Based on the labels `topology.scs.community/host-id`,
 `topology.kubernetes.io/zone`, `topology.kubernetes.io/region` and `node-role.kubernetes.io/control-plane`,
 the script then determines whether the nodes are distributed according to this standard.
 If this isn't the case, the script produces an error.
-If also produces warnings and informational outputs, if e.g. labels don't seem to be set.
+It also produces warnings and informational outputs, e.g., if labels don't seem to be set.
+
+## Previous standard versions
+
+This is version 2 of the standard; it extends [version 1](scs-0214-v1-k8s-node-distribution.md) with the
+requirements regarding node labeling.
 
 [k8s-ha]: https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/high-availability/
 [k8s-large-clusters]: https://kubernetes.io/docs/setup/best-practices/cluster-large/
 [scs-0213-v1]: https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
 [k8s-labels-docs]: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone
-
-## Previous standard versions
-
-This is version 2 of the standard; it extends [version 1](scs-0214-v1-k8s-node-distribution.md) by the
-requirements regarding node labeling.
--- a/Standards/scs-0214-v2-k8s-node-distribution.md
+++ b/Standards/scs-0214-v2-k8s-node-distribution.md
@@ -124,3 +124,8 @@ If also produces warnings and informational outputs, if e.g. labels don't seem t
 [k8s-large-clusters]: https://kubernetes.io/docs/setup/best-practices/cluster-large/
 [scs-0213-v1]: https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
 [k8s-labels-docs]: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone
+
+## Previous standard versions
+
+This is version 2 of the standard; it extends [version 1](scs-0214-v1-k8s-node-distribution.md) by the
+requirements regarding node labeling.
--- a/Standards/scs-0214-v2-k8s-node-distribution.md
+++ b/Standards/scs-0214-v2-k8s-node-distribution.md
@@ -2,7 +2,7 @@
 title: Kubernetes Node Distribution and Availability
 type: Standard
 status: Stable
-stabilized_at: 2024-02-08
+stabilized_at: 2024-03-21
 track: KaaS
 ---
 
@@ -49,7 +49,7 @@ like fault-tolerance and data redundancy. But it also understands the costs and 
 for the providers associated with this effort, since the infrastructure needs to have
 hardware which will just be used to provide fail-over safety or duplication.
 
-The document [Best practices for large clusters][k8s-large-clusters] describes the concept of a failure zone.
+The document [Best practices for large clusters] describes the concept of a failure zone.
 This term isn't defined any further, but can in this context be described as a number of
 physical (computing) machines in such a vicinity to each other (either through physical
 or logical interconnection in some way), that specific problems inside this zone would put
@@ -80,15 +80,46 @@ If the standard is used by a provider, the following decisions are binding and v
   can also be scaled vertically first before scaling horizontally.
 - Worker node distribution MUST be indicated to the user through some kind of labeling
   in order to enable (anti)-affinity for workloads over "failure zones".
+- To provide metadata about the node distribution, which also enables testing of this standard,
+  providers MUST annotate their K8s nodes with the labels listed below. These labels MUST be kept
+  up to date with the current state of the deployment.
+  - `topology.kubernetes.io/zone`
+
+    Corresponds with the label described in [K8s labels documentation][k8s-labels-docs].
+    It provides a logical zone of failure on the side of the provider, e.g. a server rack
+    in the same electrical circuit or multiple machines bound to the internet through a
+    singular network structure. How this is defined exactly is up to the plans of the provider.
+    The field gets autopopulated most of the time by either the kubelet or external mechanisms
+    like the cloud controller.
+
+  - `topology.kubernetes.io/region`
+
+    Corresponds with the label described in [K8s labels documentation][k8s-labels-docs].
+    It describes the combination of one or more failure zones into a region or domain, therefore
+    showing a larger entity of logical failure zone. An example for this could be a building
+    containing racks that are put into such a zone, since they're all prone to failure, if e.g.
+    the power for the building is cut. How this is defined exactly is also up to the provider.
+    The field gets autopopulated most of the time by either the kubelet or external mechanisms
+    like the cloud controller.
+
+  - `topology.scs.community/host-id`
+
+    This is an SCS-specific label; it MUST contain the hostID of the physical machine running
+    the hypervisor (NOT: the hostID of a virtual machine). Here, the hostID is an arbitrary identifier,
+    which need not contain the actual hostname, but it should nonetheless be unique to the host.
+    This helps identify the distribution over underlying physical machines,
+    which would be masked if VM hostIDs were used.
 
 ## Conformance Tests
 
 The script `k8s-node-distribution-check.py` checks the nodes available with a user-provided
-kubeconfig file. It then determines based on the labels `kubernetes.io/hostname`, `topology.kubernetes.io/zone`,
-`topology.kubernetes.io/region` and `node-role.kubernetes.io/control-plane`, if a distribution
-of the available nodes is present. If this isn't the case, the script produces an error.
+kubeconfig file. Based on the labels `topology.scs.community/host-id`,
+`topology.kubernetes.io/zone`, `topology.kubernetes.io/region` and `node-role.kubernetes.io/control-plane`,
+the script then determines whether the nodes are distributed according to this standard.
+If this isn't the case, the script produces an error.
 If also produces warnings and informational outputs, if e.g. labels don't seem to be set.
 
 [k8s-ha]: https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/high-availability/
 [k8s-large-clusters]: https://kubernetes.io/docs/setup/best-practices/cluster-large/
 [scs-0213-v1]: https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
+[k8s-labels-docs]: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone
--- a/Standards/scs-0214-v2-k8s-node-distribution.md
+++ b/Standards/scs-0214-v2-k8s-node-distribution.md
@@ -82,7 +82,8 @@ If the standard is used by a provider, the following decisions are binding and v
 - Worker node distribution MUST be indicated to the user through some kind of labeling
   in order to enable (anti)-affinity for workloads over "failure zones".
 - To provide metadata about the node distribution, which also enables testing of this standard,
-  providers MUST annotate their K8s nodes with the labels listed below.
+  providers SHOULD annotate their K8s nodes with the labels listed below, and the labels
+  SHOULD be kept up to date.
   - `topology.kubernetes.io/zone`
 
     Corresponds with the label described in [K8s labels documentation][k8s-labels-docs].
--- a/Standards/scs-0214-v2-k8s-node-distribution.md
+++ b/Standards/scs-0214-v2-k8s-node-distribution.md
@@ -65,50 +65,48 @@ shouldn't be defined further in this document.
 
 This standard formulates the requirement for the distribution of Kubernetes nodes in order
 to provide a fault-tolerant and available Kubernetes cluster infrastructure.
-Since some providers only have small environments to work with and therefore couldn't
-comply with this standard, it will be treated as a RECOMMENDED standard, where providers
-can OPT OUT.
 
-If the standard is used by a provider, the following decisions are binding and valid:
+The control plane nodes MUST be distributed over multiple physical machines.
+Kubernetes provides [best-practices][k8s-zones] on this topic, which are also RECOMMENDED by SCS.
 
-- The control plane nodes MUST be distributed over multiple physical machines. Kubernetes
-  provides best-practices on this topic, which are also RECOMMENDED by SCS.
-- At least one control plane instance MUST be run in each "failure zone", more are
-  RECOMMENDED in each "failure zone" to provide fault-tolerance for each zone.
-- Worker nodes are RECOMMENDED to be distributed over multiple zones. This policy makes
-  it OPTIONAL to provide a worker node in each "failure zone", meaning that worker nodes
-  can also be scaled vertically first before scaling horizontally.
-- Worker node distribution MUST be indicated to the user through some kind of labeling
-  in order to enable (anti)-affinity for workloads over "failure zones".
-- To provide metadata about the node distribution, which also enables testing of this standard,
-  providers MUST annotate their K8s nodes with the labels listed below, and the labels
-  MUST be kept up to date.
-  - `topology.kubernetes.io/zone`
+At least one control plane instance MUST be run in each "failure zone" used for the cluster,
+more instances per "failure zone" are possible to provide fault-tolerance inside a zone.
 
-    Corresponds with the label described in [K8s labels documentation][k8s-labels-docs].
-    It provides a logical zone of failure on the side of the provider, e.g. a server rack
-    in the same electrical circuit or multiple machines bound to the internet through a
-    singular network structure. How this is defined exactly is up to the plans of the provider.
-    The field gets autopopulated most of the time by either the kubelet or external mechanisms
-    like the cloud controller.
+Worker nodes are RECOMMENDED to be distributed over multiple zones. This policy makes
+it OPTIONAL to provide a worker node in each "failure zone", meaning that worker nodes
+can also be scaled vertically first before scaling horizontally.
 
-  - `topology.kubernetes.io/region`
+To provide metadata about the node distribution and possibly provide the ability
+to schedule workloads efficiently, which also enables testing of this standard,
+providers MUST annotate their K8s nodes with the labels listed below.
+These labels MUST be kept up to date with the current state of the deployment.
 
-    Corresponds with the label described in [K8s labels documentation][k8s-labels-docs].
-    It describes the combination of one or more failure zones into a region or domain, therefore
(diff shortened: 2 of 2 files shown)


Standards/scs-0214-v1-k8s-node-distribution.md
```diff
@@ -80,15 +80,27 @@ If the standard is used by a provider, the following decisions are binding and v
...
+  providers MUST label their K8s nodes with the labels "topology.kubernetes.io/region",
+  "topology.kubernetes.io/zone" and "topology.scs.openstack.org/host-id"
+  in order to provide the expected data on the KaaS layer.
+  "topology.kubernetes.io/region" and "topology.kubernetes.io/zone" correspond with the labels
+  described in the [K8s labels documentation][k8s-labels-docs] and provide logical zones of failure
+  as well as a domain made up of one or more zones. How these are defined is up to the provider and
+  the fields are most of the time autopopulated by either the kubelet, the cloud controller or another
+  external mechanism.
+  "topology.scs.openstack.org/host-id" is an SCS-specific label, which MUST contain the hostID of the
+  physical machine running the hypervisor and not the hostID of a virtual machine.
```
mbuechse, line 93: Three things:

- I would prefer more structure (use a list for the labels)
- then state for each label what the value should be (even though it's not authoritative for region and zone)
- maybe don't narrow it down to verifying compliance so much; it's also desirable in its own right for maximum transparency (and it seems t be quite common already, maybe for this very reason)

Standards/scs-0214-v1-k8s-node-distribution.md
```diff
@@ -80,15 +80,27 @@ If the standard is used by a provider, the following decisions are binding and v
   can also be scaled vertically first before scaling horizontally.
 - Worker node distribution MUST be indicated to the user through some kind of labeling
   in order to enable (anti)-affinity for workloads over "failure zones".
+- To provide data about this node distribution and be able to test and verify compliance with this standard,
+  providers MUST label their K8s nodes with the labels "topology.kubernetes.io/region",
+  "topology.kubernetes.io/zone" and "topology.scs.openstack.org/host-id"
```
matofeder, line 85: [Here](https://github.com/SovereignCloudStack/issues/issues/540#issuecomment-1997330524), @martinmo proposed changing the label name to `topology.scs.community/host-id`

As I do not have a strong opinion here, I just changed this name in the related e2e test https://github.com/SovereignCloudStack/k8s-cluster-api-provider/commit/ffb722b42e99977d6cb42817993e11558d816012

So use the same name or raise your voice in the related issue if you do not agree with the latest proposal.

Standards/scs-0214-v1-k8s-node-distribution.md
```diff
@@ -80,15 +80,27 @@ If the standard is used by a provider, the following decisions are binding and v
...
+  in order to provide the expected data on the KaaS layer.
+  "topology.kubernetes.io/region" and "topology.kubernetes.io/zone" correspond with the labels
+  described in the [K8s labels documentation][k8s-labels-docs] and provide logical zones of failure
+  as well as a domain made up of one or more zones. How these are defined is up to the provider and
+  the fields are most of the time autopopulated by either the kubelet, the cloud controller or another
+  external mechanism.
+  "topology.scs.openstack.org/host-id" is an SCS-specific label, which MUST contain the hostID of the
```
matofeder, line 92: adjust the label name based on the comment above

Standards/scs-0214-v1-k8s-node-distribution.md
mbuechse: I agree. I think so does @cah-hbaum, but I don't know for sure.

cah-hbaum: Also agree, I only had this information later on. I'm gonna change it with the next commit.

Standards/scs-0214-v1-k8s-node-distribution.md
mbuechse: Oh sorry, I didn't mean concrete values, just an explanation what kind of information we expect, namely the zone and the region -- you know, this may sound redundant, but I want to be extra clear that the label is what the label is :)

cah-hbaum: Ah ok got it

matofeder (approved):
LGTM

Standards/scs-0214-v1-k8s-node-distribution.md
```diff
@@ -80,15 +80,39 @@ If the standard is used by a provider, the following decisions are binding and v
...
+    the hypervisor and not the hostID of a virtual machine. This helps identify the distribution
+    over underlying physical machines, which would be masked if VM hostIDs would be used.
 
 ## Conformance Tests
 
 The script `k8s-node-distribution-check.py` checks the nodes available with a user-provided
-kubeconfig file. It then determines based on the labels `kubernetes.io/hostname`, `topology.kubernetes.io/zone`,
-`topology.kubernetes.io/region` and `node-role.kubernetes.io/control-plane`, if a distribution
-of the available nodes is present. If this isn't the case, the script produces an error.
+kubeconfig file. It then determines based on the labels `topology.scs.openstack.org/host-id`,
```
matofeder, line 110:
```suggestion
kubeconfig file. It then determines based on the labels `topology.scs.community/host-id,
```

cah-hbaum: Oh yeah I forgot that. Fixed it now.

Standards/scs-0214-v1-k8s-node-distribution.md
```diff
@@ -80,15 +80,42 @@ If the standard is used by a provider, the following decisions are binding and v
...
+    containing racks that are put into such a zone, since they're all prone to failure, if e.g.
+    the power for the building is cut. How this is defined exactly is also up to the provider.
+    The field gets autopopulated most of the time by either the kubelet or external mechanisms
+    like the cloud controller.
+
+  - "topology.scs.community/host-id"
+
+    This is an SCS-specific label, which MUST contain the hostID of the physical machine running
+    the hypervisor and not the hostID of a virtual machine. This helps identify the distribution
+    over underlying physical machines, which would be masked if VM hostIDs would be used.
```
mbuechse, line 108: I like the new version very much! One minor thing with this paragraph: we have to say what we mean by `hostID`. We should say that this is more or less an arbitrary identifier for the host (need not be identical with the hostname or anything tangible), but it must be unique to the host so it is clear whether two nodes are placed on the same host.

Standards/scs-0214-v1-k8s-node-distribution.md
```diff
@@ -80,15 +80,42 @@ If the standard is used by a provider, the following decisions are binding and v
...
 
 ## Conformance Tests
 
 The script `k8s-node-distribution-check.py` checks the nodes available with a user-provided
-kubeconfig file. It then determines based on the labels `kubernetes.io/hostname`, `topology.kubernetes.io/zone`,
-`topology.kubernetes.io/region` and `node-role.kubernetes.io/control-plane`, if a distribution
-of the available nodes is present. If this isn't the case, the script produces an error.
+kubeconfig file. It then determines based on the labels `topology.scs.community/host-id`,
+`topology.kubernetes.io/zone`, `topology.kubernetes.io/region` and `node-role.kubernetes.io/control-plane`,
+if a distribution of the available nodes is present. If this isn't the case, the script produces an error.
```
mbuechse, line 115:
```suggestion
kubeconfig file. Based on the labels `topology.scs.community/host-id`,
`topology.kubernetes.io/zone`, `topology.kubernetes.io/region` and `node-role.kubernetes.io/control-plane`,
the script then determines whether the nodes are distributed according to this standard.
If this isn't the case, the script produces an error.
```

Standards/scs-0214-v1-k8s-node-distribution.md
```diff
@@ -80,15 +80,45 @@ If the standard is used by a provider, the following decisions are binding and v
...
+    containing racks that are put into such a zone, since they're all prone to failure, if e.g.
+    the power for the building is cut. How this is defined exactly is also up to the provider.
+    The field gets autopopulated most of the time by either the kubelet or external mechanisms
+    like the cloud controller.
+
+  - "topology.scs.community/host-id"
+
+    This is an SCS-specific label, which MUST contain the hostID of the physical machine running
+    the hypervisor and not the hostID of a virtual machine. The hostID is an arbitrary identifier,
+    which doesn't need to contain things like hostname, but it should nonetheless be unique to the host.
```
mbuechse, line 108:
```suggestion
    which need not contain the actual hostname, but it should nonetheless be unique to the host.
```

mbuechse:
Let Team Container in its next meeting vote on whether this needs a v2; apart from that: good to go!

mbuechse:
@cah-hbaum you are quick! Let's add a paragraph that explains the changes vs v1.

martinmo (commented):
I agree, this was really quick! Very good. I have left some review comments we should discuss about, otherwise LGTM.

Standards/scs-0214-v2-k8s-node-distribution.md
```diff
@@ -0,0 +1,124 @@
...
+    The field gets autopopulated most of the time by either the kubelet or external mechanisms
+    like the cloud controller.
+
+  - `topology.scs.community/host-id`
+
+    This is an SCS-specific label; it MUST contain the hostID of the physical machine running
+    the hypervisor (NOT: the hostID of a virtual machine). Here, the hostID is an arbitrary identifier,
+    which doesn't need to contain the actual hostname, but it should nonetheless be unique to the host.
+    This helps identify the distribution over underlying physical machines,
+    which would be masked if VM hostIDs were used.
```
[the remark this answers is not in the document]
mbuechse: - Probably unnecessary. I think it's clear that the hostID is a property of the physical host, so each physical host has exactly one such id. However, if we find a non-awkward way of putting it, we can still do so.
- This is beyond my depth. Maybe @matofeder can say something about this? What happens when a k8s node is migrated to another physical host?

Standards/scs-0214-v2-k8s-node-distribution.md
```diff
@@ -0,0 +1,124 @@
...
+Most design considerations of this standard follow the previously written Decision Record
+[Kubernetes Nodes Anti Affinity][scs-0213-v1] as well as the Kubernetes documents about
+[High Availability][k8s-ha] and [Best practices for large clusters][k8s-large-clusters].
+
+SCS wishes to prefer distributed, highly-available systems due to their obvious advantages
+like fault-tolerance and data redundancy. But it also understands the costs and overhead
+for the providers associated with this effort, since the infrastructure needs to have
+hardware which will just be used to provide fail-over safety or duplication.
+
+The document [Best practices for large clusters] describes the concept of a failure zone.
```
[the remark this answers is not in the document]
mbuechse: 92a683c7e7903c5c934ce0802c3945d72a9503e9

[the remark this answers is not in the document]
mbuechse: 5c263411f921ea22d2c11432ea24210a7e53cd13

Standards/scs-0214-v2-k8s-node-distribution.md
```diff
@@ -0,0 +1,124 @@
...
+  provides best-practices on this topic, which are also RECOMMENDED by SCS.
+- At least one control plane instance MUST be run in each "failure zone", more are
+  RECOMMENDED in each "failure zone" to provide fault-tolerance for each zone.
+- Worker nodes are RECOMMENDED to be distributed over multiple zones. This policy makes
+  it OPTIONAL to provide a worker node in each "failure zone", meaning that worker nodes
+  can also be scaled vertically first before scaling horizontally.
+- Worker node distribution MUST be indicated to the user through some kind of labeling
+  in order to enable (anti)-affinity for workloads over "failure zones".
+- To provide metadata about the node distribution, which also enables testing of this standard,
+  providers MUST label their K8s nodes with the labels listed below.
```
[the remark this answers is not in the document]
mbuechse: 5c263411f921ea22d2c11432ea24210a7e53cd13

mbuechse:
@cah-hbaum @martinmo @matofeder Please be careful before merging. This document will be merged in stable state, so let's be extra certain that it is good this time... I pushed the date to April 4, so there's no need to hurry.

Standards/scs-0214-v2-k8s-node-distribution.md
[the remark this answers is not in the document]
matofeder: > Should we explicitely say something about what should happen if workload (e.g., an OpenStack VM) is migrated to another physical host? My expectation is that the host-id label must be updated.


IMO we could extend this question also with `zone` and `region` labels (AFAIK live migration is also possible across zones, and potentially some cold migration could be possible even across regions).

Currently, the approach used by KaaS V1/V2 - [CAPO](https://cluster-api-openstack.sigs.k8s.io/) reconciles 
`zone` and `region` labels, so potentially `zone` and `region` labels will be updated in case of migration. 
The `host-id` label is not covered by the current approach at all. Because of that, we introduced a custom pre-scs-standard-test task in the KaaS E2E test that does the `host-id` labeling on the fly, i.e. this `host-id` labeling workaround works also in case of hot/cold migration. 

But IMO SCS standards should be generic, and not CAPO specific, therefore I would say that the CSP SHOULD ensure that the REQUIRED labels are present and CSP SHOULD ensure that the REQUIRED labels are updated in case of hot or cold migrations. 

Currently, we do not have a production tool that ensures the above (especially for the `host-id` label), just the e2e workaround I mentioned. There is some related discussion in CAPO upstream, where we could contribute, let's see. I will keep you informed.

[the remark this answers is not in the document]
mbuechse: What matters for me: if these migrations are relevant. And that seems to be the case. So I guess we should mention that the labels have to reflect the current state at all times (with some room for latency) or something. And yes, the reference implementation should do it ;)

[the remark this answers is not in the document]
mbuechse: I added language that the labels must be kept up to date. Feel free to put it more elegantly. And please ignore the intermediate version where I replaced MUST by SHOULD. That was probably a bad idea.

Standards/scs-0214-v2-k8s-node-distribution.md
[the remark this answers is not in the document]
cah-hbaum: This is written now as
```
- To provide metadata about the node distribution, which also enables testing of this standard,
  providers MUST annotate their K8s nodes with the labels listed below, and the labels
  MUST be kept up to date.
```

@martinmo if thats also ok with you, please resolve =)

Standards/scs-0214-v2-k8s-node-distribution.md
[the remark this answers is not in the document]
cah-hbaum: Looks good.

Standards/scs-0214-v1-k8s-node-distribution.md
cah-hbaum:
```
Here, the hostID is an arbitrary identifier, which doesn't need to contain the actual hostname, but it should nonetheless be unique to the host.
```
Is this version ok @mbuechse ?

Standards/scs-0214-v1-k8s-node-distribution.md
cah-hbaum:
```

File: Standards/scs-0214-v1-k8s-node-distribution.md (hunk 7 above)
Comment (mbuechse) [reply Standards/scs-0214-v1-k8s-node-distribution.md]: What's wrong with `need not contain`?

Comment (mbuechse) [reply Standards/scs-0214-v1-k8s-node-distribution.md]: https://english.stackexchange.com/questions/29409/why-use-need-not-instead-of-do-not-need-to

Comment (cah-hbaum) [reply Standards/scs-0214-v1-k8s-node-distribution.md]: I guess mainly personal preference, auxiliary verb usage doesn't sound that good to me.
But no hard preference on that.

Comment (cah-hbaum) [reply Standards/scs-0214-v1-k8s-node-distribution.md]: Updated it.

File: Standards/scs-0214-v2-k8s-node-distribution.md (hunk 10 above)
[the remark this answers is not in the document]
Comment (martinmo) [reply Standards/scs-0214-v2-k8s-node-distribution.md]: @cah-hbaum done.

**Comment by cah-hbaum** [comment]
This should be done now. After some more approvals fly in, I will merge this.

**Review by martinmo — APPROVED** [review]
Just found two typos and a cosmetic issue – apart from that, it is good to go!

File: Standards/scs-0214-v2-k8s-node-distribution.md [hunk 11]
```diff
@@ -0,0 +1,131 @@
...
+kubeconfig file. Based on the labels `topology.scs.community/host-id`,
+`topology.kubernetes.io/zone`, `topology.kubernetes.io/region` and `node-role.kubernetes.io/control-plane`,
+the script then determines whether the nodes are distributed according to this standard.
+If this isn't the case, the script produces an error.
+If also produces warnings and informational outputs, if e.g. labels don't seem to be set.
+
+[k8s-ha]: https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/high-availability/
+[k8s-large-clusters]: https://kubernetes.io/docs/setup/best-practices/cluster-large/
+[scs-0213-v1]: https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0213-v1-k8s-nodes-anti-affinity.md
+[k8s-labels-docs]: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone
```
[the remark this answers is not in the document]
cah-hbaum: Fixed with https://github.com/SovereignCloudStack/standards/pull/524/commits/addb3085a4f7456e8166f693f171197fd3652ffc

Standards/scs-0214-v2-k8s-node-distribution.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Previous standard versions
+
+This is version 2 of the standard; it extends [version 1](scs-0214-v1-k8s-node-distribution.md) by the
+requirements regarding node labeling.
```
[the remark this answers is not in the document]
cah-hbaum: Fixed with https://github.com/SovereignCloudStack/standards/pull/524/commits/addb3085a4f7456e8166f693f171197fd3652ffc

Standards/scs-0214-v2-k8s-node-distribution.md
```diff
@@ -0,0 +1,131 @@
...
+    which would be masked if VM hostIDs were used.
+
+## Conformance Tests
+
+The script `k8s-node-distribution-check.py` checks the nodes available with a user-provided
+kubeconfig file. Based on the labels `topology.scs.community/host-id`,
+`topology.kubernetes.io/zone`, `topology.kubernetes.io/region` and `node-role.kubernetes.io/control-plane`,
+the script then determines whether the nodes are distributed according to this standard.
+If this isn't the case, the script produces an error.
+If also produces warnings and informational outputs, if e.g. labels don't seem to be set.
```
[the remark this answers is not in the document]
cah-hbaum: Fixed with https://github.com/SovereignCloudStack/standards/pull/524/commits/addb3085a4f7456e8166f693f171197fd3652ffc

cah-hbaum:
I changed the document state to `Draft` now like we discussed in the Standardization/Certification meeting.

I also fixed (probably) all problems that had remained, so it should be mergable now.

cah-hbaum:
@mbuechse Please check this again, since this is in `Draft` now.

Standards/scs-0214-v2-k8s-node-distribution.md
```diff
@@ -0,0 +1,130 @@
...
+coupled by things like electrical grids. They're therefore purely logical entities, which
+shouldn't be defined further in this document.
+
+## Decision
+
+This standard formulates the requirement for the distribution of Kubernetes nodes in order
+to provide a fault-tolerant and available Kubernetes cluster infrastructure.
+Since some providers only have small environments to work with and therefore couldn't
+comply with this standard, it will be treated as a RECOMMENDED standard, where providers
+can OPT OUT.
```
mbuechse, line 70: I think I would drop this sentence. We should, however, provide implementation notes.

And we should clarify with the help of Team Container what really happens with Gardener (and the like): if there is a set of control-plane nodes that are being shared by the control planes of multiple clusters -- is it really the case that these nodes cannot be enumerated? Can we maybe get access to such a cluster (I think @berendt is working on that right now and we should get access soon)?

Standards/scs-0214-v2-k8s-node-distribution.md
```diff
@@ -0,0 +1,130 @@
...
+
+If the standard is used by a provider, the following decisions are binding and valid:
```
mbuechse, line 72: I suggest to drop this as well. Maybe even make the following list into a sequence of paragraphs.

Standards/scs-0214-v2-k8s-node-distribution.md
```diff
@@ -0,0 +1,130 @@
...
+
+- The control plane nodes MUST be distributed over multiple physical machines. Kubernetes
+  provides best-practices on this topic, which are also RECOMMENDED by SCS.
```
mbuechse, line 75: Can we add a link to the best practices?

Standards/scs-0214-v2-k8s-node-distribution.md
```diff
@@ -0,0 +1,130 @@
...
+- At least one control plane instance MUST be run in each "failure zone", more are
+  RECOMMENDED in each "failure zone" to provide fault-tolerance for each zone.
```
mbuechse, line 77: I'm not sure if we test this. We only check whether multiple zones are used, right? But we don't check whether each zone is used. But I'm also not sure how useful this assertion would be. Maybe you have 5 failure zones, but you only want 3 control-plane nodes?

Standards/scs-0214-v2-k8s-node-distribution.md
```diff
@@ -0,0 +1,130 @@
...
+- Worker nodes are RECOMMENDED to be distributed over multiple zones. This policy makes
+  it OPTIONAL to provide a worker node in each "failure zone", meaning that worker nodes
+  can also be scaled vertically first before scaling horizontally.
+- Worker node distribution MUST be indicated to the user through some kind of labeling
+  in order to enable (anti)-affinity for workloads over "failure zones".
```
mbuechse, line 82: If this is subsumed by the next item, it should probably be dropped.
Otherwise, we should refine it and state what kind of labeling we want.
And then we should add a test, I presume.

Standards/scs-0214-v2-k8s-node-distribution.md
```diff
@@ -0,0 +1,130 @@
...
+- To provide metadata about the node distribution, which also enables testing of this standard,
+  providers MUST annotate their K8s nodes with the labels listed below. These labels MUST be kept
+  up to date with the current state of the deployment.
```
mbuechse, line 85: As of now, we don't have any idea when this can be implemented with CAPO. @matofeder Do you have an update?
Unless the implemenation is in sight, we should demote this from a "MUST annotate" to a "SHOULD annotate". (The "MUST be kept up to date" can remain, so long as it's clear that it only applies if labels are being set at all.)

mbuechse (commented):
As discussed yesterday in SIG std/cert, we still want to refine a few things.

Standards/scs-0214-v2-k8s-node-distribution.md
matofeder: @chess-knight as you have better visibility of this topic now, please provide an update here, thx

chess-knight: The draft PR https://github.com/SovereignCloudStack/cluster-stacks/pull/53 with hostId labeling workaround was created by @matofeder and @SovereignCloudStack/vp06c will continue that work ASAP.
Regarding upstream, there is no progress AFAIK.

chess-knight: I created an upstream issue https://github.com/kubernetes/cloud-provider-openstack/issues/2579 so let's wait for their opinion on that.

Standards/scs-0214-v2-k8s-node-distribution.md
cah-hbaum: Changed it into a list of paragraphs.

Standards/scs-0214-v2-k8s-node-distribution.md
cah-hbaum: Sure, added the link with the document I think this was refering to (but its already a bit away, so I could be wrong).

Standards/scs-0214-v2-k8s-node-distribution.md
cah-hbaum: Maybe this was a bit misunderstanding (or I misunderstand your comment).
The standard wants that you use multiple zones for the control-plane and run at least one CP instance in each of these zones. We dont necessarily need a zone per control-plane instance and we certainly don't want to use all failure-zones available to a CSP.

I nevertheless rephrased this part a bit.

Standards/scs-0214-v2-k8s-node-distribution.md
cah-hbaum: Dropped this sentence and updated the following one a bit.

Standards/scs-0214-v2-k8s-node-distribution.md
```diff
@@ -69,46 +69,44 @@ Since some providers only have small environments to work with and therefore cou
...
+Kubernetes provides [best-practices][k8s-zones] on this topic, which are also RECOMMENDED by SCS.
+At least one control plane instance MUST be run in each "failure zone" used for the cluster,
+more instances per "failure zone" are possible to provide fault-tolerance inside a zone.
+Worker nodes are RECOMMENDED to be distributed over multiple zones. This policy makes
+it OPTIONAL to provide a worker node in each "failure zone", meaning that worker nodes
+can also be scaled vertically first before scaling horizontally.
+To provide metadata about the node distribution and possibly provide the ability
+to schedule workloads efficiently, which also enables testing of this standard,
+providers MUST annotate their K8s nodes with the labels listed below.
+These labels MUST be kept up to date with the current state of the deployment.
```
mbuechse, line 82:
```suggestion
The control plane nodes MUST be distributed over multiple physical machines.
Kubernetes provides [best-practices][k8s-zones] on this topic, which are also RECOMMENDED by SCS.

At least one control plane instance MUST be run in each "failure zone" used for the cluster,
more instances per "failure zone" are possible to provide fault-tolerance inside a zone.

Worker nodes are RECOMMENDED to be distributed over multiple zones. This policy makes
it OPTIONAL to provide a worker node in each "failure zone", meaning that worker nodes
can also be scaled vertically first before scaling horizontally.

To provide metadata about the node distribution and possibly provide the ability
to schedule workloads efficiently, which also enables testing of this standard,
providers MUST annotate their K8s nodes with the labels listed below.
These labels MUST be kept up to date with the current state of the deployment.
```

mbuechse (commented):
Fine apart from the change that I indicated.

Standards/scs-0214-v2-k8s-node-distribution.md
mbuechse: Wait, why is this sentence still in here? Have you found out in the meantime (with the help of Team Container) what the situation is like with a shared pool of control-plane nodes, and how this would present itself to the test?

mbuechse (changes requested):

Standards/scs-0214-v2-k8s-node-distribution.md
cah-hbaum: Updated in [a33dd56](https://github.com/SovereignCloudStack/standards/pull/524/commits/a33dd56ecdd41182ece9f8bf0d41a1505dbcbed3)

cah-hbaum:
> Wait, why is this sentence still in here? Have you found out in the meantime (with the help of Team Container) what
> the situation is like with a shared pool of control-plane nodes, and how this would present itself to the test?

No, I haven't (because I had two other meetings conflicting with the last container meeting).

My problem cutting this part would still be that smaller CSPs can't necessarily provide a setup like this, especially for all their clusters. And that probably isn't necessary either.
IMO what would be needed here is a requirement for this standard to be used. Now I wouldn't call such a setup "high availability" necessarily, but something in that direction would be nice to have.

Standards/scs-0214-v2-k8s-node-distribution.md
cah-hbaum: Since this seems to be a low-hanging fruit in the opinion of the upstream guys, should I/we put the code in ourselves in order to speed the process up (https://github.com/kubernetes/cloud-provider-openstack/issues/2579) ?

cah-hbaum:
@mbuechse Please check this again and resolve the parts that are fine with you. I'm gonna coordinate the upstream work with @chess-knight now.

Standards/scs-0214-v2-k8s-node-distribution.md
chess-knight: PR https://github.com/SovereignCloudStack/cluster-stacks/pull/53 is ready for review and it includes a workaround (creating server groups, instructing CAPO to use these server groups for nodes, labelling nodes with custom host-id label). https://github.com/kubernetes/cloud-provider-openstack/issues/2579 is about labelling part. The server group part is still missing in CAPO. I will try to comment in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1912 to speed it up.

chess-knight: Also, I am unsure if it is low-hanging fruit, as I wrote in https://github.com/kubernetes/cloud-provider-openstack/issues/2579 - Live migration can be problematic because the label will not be updated. IMO, it needs a discussion.

---

And what about bare-metal clusters? In that case, we do not need this specific label right?

mbuechse:
@cah-hbaum Looking good, thanks for all the work! Now, only matter of the "optionality" of this standard needs some discussion IMO. The certificate scope yaml permits marking a standard optional, but then I don't know what good the standard is, because most people, when they look at the certificate, they will never notice whether this particular standard was satisfied or not. And I still don't understand why this standard is so hard to satisfy after all, even for small providers. If it's indeed possible that multiple k8s clusters share the same hosts for their control plane, then each cluster can have three different hosts, and all clusters in total still only need three hosts, so where's the problem?

cah-hbaum:
> @cah-hbaum Looking good, thanks for all the work! Now, only the matter of the "optionality" of this standard needs some discussion IMO. The certificate scope yaml permits marking a standard optional, but then I don't know what good the standard is, because most people, when they look at the certificate, they will never notice whether this particular standard was satisfied or not. And I still don't understand why this standard is so hard to satisfy after all, even for small providers. If it's indeed possible that multiple k8s clusters share the same hosts for their control plane, then each cluster can have three different hosts, and all clusters in total still only need three hosts, so where's the problem?

I'm gonna update the standard to make it NOT optional and take it to a vote or request in the "SCS Container" Room.

cah-hbaum:
Posted into the "SCS Container" Matrix room, just putting the whole thing in here as a paraphrase to make it persistent:

> Hello again,
> 
> I got some standardization work again for you today and a request for some input for https://github.com/SovereignCloudStack/standards/pull/524.
> This PR updates the "Node Distribution" standard. Before the latest update, the standard was OPTIONAL for a CSP, since
> we also had small CSPs with not much available infrastructure in mind. Sure there are possible solutions for these
> scenarios (e.g. solutions with shared control-planes to make the overall footprint of clusters smaller allegidly), but a
> change to these solutions isn't always possible.
> 
> Now while we tried to figure these problems out, some more potential points for problems appeared: https://github.com/SovereignCloudStack/standards/issues/527
> With these failsafe levels, some things we probably need to talk about in the context of this standards come up, like
> (High) Availability, Failsafe, Failover, Redundancy and so on.
> 
> Personally, at this point in time, I would probably rework the standard in a broader fashion. I don't think that we can or
> should dictate this "Node Distribution" (which is kind of a "Redundancy"/"Availability" feature) for every K8s cluster that
> is offered through a KaaS offering.
> Instead, the standard should define the Use Case it wishes to accomplish/help with, which is probably some kind of
> "(High) Availability", and make this (at least a bit) dependent on the now developing "Failsafe Taxonomy" (https://github.com/SovereignCloudStack/standards/issues/527).
> We could then define the requirements (in this case the "Node Distribution") for such an offering. The standard should
> probably made broarder in this case, since "Availability" in K8s requires more than just distribution of nodes. But the
> currently used document could become a part of this future document.
> 
> Now I would like to think what your opinion about this is or if you would like this to go in another direction?
> 
> Thanks and best regards
> Hannes

martinmo:
IMHO it is ok to require conformant implementations to set some additional labels, even if some work needs to be done initially to support it (e.g., in yaook/k8s). And especially with `topology.kubernetes.io/region` and `topology.kubernetes.io/zone`, we have some well-known labels defined upstream by Kubernetes (see https://kubernetes.io/docs/reference/labels-annotations-taints) – basically, they are already standardized, but optional.

Standardized labels are not only useful for the conformance test. Another important use case is that the customers can use these labels to distribute their own applications over failure zones using [Pod topology spread constraints](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/).

Also, in https://github.com/SovereignCloudStack/standards/pull/524#discussion_r1609751538, an important question about `topology.scs.community/host-id` was raised:

> And what about bare-metal clusters? In that case, we do not need this specific label right?

Theoretically, in a bare-metal cluster, the `kubernetes.io/hostname` label would be enough. However, the conformance test would need to detect this situation and correctly decide which label (host-id or hostname) should be used. This is an additonal source of error. I think even a bare metal K8s node should have a host-id label (every Linux machine should have it, accessible via `hostid` or in binary form at `/etc/hostid`).

mbuechse:
> I think even a bare metal K8s node should have a host-id label (every Linux machine should have it, accessible via `hostid` or in binary form at `/etc/hostid`).

Yes, that would be desireable from my POV. People shouldn't be expected to check whether the cluster is running on bare metal or not.

martinmo:
> (Labels default in `yaook/k8s` are `kubernetes.io/hostname=XXX` and `topology.cinder.csi.openstack.org/zone=YYY`, a region wasn't available)

I double checked this: [yaook/k8s connects to the OpenStack layer](https://yaook.gitlab.io/k8s/devel/user/explanation/abstraction-layers.html#yk8s-on-openstack) via the [OpenStack cloud controller manager (CCM)](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md) which sets the `topology.kubernetes.io/region` and `topology.kubernetes.io/zone` labels, amongst others. But it requires that the underlying OpenStack instance [has the corresponding metadata set](https://github.com/kubernetes/cloud-provider/blob/master/controllers/node/node_controller.go#L523) (in other words, it requires proper AZ and region configuration at the IaaS layer).

cah-hbaum:
Ok makes sense to me, thank you for your answers.
We're now just waiting for the discussion this thursday in the "SCS Container Meeting" about keeping/removing/rewriting the part of this standard that makes it optional.
After this is resolved, this will be merged along with https://github.com/SovereignCloudStack/standards/pull/558.

cah-hbaum:
I took to the question to the "SCS Container" Matrix room. Since nobody brought something forward against it, I would say we can merge this issue.

cah-hbaum:
@artificial-intelligence I added you here for review, because we had the question, if this standard is possible to do with a Gardener setup and @mbuechse mentioned that you have some knowledge here. If that is not the case, you can "unrequest" yourself and maybe put in someone else who has more knowledge about this.

cah-hbaum:
Also because we talked about this in the meeting:
This standard is kind of "blocked" right now, since neither the reference implementation nor a few other K8s examples automatically provide the labels mentioned here.
BUT the standard could be merged nonetheless, since it is only in `Draft` for now.

mbuechse (approved):

chess-knight (approved):
Draft approved

Standards/scs-0214-v2-k8s-node-distribution.md
```diff
@@ -0,0 +1,129 @@
...
+## Decision
+
+This standard formulates the requirement for the distribution of Kubernetes nodes in order
+to provide a fault-tolerant and available Kubernetes cluster infrastructure.
+
+The control plane nodes MUST be distributed over multiple physical machines.
+Kubernetes provides [best-practices][k8s-zones] on this topic, which are also RECOMMENDED by SCS.
+
+At least one control plane instance MUST be run in each "failure zone" used for the cluster,
+more instances per "failure zone" are possible to provide fault-tolerance inside a zone.
```
chess-knight, line 73: I experienced an interesting scenario, where the node distribution test will succeed.
**SCS Compliance results** in https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/742#issuecomment-2172470815 show `INFO: The nodes are distributed across 2 host-ids.`
I am wondering now if 3 control plane nodes are distributed over 2 hosts, it means that 2 control plane nodes are located on the same physical host. What will happen with the k8s cluster if this host goes down? AFAIK etcd has in this case only one member failure tolerance, it will stop functioning!

chess-knight: These lines in the standard and corresponding test don't require hard anti-affinity.

cah-hbaum: I understand the concern. We would probably need to adapt the lines above the ones you mentioned from
```
The control plane nodes MUST be distributed over multiple physical machines.
Kubernetes provides [best-practices][k8s-zones] on this topic, which are also RECOMMENDED by SCS.
```
to
```
The control plane nodes MUST be distributed over multiple physical machines; a control plane MUST contain at least three nodes and be distributed over three machines.
Kubernetes provides [best-practices][k8s-zones] on this topic, which are also RECOMMENDED by SCS.
```

Would that be alright?

chess-knight: `at least three nodes` can this be considered as node distribution standard or is it more suited for some k8s HA standard?

chess-knight: Also, can we somehow incorporate the fact that etcd can be external? Then we are probably good with the current formulation, this comment is valid only if stacked etcd for k8s cluster is used.

chess-knight (commented):
I am wondering if this standard is valid also for clusters with the managed control plane. E.g. Kamaji, where CP runs inside pods in the management cluster and when you do `kubectl get nodes` in the workload cluster you see only worker nodes there.

Standards/scs-0214-v2-k8s-node-distribution.md
cah-hbaum: Probably more relevant for a `HA` standard, but TBF that was always the problem I had with this standard in general, since `node distribution` or better `fault tolerance` also kind of (at least partly) includes `(high)  availability`.
Personally, after working on this standard this long I would probably change it up altogether to create a `(high) availability` that is based on this document and make this `distribution` standard optional, since not every cluster needs to be distributed (IMO).


Coming back to the case presented by you above:
> I am wondering now if 3 control plane nodes are distributed over 2 hosts, it means that 2 control plane nodes
> are located on the same physical host. What will happen with the k8s cluster if this host goes down?
> AFAIK etcd has in this case only one member failure tolerance, it will stop functioning!
I understand this scenario (and as far as I know you're right about the etcd functionality). The question would be how to fix this scenario in the standard. There  shouldn't be a number of physical nodes smaller than three, if etcd in use has 3 members.
So I guess this means this is still a distribution issue for us and not a `HA` issue?

chess-knight: > Also, can we somehow incorporate the fact that etcd can be external? Then we are probably good with the current formulation, this comment is valid only if stacked etcd for k8s cluster is used.

Now I am [reading](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/ha-topology/#external-etcd-topology) that also for external etcd you need 3+3 nodes. Do you know why it is so?

> I understand this scenario (and as far as I know you're right about the etcd functionality). The question would be how to fix this scenario in the standard. There  shouldn't be a number of physical nodes smaller than three, if etcd in use has 3 members.
So I guess this means this is still a distribution issue for us and not a `HA` issue?

https://kubernetes.io/docs/setup/best-practices/multiple-zones/#control-plane-behavior mentions:
```
If availability is an important concern, select at least three failure zones and replicate each individual control plane component (API server, scheduler, etcd, cluster controller manager) across at least three failure zones.
```
So maybe we can keep it and put it only later to the HA standard?

cah-hbaum:
> I am wondering if this standard is valid also for clusters with the managed control plane. E.g. Kamaji, where CP runs inside pods in the management cluster and when you do `kubectl get nodes` in the workload cluster you see only worker nodes there.

Thats also something we had some discussions about already and we decided that we want to have a discussion about this in the follow up issue that would make this standard `Stable`.

cah-hbaum:
I'm gonna keep it like it is for now and will update the standard in the [follow-up issue](https://github.com/SovereignCloudStack/standards/issues/639) with regards to the things discussed here.

Standards/scs-0214-v2-k8s-node-distribution.md
cah-hbaum: Interesting, I already read about this some time ago.
I guess the idea behind the `external` etcd is separation of concerns in your infrastructure and better control (and possible security) for a cluster. So it makes sense to separate the underlying etcd nodes and the kubernetes control plane. As the linked article mentions
```
This topology decouples the control plane and etcd member. It therefore provides an HA setup where losing a
control plane instance or an etcd member has less impact and does not affect the cluster redundancy
as much as the stacked HA topology.
```
Now I think that use-case wouldn't be implemented that often, but its an interesting thing to consider for a setup with required redundancy.

Personally, I would put it into a future `High availability` standard.

I will mention all this in #639 and we can keep on discussing everything there.

Issue #526, opened 2024-03-18
Refine CVE check in scs-0210-v2 test script.


mbuechse:
The test script currently does not really check whether any patch-level update that **targets** any critical CVEs is deployed in time.

Furthermore, the standard is a bit vague about whether this part is actually required or recommended.

Thirdly, could you make some kind of suggestion of how to best integrate with CVE check tools? For instance, the test script could accept a log file by one of these tools and just verify that the tool ran fine. You could thenn add this to the standard as a recommendation, I think we might get this in even with the now stable standard because it wouldn't turn any compliant clouds non-compliant.

martinmo:
Note to myself: while working on #476 I noticed that the `VersionRange` class doesn't cover all use cases nicely. Without workarounds, it covers the cases:

- a single version is affected (`upper_version` is `None`)
- a range is affected, but both ends must be known

It doesn't cover the case that all versions prior `upper_version` are affected, which I have worked around using a version `0.0.0` as the lower version. And also the other way around (which would mean, there is no patched version available/known).

martinmo:
And a general note: I would actually like to replace our custom CVE retrieval and parsing with an existing library, if possible. An obvious candidate I want to evaluate in this regard is [cvelib](https://pypi.org/project/cvelib/).

mbuechse:
@martinmo Yes, very well. Also (as stated in the description of this issue) we might require the use of some external CVE check tool, if you can find something appropriate.

mbuechse:
Oh, and one other note: maybe in the course of this issue, you can also try to make the check work for any given date, instead of just the current, so that unit tests can work without monkeypatching.

martinmo:
Today I brought the question about which scanning tools could be used into the Team Container call. However, because of holidays/vacation, we were only two people and this couldn't be discussed with a broader audience (I'll try again in the next week if necessary).

In the meantime, I picked up another tool that I will evaluate: `kubescape` (https://kubescape.io/).

martinmo:
I performend some evaluation on more CVE scanner tools for K8s. Unfortunately, most of them are not suited for our purpose – they either do not scan cluster components (e.g., the `kubelet` or `apiserver`) or they cannot easily be included in a CI pipeline (some of them are nice UI dashboards):

- `kubescape` (https://kubescape.io) cli does only vulnerability scanning of the container images
- clair (https://github.com/quay/clair) does the same
[...]

PR #531, opened 2024-03-20
Add unit and regression tests for conformance tests

### LINKED ISSUE #476: Regression and unit tests for the standards testsuite
As a developer of SCS conformance tests, I would like to use unit tests to be able to verify that the building blocks of my conformance test scripts work as intended, in a repeatable and automated fashion.

An example of a component that could be (unit) tested is the flavor name parser. There are already some manual testing examples that can be invoked via `Tests/iaas/flavor-naming/check-yaml.py Tests/testing`. However, they require the developer to manually compare against the expected outcome.

(Context: A recent case where such a unit test suite would have helped to spot regressions can be found in [PR #455](https://github.com/SovereignCloudStack/standards/pull/455#pullrequestreview-1890723894).)

martinmo:
This PR adds basic unit and regression test infrastructure for our conformance tests in `./Tests` using the `pytest` tool.

Included are

- a new GitHub workflow that runs `pytest` tests and blocks merging if tests fail,
- unit and regression tests for the most important bits of the version policy and flavor naming conformance tests,
- a section in `Tests/README.md` explaining how to use the unit tests,
- bug fixes for bugs found by the new tests,
- and last but not least some refactorings I couldn't resist to perform already.

Resolves #476.


mbuechse:
This seems to be progressing nicely. Will you remove the Draft state soon?

martinmo:
> This seems to be progressing nicely. Will you remove the Draft state soon?

Thank you! It is ready for review now. I've updated the PR description accordingly.

I took a bit longer because of layer 8 problem on my part – I spent a lot of time debuggig an issue which turned out to be caused by a tiny typo :see_no_evil:

Tests/README.md
```diff
@@ -70,20 +70,54 @@ docker run -it --env OS_CLOUD=CLOUDNAME -v ~/.config/openstack:/root/.config/ope
...
+
+You are encouraged to cover new conformance tests with unit tests!
+We run the tests on a regular basis in our GitHub workflows.
+
+### Maintaining the Python dependencies
+
+We list our main Python dependencies in `requirements.in`. Additionally, we list
+[unit tests](#unit-and-regression-tests) dependencies in `test-requirements.in`.
+The `*.in` files are fed to `pip-compile` to produce corresponding `*.txt` files
+that contain an exact, version pinned listing of *all* dependencies, including
```
mbuechse, line 109:
```suggestion
that contain an exact, version-pinned listing of *all* dependencies, including
```

Tests/README.md
```diff
@@ -70,20 +70,54 @@ docker run -it --env OS_CLOUD=CLOUDNAME -v ~/.config/openstack:/root/.config/ope
...
 `pip-compile` can be installed via `pip install pip-tools`.
 It needs to be run in two cases:
 
-1. You modified `requirements.in` (e.g., added a new dependency): run
-   `pip-compile requirements.in`.
-2. You want to bump the pinned dependencies: run `pip-compile --upgrade requirements.in`.
+1. You modified an `*.in` file: run `pip-compile <INFILE>` (e.g.,
+   `pip-compile test-requirements.in`).
+2. You want to bump the pinned dependencies: add the `--upgrade` flag to the
+   `pip-compile` invocation (e.g., `pip-compile --upgrade requirements.in`).
```
mbuechse, line 118:
```suggestion
1. You modified an `*.in` file: run `pip-compile <INFILE>`; e.g., 
    
    `pip-compile test-requirements.in`
    
2. You want to bump the pinned dependencies: add the `--upgrade` flag to the
   `pip-compile` invocation; e.g.,
   
   `pip-compile --upgrade requirements.in`.
```

not sure if this works this way though, I'm not a Markdown wiz

Tests/iaas/flavor-naming/check_yaml.py
```diff
@@ -8,9 +8,9 @@
 
 Also check that the flavor properties encoded in the YAML match those encoded in the name.
 """
-import os
-import os.path
+
 import sys
+from pathlib import Path
```
mbuechse, line 13: I would sort imports by the module, regardsless of `import` vs `from ... import`. No?

Tests/iaas/flavor-naming/check_yaml_test.py
```diff
@@ -0,0 +1,40 @@
...
+ERROR: flavor 'SCS-8V-32': field 'disk' contradicting name-v2 'SCS-8V-32'; found 128, expected undefined
+ERROR: flavor 'SCS-1V-2': field 'cpus' contradicting name-v2 'SCS-1V-2'; found 2, expected 1
+ERROR: flavor 'SCS-2V-4-20s': field 'disk0-type' contradicting name-v2 'SCS-2V-4-20s'; found 'network', expected 'ssd'
+ERROR: flavor 'SCS-4V-16-100s': field 'disk' contradicting name-v2 'SCS-4V-16-100s'; found 10, expected 100
+ERROR: file 'scs-0103-v1-flavors-wrong.yaml': found 7 errors
+""".strip()
+
+TEST_PARAMS = [
+    (CLEAN_YAML_DIR, 0, ""),
+    (BUGGY_YAML_DIR, 1, EXPECTED_ERRORS)
```
mbuechse, line 32: I prefer a trailing comma, and why did you opt for a list instead of a tuple?

Tests/kaas/k8s-version-policy/k8s_version_policy.py
```diff
@@ -239,38 +239,25 @@ def parse_github_release_data(release_data: dict) -> K8sRelease:
 
 @dataclass(frozen=True, eq=True)
 class VersionRange:
-    """Version range with a lower and upper bound."""
+    """
+    Version range with a lower and upper bound. Supports checking if a
```
mbuechse, line 243:
```suggestion
    Version range with a lower and upper bound.

    Supports checking if a
```

I think PEP-8 says to use the first line for the summary only.

Tests/kaas/k8s-version-policy/k8s_version_policy.py
```diff
@@ -239,38 +239,25 @@ def parse_github_release_data(release_data: dict) -> K8sRelease:
...
+    K8sVersion is in the range using the `in` operator.
+    If `inclusive` is true, `upper_version` is inside the range (i.e.,
```
mbuechse, line 245: Should we change true to (a) True or (b) true-ish? What is the Pythonic way?

Tests/kaas/k8s-version-policy/k8s_version_policy.py
```diff
@@ -411,11 +405,11 @@ def check_k8s_version_recency(my_version: K8sVersion, releases_data: list[dict],
         if release.age > PATCH_VERSION_CADENCE:
             # whoops, the cluster should have been updated to this (or a higher version) already!
             return False
-        if my_version in cve_version_list and release.age > CVE_VERSION_CADENCE:
-            # -- three FIXMEs:
-            # (a) can the `in` ever become true, when we have a version vs a set of ranges
-            # (b) if the release still has the CVE, then there is no use if we updated to it?
-            # (c) the standard says "time period MUST be even shorter ... it is RECOMMENDED that ...",
+        is_affected = any(my_version in _range for _range in cve_affected_ranges)
+        if is_affected and release.age > CVE_VERSION_CADENCE:
```
mbuechse, line 409: Now this is just a matter of taste, but I usually prefer this:
```suggestion
        ranges = [my_version in _range for _range in cve_affected_ranges]
        if ranges and release.age > CVE_VERSION_CADENCE:
```
My reasoning is that this has similar computing cost without throwing away the information.
The information is usually nice to have for debug logging, among other things.
But still, it is a tiny matter of taste.

Tests/kaas/k8s-version-policy/k8s_version_policy_test.py
```diff
@@ -0,0 +1,181 @@
...
+EXPECTED_RECENCIES = {
+    datetime(2024, 1, 17): True,
+    datetime(2024, 1, 24): True,
+    datetime(2024, 1, 25): False,
+}
+
+
+@pytest.mark.parametrize("ref_time, expected_recent", EXPECTED_RECENCIES.items())
+def test_check_version_recency_without_cve(release_data, ref_time, expected_recent):
+    with mock.patch("k8s_version_policy.datetime", wraps=datetime) as dt:
```
mbuechse, line 54: I know that I'm being obnoxious right now, but should we just make the `check_k8s_version_recency` API more general? It could have an optional argument `check_date=None` (and if `None` is supplied, the function will use `datetime.now`).

Tests/kaas/k8s-version-policy/testdata/README.md
```diff
@@ -0,0 +1,10 @@
+# Testdata for unit tests
+
+## releases.json
+
+was created from real release data filtered through `jq`:
```
mbuechse, line 5: maybe add some note that this file can be left as-is and the command is just FYI
(otherwise one might think that this command is meant to be issued by the user)

mbuechse (approved):
Great work! I have a few minor remarks, as usual... but that is not meant to diminish anything!

Tests/kaas/k8s-version-policy/k8s_version_policy.py
martinmo: With this suggestion, `ranges` would be a list of `bool`s. And the lhs of the `if` already evaluates to `True` if the list is not empty, even if `my_version` is not affected.

martinmo: By "throwing away information" you mean that I used a generator instead of a list?

mbuechse: sorry, I was too quick here, what I meant was rather
```suggestion
        ranges = [_range for _range in cve_affected_ranges if my_version in _range]
        if ranges and release.age > CVE_VERSION_CADENCE:
```

mbuechse: By throwing away I meant that we don't have the "witnesses" (or "culprits", if you will). Whatever you would show to someone who wanted to debug this thing.

Tests/kaas/k8s-version-policy/k8s_version_policy_test.py
martinmo: Yeah, let's try to avoid monkey patching were possible. This `mock.patch(…)` call is also error-prone, because it relies so much on internal details of the module under test.

martinmo: Ahh, unfortunately we would need to carry down this through multiple functions or methods. Without a refactoring this will look ugly, so I would leave it as-is for the time being.

Tests/kaas/k8s-version-policy/k8s_version_policy.py
martinmo: I agree!

martinmo:
The markdown linter complains, I'm on it...

martinmo:
Converted to draft again to block merging, I found another bug when testing with minikube, unfortunately...

    CRITICAL: '<=' not supported between instances of 'NoneType' and 'K8sVersion'

I'm currently investigating.

martinmo:
The issue is now resolved. TL;DR: it was bad handling of the case when the lower version of the affected version range was not specified. (The bigger picture is that the `VersionRange` class is not the best abstraction to cover all use cases – but because this PR is scoped to the tests, I don't want to change it completely. This will be done in #526.)

mbuechse:
You are GO for launch!
<https://www.youtube.com/watch?v=lMtWWls4oas>

Issue #534, opened 2024-03-21
SCS Standard Images reference YAML not clearly communicated


markus-hentsch:
Putting myself into the perspective of a CSP that aims for SCS conformance I noticed the following shortcomings while reading the "SCS Standard Images" document in the official SCS docs at https://docs.scs.community/standards/scs-0104-v1-standard-images and trying to reach conformance to it.

### Current shortcomings

The standard spends a lot of time elaborating on the process of how images are standardized and how the standardization is implemented using YAML descriptions but little on how to figure out the currently standardized images.
My primary question as a CSP would be: "What images do I have to provide to be SCS compliant?". I don't feel this question is answered in a confident way. The standard currently only states this small sentence under the headlines "Lifecycle considerations"/"YAML lifecycle":

> The YAML file is generally located in this repository under `/Tests/iaas`.

Problems with that:

- This should be some kind of clickable permalink. The "in this repository" is a bad reference when reading the docs.scs.community page as a docs page.
    - Even if discovering and clicking the "GitHub" link on the top right of the page, it forwards to the docs repository, not the standards repository where the file resides.
- In my opinion, the wording of the sentence and its location in the standard does a poor job at communicating that this the central important YAML to achieve conformance with the standard. It sounds more like an example file similar to the code blocks illustrated in the preceeding sections.
    - At the bottom of the standard there is a link to OSISM's image manager which also ships YAML templates for images[^1] where as a CSP I could see myself asking instead: "are these the YAMLs I have to apply for the standard?"

### Suggestions for improvement

In general, I would like to improve the discoverability of the images YAML file[^2] and emphasize its role in relation to achieving standard conformance, just like in other standards (e.g. the flavor standard where mandatory and recommended flavors are listed directly in the standard):

- Provide a clickable reference link to the currently valid images YAML file[^2] (that also works on the rendered docs page) OR if possible, directly render the YAML into the docs page.
- Better communicate that the referenced YAML file is the binding standard regarding the list of mandatory/recommended images that need to be provisioned to achieve conformance to the standard.

I'd like to hear your opinion on this.

[^1]: https://github.com/osism/openstack-image-manager/tree/main/etc/images
[^2]: https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/scs-0104-v1-images.yaml

mbuechse:
This is not controversial – I agree on all accounts. Thanks for spotting this.

mbuechse:
[...]

PR #536, opened 2024-03-22
Make yaml file location more explicit and more user friendly

mbuechse:
This is related to #534 -- not yet a solution, but a significant improvement from my point of view.

### DIFF
--- a/Standards/scs-0104-v1-standard-images.md
+++ b/Standards/scs-0104-v1-standard-images.md
@@ -103,7 +103,8 @@ recommended, while only Ubuntu 22.04 is mandatory.
 
 ### YAML lifecycle
 
-The YAML file is generally located in this repository under `/Tests/iaas`.
+The YAML file is generally located under
+<https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/>.
 
 Any change that could render existing installations non-conformant (i.e., when new
 specifications are added, when the name scheme of a specification is changed so as to


markus-hentsch:
Thanks for getting this started!

I'm currently pondering how we could communicate the scope of the standard and implications for CSPs more clearly. I'm currently thinking of something along the lines of:

> Note that this standard primarily defines the standardization _process_ for images resulting in the YAML descriptions.
> To achieve conformance with the standard, a CSP has to provision images as per the currently valid YAML resulting from this standard provided by SCS at a particular time.
> Due to the more short-lived nature of image releases, the actual YAML that specifies the images is regularly updated and therefore not a direct part of this standard document.
>
> A CSP may retrieve the currently valid YAML at any time from https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/

... and putting this section somewhere where it makes sense and is easily discoverable by a reader who is interested in how to apply the standard.

mbuechse:
@markus-hentsch I agree with the general direction, yet I'm puzzled as to where to put such a paragraph, and we can also debate about the correctness of it:

> To achieve conformance with the standard, a CSP has to provision images as per the currently valid YAML

I'm not sure that this is true. I think this is a real "hole" in the standard. The standard makes assertions such as the following (and I quote):

> If the status is mandatory, then the image MUST be present.

The following premise is NOT stated:
- a YAML file is given
- a cloud environment is given
- we interpret this YAML file with respect to this cloud environment.

And, worse, what is not stated either: that of course we assume that this premise is true.

And without all that, your suggestion doesn't work. I will think of something, but in the meantime, it would be good to get this tiny improvement into main. Note that this PR will NOT close the issue #536, and deliberately so.

Standards/scs-0104-v1-standard-images.md
```diff
@@ -103,7 +103,8 @@ recommended, while only Ubuntu 22.04 is mandatory.
 
 ### YAML lifecycle
 
-The YAML file is generally located in this repository under `/Tests/iaas`.
+The YAML file is generally located under
+<https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/>.
```
markus-hentsch, line 107: I'm not familiar with this `<` and `>` syntax for links in Markdown. It seems to render fine using "View file" here on GH and passes the PR checks but I could not find this links syntax at:

- https://www.markdownguide.org/basic-syntax/#links
- https://docs.gitlab.com/ee/user/markdown.html#links
- https://docusaurus-archive-october-2023.netlify.app/docs/markdown-features/links

I'm fine with it in case it works on the rendered docs.scs.communit page as well, I'm just curious which Markdown flavor this belongs to.

mbuechse: Oh, that's a good question. I honestly don't know, but: I think I started doing it because our very own linter demanded it!

mbuechse: It works with HedgeDoc, Github, Docusaurus and, AFAICT with the Python markdown package.

markus-hentsch:
> @markus-hentsch I agree with the general direction, yet I'm puzzled as to where to put such a paragraph, and we can also debate about the correctness of it:
> 
> > To achieve conformance with the standard, a CSP has to provision images as per the currently valid YAML
> 
> I'm not sure that this is true. I think this is a real "hole" in the standard. The standard makes assertions such as the following (and I quote):
> 
> > If the status is mandatory, then the image MUST be present.
> 
> The following premise is NOT stated:
> 
>     * a YAML file is given
> 
>     * a cloud environment is given
> 
>     * we interpret this YAML file with respect to this cloud environment.
> 
> 
> And, worse, what is not stated either: that of course we assume that this premise is true.
> 
> And without all that, your suggestion doesn't work. I will think of something, but in the meantime, it would be good to get this tiny improvement into main. Note that this PR will NOT close the issue #536, and deliberately so.

Interesting perspective. I will need to think about this for a bit. I was not involved in the creation of the standard so I approached it from a totally different angle. But I agree, let's take this back into the issue and focus on fixing the existing reference here.

Standards/scs-0104-v1-standard-images.md
markus-hentsch: Found it: https://www.markdownguide.org/basic-syntax/#urls-and-email-addresses

URL was the keyword, not "link" ;)

markus-hentsch: Never used that before, TIL.

markus-hentsch (approved):

Issue #366, opened 2023-10-23
[Standardization] GPU naming convention needs further refinements


anjastrunk:
GPU name in [SCS Flavor Naming Standard](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0100-v3-flavor-naming.md#optional-gpu-support) need further refinement. The following aspects are described insufficiently:   

- [x] GPU capabilities are added to flavor name as extension. See [Complete Proposal for systematic flavor naming](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0100-v3-flavor-naming.md#complete-proposal-for-systematic-flavor-naming). However, the order of extensions is unclear. As there are three extensions currently, I would make sense and facilitate parsing/mapping if extensions have a dedicated order.
- [x] GPU naming supports suffix `h`, which can be set multiple times indicating a high-performance GPU. However "high-performance" is neither explained in detail, nor there is a mapping from `h`, `hh`, `hhh`... to measure a GPU properly as it is done for [CPU Frequency](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0100-v3-flavor-naming.md#frequency-suffixes)`. In favor of comparison and interoperability, standard for GPU naming SHOULD be very strict and clear here.
- [x] There are no examples provided for flavor naming with GPU support, as is is done for [CPU]() or [Memory](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0100-v3-flavor-naming.md#examples-5)
- [x] #375
- [x] According to SCS flavor naming, GPU generation, such as Ampere or Hopper for Nvidia, can be defined by adding appropriate suffix to GPU definition. IMO, placing generation is not sufficient as there is a huge performance difference between A10 and A100, both GPUs of generation "Ampere". Hence, we need a further refinement here, to point out GPU capabilities more precisely.  
- [ ] Standard does not support definition of number of physical or virtual GPUs

garloff:
> * [ ]  GPU capabilities are added to flavor name as extension. See [Complete Proposal for systematic flavor naming](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0100-v3-flavor-naming.md#complete-proposal-for-systematic-flavor-naming). However, the order of extensions is unclear. As there are three extensions currently, I would make sense and facilitate parsing/mapping if extensions have a dedicated order.

The order is and always has been fixed.
A sentence to make this clear is added with PR #374.

[...]

PR #546, opened 2024-04-03
GPU-flavor-naming-refinement

cah-patrickthiem:
This PR handles the refinement of GPU flavor naming. It clarifies things and overhauls some possible inconsistencies in the current naming convention as well as the description. Therefore, this PR introduces an update to the document: [scs-0100-v3-flavor-naming.md](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0100-v3-flavor-naming.md).
For reference see issue [366-GPU naming convention needs further refinements](https://github.com/SovereignCloudStack/standards/issues/366).

**Note**: The initial commit just added the flavor naming document in version 4.


cah-patrickthiem:
After some research, see [here](https://github.com/SovereignCloudStack/standards/issues/366#issuecomment-2045429707) and below, I came to following conclusions:
The GPU flavor naming should be changed in a way that is more clear what to expect and most important unfortunately we cannot use a general performance indicator, which was introduced in the prior flavor naming standard. In general the current standard has 3 major problems I want to tackle here.

Derivation:

**current standard**: 
- right now the standard suggests something like this: SCS-16V-64-500s_GNa-14h
- lets translate it, this flavor indicates a 16 vCPU core flavor with 64gb RAM, 500gb disk and a passthrough Nvidia GPU (N) from the "Ampere" (a) generation with 14 "streaming multiprocessors (sm)" and an "h" for "high-performance"
- so far so good

**Problem 1 - transparency**

What GPU exactly is a GNa-14h? 
- in regard to the corresponding comment in the standard, this flavor should imply 1/4 of a Nvidia A30 GPU with the SM number of "14" (besides that, the capital "G" implies that it is a passthrough GPU rather than a virtual GPU and thus the "1/4" indicator does not really make sense - just a mistake in the standard) 
- the user would just not know that it is an A30 GPU, even if we make it "14*4=56" to  get the real number of SMs for this GPU, the user would still not know what to get

**Solution of Problem 1**
- make a GPU list for SCS with all corresponding specs and maybe even extend it to vGPUs where fractions of the real SM numbers are given
- "There we go, fixed!" - you might think

but there is more...

**Problem 2 - inconsistency**
- a "streaming multiprocessor" for Nvidia and a "computing unit" for AMD CANNOT be compared, at least not in a clear and logical way...
Why?

1. there are discrepancies between the performance of a GPU and the number of SMs or CUs or whatever
2. there are discrepancies if we look at one GPU generation from the same vendor 
3. there are discrepancies if we look at different GPUs from different vendors

e.g.:
- 


**Problem 3 - other factors**

cah-patrickthiem:
For the record, since I presented the current state in todays (10.07.24) IaaS call.
My favorite way of gpu flavor naming looks like this:

1. do not use SMs, CUs etc. and even do not use "h" indicators for high performance
2. handle vGPUs as described above, meaning: ...-5gNa-A100 translates to: you get 1 part out of 5 parts of a Nvidia A100
3. for vRAM I would go with the second proposal: "always include vRAM, don't split for vGPUs", but I think the third proposal would work as well

That means we would get something like: SCS-16V-64-500s_GNa-A100-40g for passthrough GPUs or SCS-16V-64-500s_3gNa-A100-40g for virtualized GPUs

mbuechse (commented):
I know this is just a draft, but for the sake of completeness, let me say it nonetheless:

- the new version would start out with `status: Proposal` and without the fields `stabilized_at` and `replaces` (I think, you can verify using scs-0001-v1)
- naturally, the introduction needs to be updated as well

mbuechse:
It would be good to see a tentative list of GPU models.
I think the vendor and generation letters would become superfluous, depending on whether the model identifier is required or not (but the generation is most definitely redundant). So this part should probably be adjusted.
It would also be good to see a completely spelled out syntax proposal, maybe for your "favorite" way that you described in your most recent comment above.
Finally, it would be good to have some background info regarding virtualization vs. passthrough. How does virtualization work (or rather, what is posible and what isn't), and why is passthrough limited to the whole GPU (I could imagine that the GPU might have independent units?), stuff like that. This explanation could go into a non-authoritative section of the standard, or it could go into a decision record.

Speaking of a decision record: frankly, your large comment above (phrased a bit more soberly) should probably be turned into one, in order to document the development process for posterity, but maybe also to structure the process even better.

Standards/scs-0100-v4-flavor-naming.md
```diff
@@ -0,0 +1,549 @@
...
+This list is likely not comprehensive and will grow over time.
+
+Rather than using random names `s5a.medium` and assigning a discrete set of properties
+to them, we wanted to come up with a scheme that allows to systematically derive
+names from properties and vice versa. The scheme allows for short names (by not
+encoding all details) as well as very detailed longer names.
+
+## Complete Proposal for systematic flavor naming
+
+| Prefix | CPUs & Suffix     | RAM[GiB]           | optional: Disk[GB]&type       | opt: extensions |
```
markus-hentsch, line 74: Shouldn't this be "opt: Disk[GB]&type"?

Quote from below:

> The optional fields are denoted in brackets (and have **opt:** in the header.
See below for extensions.

mbuechse: Funny. It's an issue of its own though, and that issue then should fix the other versions of the standard as well.

Standards/scs-0100-v4-flavor-naming.md
```diff
@@ -0,0 +1,549 @@
...
+| ------ | ----------------- | ------------------ | ----------------------------- | ----------------|
+| `SCS-` | N`L/V/T/C`\[`i`\] | `-`N\[`u`\]\[`o`\] | \[`-`\[M`x`\]N\[`n/h/s/p`\]\] | \[`_`EXT\]      |
```
markus-hentsch, line 76: Syntactically, I believe using `|` as an OR operator is more common than `/`.
Furthermore, I think these operators should not be in backticks - if my assumption is correct that the monospace characters are supposed to be literals whereas everything else outside the backticks is either a placeholder (such as M or N) or an operator (such as [ ] for optionals).

```suggestion
| Prefix | CPUs & Suffix              | RAM[GiB]           | opt: Disk[GB]&type                     | opt: extensions |
| ------ | -------------------------- | ------------------ | -------------------------------------- | ----------------|
| `SCS-` | N`L`\|`V`\|`T`\|`C`\[`i`\] | `-`N\[`u`\]\[`o`\] | \[`-`\[M`x`\]N\[`n`\|`h`\|`s`\|`p`\]\] | \[`_`EXT\]      |
```

mbuechse: I will repeat myself:
> It's an issue of its own [...], and that issue then should fix the other versions of the standard as well.

markus-hentsch: Whoopsie, ignore my comment - I completely missed the scope of this PR - my bad.

mbuechse: In other words: can you please only review things that are new in v4? As of now, IIRC, there aren't any, so you're done...

Standards/scs-0100-v4-flavor-naming.md
markus-hentsch: You are right. I'm sorry, I completely mistook the scope of this PR for a whole new standard draft since the diff shows the whole file but it's just a version increment and as such contains a full copy of the file.

Standards/scs-0100-v4-flavor-naming.md
markus-hentsch: > In other words: can you please only review things that are new in v4? As of now, IIRC, there aren't any, so you're done...

Yes, as I stated: my bad. I completely mistook this PR for something else because I was misled by the PR diff showing a whole new file. Sorry.

Standards/scs-0100-v4-flavor-naming.md
```diff
@@ -0,0 +1,549 @@
...
+
+This extension provides more details on the specific GPU:
+
+- pass-through (`G`) vs. virtual GPU (`g`)
+- vendor (X)
+- generation (N)
+- number (M) of processing units that are exposed (for pass-through) or assigned; see table below for vendor-specific terminology
+- high-performance indicator (`h`)
+
+Note that the vendor letter X is mandatory, generation and processing units are optional.
```
markus-hentsch, line 425: Just for my understanding: judging from the format above, the `G/g` marker and the high-perf indicator `h` are also optional right?

So `SCS-16V-64-500s_N` would be a valid minimal example just stating the existence of an nVidia GPU, correct?

mbuechse: I think even that section is so far unchanged compared to v3. As Patrick stated

> Note: The initial commit just added the flavor naming document in version 4.

(Which admits multiple readings, but my interpretation was that he merely copied the file with an increased version number.)

markus-hentsch: I am aware. As stated this question is just for my understanding. The format syntax of the original standard suggests that the `G/g` might be optional but I can't find any examples or statements confirming this.
I wanted to make sure that *just in case this is unintended* we could address this potential inaccuracy when @cah-patrickthiem will be editing this section anyway.

mbuechse: It's not optional AFAICT
https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/flavor-naming/flavor_names.py#L358

markus-hentsch: Then I think the usage of square brackets in "\[`G/g`\]" of the format specification is wrong and the brackets should be removed.

We could adjust this in the course of this refinement PR while we're at it since the syntax and explanations will most likely need to be updated anyway.

mbuechse: I don't think this format specification can be "wrong" it that sense, because it's not that precise. We would have to use some more precise syntax (such as EBNF or something). We might do that, but then we should change the standard throughout to use this syntax.

garloff:
> * in regard to the corresponding comment in the standard, this flavor should imply 1/4 of a Nvidia A30 GPU with the SM number of "14" (besides that, the capital "G" implies that it is a passthrough GPU rather than a virtual GPU and thus the "1/4" indicator does not really make sense - just a mistake in the standard)

Not a mistake: nVidia allows one physical GPU to be partitioned and then be exposed directly via pass-through as several PCIe devices for direct access by several VMs.

garloff:
> **Conclusion to inconsistency**: we can see, that the H100 has significantly less SMs than the MI250(x) has CUs but outperforms the AMD counterparts x times. Therefore it is not really consistent to assume a somewhat linear or understandable relation between SMs, CUs and performance.

Within one vendor and one generation of GPUs (e.g. nVidia Ampere), this number (SMs in case of nVidia) does have a well-defined meaning and allows you to see how much GPU compute performance you get. A `_GNa-56` is roughly 4x faster than a `GNa-14`. 
This number is in no way meant to compare GPUs from different generations let alone different vendors.

garloff:
> Another idea could be to use the "h", "hh" and "hhh" indicators always inside the same gpu generation. For example for Nvidia Ampere that would look something like this: A10 no "h", A14 "h", A30 and A40 "hh", A100 "hhh".

That was exactly the way it was meant to be used. Within one vendor and one generation, there may be variants with specially high frequency or specially high bandwidth memory (HBM). This indicator was meant to allow a provider that uses both variants and wants to also indicate this in the flavor name.

That said, I agree that this is not specific enough and maybe we should just say that vendors can add these modifiers in case they have several variants of GPUs from the same vendor and same generation with significantly different performance and in general discourage its use otherwise.

Standards/scs-0100-v4-flavor-naming.md
garloff: > Just for my understanding: judging from the format above, the `G/g` marker and the high-perf indicator `h` are also optional right?
> 
> So `SCS-16V-64-500s_N` would be a valid minimal example just stating the existence of an nVidia GPU, correct?

So we break the compatibility with the old naming and make life harder for the parser?

garloff: > Then I think the usage of square brackets in "[`G/g`]" of the format specification is wrong and the brackets should be removed.

True. Either `G` or `g` is needed to indicate a GPU and indicates PT or Virt

garloff:
General comments:
* I think it's an excellent idea to include the amount of VRAM that the VM will have access to.
* To get an indication of how much work you can do and whether your workload even functions, you need
  - Vendor and generation
  - Amount of compute power (SMs/CUs/EUs/ ... - YES, this is vendor and generation dependent) exposed to the VM
  - Amount of VRAM exposed to the VM

As user I know what I need:
1. I have a model that e.g. runs only on nVidia as it uses CUDA. (This is the case for quite some GPU workloads.), so I need `_GN` or `_gN`
2. I'm looking for recent generation, ideally AdaLovelace, so looking for `_GNl`
3. Like most things done on GPUs, my model is massively parallel, so I'm looking for a flavor with lots of SMs, the more the better.
4. I now my workloads needs at least 16GiB VRAM to perform well, so I need a flavor that has at least this

If I look at the original spec, we have a few shortcomings and a few things done well.
- There seems to be a misunderstanding that SMs/CUs/EUs (nVidia/AMD/intel) can be compared across vendors or even generations within a vendor. That is not the case, was never intended and should be clarified.
- I would argue that this is absolutely important. 1/4 of a GPU inside one generation has significantly less compute power than the full one. Inside a generation, the hardware vendors have several cards, ranging from small to medium to large, so even if the full card is always exposed, the performance can vary by a factor or 10 or so (!). IMVHO, we definitely need a size indicator.
- Note that at least nVidia has the capability of partitioning their cards, so they can be pass-through exposed to several VMs. Some allow quarters, some allow for 1/7 ... This is not the same as exposing a virtualized GPU, where the hypervisor would do it arbitrarily and not just expose a PCIe device.
- We had completely missed the aspect of VRAM. This is significant and knowing whether your LLM fits into VRAM decides whether or not your workload even works. So this absolutely is needed, IMVHO.
- The high-performance indicators `h` are indeed fuzzy. I'm unsure what to do with them. Sometimes we have several variants of a card, one with significantly higher frequencies, which we somehow would want to differentiate.
- The memory could also be high-performance. If we have an HBM variant and a GDDR variant, this does make a significant difference ... (that would be an `h` qualifier for the VRAM).
- I prefer NOT using model names from GPUs -- the GPU vendors may or may not make a mess out of it (such as the CPU vendors have where you can't easily deduce the generation/microarchitecture from the name any more).

Here would be my suggestion:
* We keep the SM/CU/EU, with clarifying words
* We add VRAM
* We create a table of existing GPUs and their names
  - The table should also contain the commonly used hardware partitioned options
  - Names for new hardware could still be derived systematically, so if we don't update this table every month, it still makes sense ...
* If there are variants with signifcantly different frequencies *within one generation*, we can have a well defined `h` or even `hh` modifier
* Some for VRAM -- if there are GDDR and HBM variants within one generation, we can have a well defined `h` modifier
* If we would keep VRAM optional, this would even be backwards compatible (though on the long run, we'd want VRAM)   

`_<G/g><Vendor>[<Gen>-<SM/CU/EU>[h[h]][-<VRAM>[h[h]]]]` would be my choice ...
This has the advantage of being rather straight forward to implement in the code (flavor name generator, parser) and also is backwards compatible, i.e. old names would still be compliant. We would of course encourage the indication of VRAM and possibly mandate it in a later version of the spec.

Users and vendors may want marketing names. We could always encode them in extra specs, such as scs:gpu-vendor="nVidia", scs:gpu-generation="AdaLovelace", scs:gpu-model="L40", scs:gpu-fraction="1/4", scs:gpu-vram="20".
Long-term, this may be all we need and we stop compressing this information into a name. As long as we do, I would refrain from reflecting vendor-chosen names in our flavor names.

Just my 0.02€.

cah-patrickthiem:
> General comments:
> 
> * I think it's an excellent idea to include the amount of VRAM that the VM will have access to.
> * To get an indication of how much work you can do and whether your workload even functions, you need
>   
>   * Vendor and generation
>   * Amount of compute power (SMs/CUs/EUs/ ... - YES, this is vendor and generation dependent) exposed to the VM
>   * Amount of VRAM exposed to the VM
> 
> As user I know what I need:
> 
> 1. I have a model that e.g. runs only on nVidia as it uses CUDA. (This is the case for quite some GPU workloads.), so I need `_GN` or `_gN`
> 2. I'm looking for recent generation, ideally AdaLovelace, so looking for `_GNl`
> 3. Like most things done on GPUs, my model is massively parallel, so I'm looking for a flavor with lots of SMs, the more the better.
> 4. I now my workloads needs at least 16GiB VRAM to perform well, so I need a flavor that has at least this
> 
> If I look at the original spec, we have a few shortcomings and a few things done well.
> 
> * There seems to be a misunderstanding that SMs/CUs/EUs (nVidia/AMD/intel) can be compared across vendors or even generations within a vendor. That is not the case, was never intended and should be clarified.
> * I would argue that this number is absolutely important. 1/4 of a GPU inside one generation has significantly less compute power than the full one. Inside a generation, the hardware vendors have several cards, ranging from small to medium to large, so even if the full card is always exposed, the performance can vary by a factor or 10 or so (!). IMVHO, we definitely need a size indicator.
> * Note that at least nVidia has the capability of partitioning their cards, so they can be pass-through exposed to several VMs. Some allow quarters, some allow for 1/7 ... This is not the same as exposing a virtualized GPU, where the hypervisor would do it arbitrarily and not just expose a PCIe device.
> * We had completely missed the aspect of VRAM. This is significant and knowing whether your LLM fits into VRAM decides whether or not your workload even works. So this absolutely is needed, IMVHO.
> * The high-performance indicators `h` are indeed fuzzy. I'm unsure what to do with them. Sometimes we have several variants of a card, one with significantly higher frequencies, which we somehow would want to differentiate.
> * The memory could also be high-performance. If we have an HBM variant and a GDDR variant, this does make a significant difference ... (that would be an `h` qualifier for the VRAM).
> * I prefer NOT using model names from GPUs -- the GPU vendors may or may not make a mess out of it (such as the CPU vendors have where you can't easily deduce the generation/microarchitecture from the name any more).
> 
> Here would be my suggestion:
> 
> * We keep the # of SM/CU/EU, with clarifying words
> * We add VRAM
> * We create a table of existing GPUs and their names
>   
>   * The table should also contain the commonly used hardware partitioned options
>   * Names for new hardware could still be derived systematically, so if we don't update this table every month, it still makes sense ...
> * If there are variants with signifcantly different frequencies _within one generation_, we can have a well-defined `h` or even `hh` modifiers
> * Same for VRAM -- if there are GDDR and HBM variants within one generation, we can have a well defined `h` modifier for the HBM variant
> * If we would keep VRAM optional, this would even be backwards compatible (though on the long run, we'd want to mandate VRAM)
> 
> `_<G/g><Vendor>[<Gen>-<SM/CU/EU>[h[h]][-<VRAM>[h[h]]]]` would be my choice ... This has the advantage of being rather straight forward to implement in the code (flavor name generator, parser) and also is backwards compatible, i.e. old names would still be compliant. We would of course encourage the indication of VRAM and possibly mandate it in a later version of the spec.
> 
> Users and vendors may want marketing names. We could always encode them in extra specs, such as scs:gpu-vendor="nVidia", scs:gpu-generation="AdaLovelace", scs:gpu-model="L40", scs:gpu-fraction="1/4", scs:gpu-vram="20". Long-term, this may be all we need and we stop compressing this information into a name. As long as we do, I would refrain from reflecting vendor-chosen names in our flavor names. Short-term, I would point them to the tables. These IMVHO should be Annexes to the naming spec, so we can have a different update schedule and don't need to revise the standard just because there is a new piece of hardware available.
> 
> Just my 0.02€.

I took some time thinking about your comment. 
First of all I absolutely agree on the VRAM aspect. I also agree on the memory type with "h" marker.

I am not strictly against the SM/CU/EU thing, but let me say that we need to really think about the specifics of different GPUs if we go that way. The connection between performance of a GPU and its Processing-Unit count is NOT linear. One can assume or build a rough estimate of the performance, yes, but twice the unit count does not make it twice as fast. Other aspects, such as memory (bandwidth), clock, drivers or even the core type (cuda, raytracing etc.) can massively impact the computing power of a GPU.

Example: 
A100 vs A10:
The Nvidia A100 is the top performing allrounder GPU in the Ampere generation, especially when it comes to high performance computing/AI workloads etc.
The A100 has 108 SMs, the A10 has 72 SMs.

**FP16 Tensor Core Performance:**
- A100: Up to 312 teraFLOPS
- A10: Up to 148 teraFLOPS
- Factor: 312/148 ≈ **2.11**

**INT8 Tensor Core Performance:**
- A100: Up to 624 TOPS
- A10: Up to 148 TOPS
- Factor: 624/148 ≈ **4.22**

FP64 CUDA Core Performance:
- A100: Up to 19.5 teraFLOPS
- A10: Up to 7.4 teraFLOPS
- Factor: 19.5/7.4 ≈ **2.63**

**Streaming Multiprocessors (SMs):**
- A100 SMs: 108
- A10 SMs: 72
- Factor: 108/72 = **1.5**

Another example:
A100 vs A40:

FP16 Tensor Core Performance:
- A100: Up to 312 teraFLOPS
- A40: Up to 156 teraFLOPS
- Factor: 312/156 = **2.0**

INT8 Tensor Core Performance:
- Factor: 624/312 = **2.0**

FP64 CUDA Core Performance:
- Factor: 19.5/9.7 ≈ **2.01**

Streaming Multiprocessors (SMs):
- A100 SMs: 108
- A40 SMs: 72
- Factor: 108/72 = 1.5


Last example:

garloff:
Hi Patrick,

thanks for the very detailed analysis.

No, performance is not linear, as it depends on the bottleneck.
Sometimes memory bandwidth is the limiting factor, sometimes the amount of compute resources (SMs/CudaCores), sometimes your code has synchronization mechanisms that limit the parallelism, ...

I probably triggered another wrong assumption when saying that you got 4x as much compute power from a GNa-56h (A30) compared to a GNa-14h (a quarter A30). This is very much dependent on your workload, as you correctly say. You may even create workloads, where GNa-56h is not much faster than FNa-14h.

Nevertheless, I stand by design idea that we should tell a user how much resources she gets. In one case it's 14 SMs (896 Cuda Cores for Ampere) for a quarter A30 and in the other it's 56 SMs (3584 Cuda Cores) for a full one.
For a H100, this could be 132SMs (18896 Cuda Cores - full H100) vs 18SMs (2304 Cuda Cores - 1/7 H100) or 2/7 or ...

It's like vCPUs in instances -- 2 vCPU does not necessarily make your workload twice as fast, but you know how much CPU power you roughly get ...

garloff:
Another note:
Let's please call the Multi-Instance-GPU thing from nVidia partitioning and NOT virtualization.

cah-patrickthiem:
> Hi Patrick,
> 
> thanks for the very detailed analysis.
> 
> No, performance is not linear, as it depends on the bottleneck. Sometimes memory bandwidth is the limiting factor, sometimes the amount of compute resources (SMs/CudaCores), sometimes your code has synchronization mechanisms that limit the parallelism, ...
> 
> I probably triggered another wrong assumption when saying that you got 4x as much compute power from a GNa-56h (A30) compared to a GNa-14h (a quarter A30). This is very much dependent on your workload, as you correctly say. You may even create workloads, where GNa-56h is not much faster than FNa-14h.
> 
> Nevertheless, I stand by design idea that we should tell a user how much resources she gets. In one case it's 14 SMs (896 Cuda Cores for Ampere) for a quarter A30 and in the other it's 56 SMs (3584 Cuda Cores) for a full one. For a H100, this could be 132SMs (18896 Cuda Cores - full H100) vs 18SMs (2304 Cuda Cores - 1/7 H100) or 2/7 or ...
> 
> It's like vCPUs in instances -- 2 vCPU does not necessarily make your workload twice as fast, but you know how much CPU power you roughly get ...

That means you would be still in favor for the SM/CU etc. way, or?

cah-patrickthiem:
I close this PR because we finally came to a common conclusion, see this PR https://github.com/SovereignCloudStack/standards/pull/780, which by now already was merged.

Issue #489, opened 2024-02-28
Small updates to conformance test for scs-0214-v1


cah-hbaum:
We need to update the conformance test for scs-0214-v1 done in https://github.com/SovereignCloudStack/standards/pull/486 , since there were some relevant but non-critical comments.

- Remove "internal" config (https://github.com/SovereignCloudStack/standards/pull/486#discussion_r1504766932 and https://github.com/SovereignCloudStack/standards/pull/486#discussion_r1504050875)
  The config used in the code is a duplication of the one used during the input. Either we remove it and require a config to be inputed or the "internal" config is narrowed down.

- Introduce another source for labels-by-node list (https://github.com/SovereignCloudStack/standards/pull/486#pullrequestreview-1904518923)
  The labels-by-node list should be able to come from other sources and not only the `get_k8s_cluster_labelled_nodes()`

- Think about the possibility to provide some scenarios to test this test

## Definition of Done:

- [x] Test is updated with reference to the comments from https://github.com/SovereignCloudStack/standards/pull/486
- [x] PR created
- [ ] Changes have been reviewed and merged

PR #558, opened 2024-04-10
Updates to Node distribution test (#489)

cah-hbaum:
- added a new input that enables testing the test by providing yaml files containing label information for different nodes
- added "test-example.yaml" containing an example for such a test file
- removed the internal config, since the normal config file is already provided and read in as a default


cah-hbaum:
Also, like its written in the issue itself, we could think about providing some more scenarios for this test, especially with the newly added testing possibilities via files.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -225,7 +212,13 @@ async def main(argv):
...
     )
 
-    nodes = await get_k8s_cluster_labelled_nodes(config.kubeconfig, labels + ("node-role.kubernetes.io/control-plane", ))
+    if isinstance(config.testconfig, dict):
+        nodes = [v for _, v in config.testconfig.items()]
+    else:
+        nodes = await get_k8s_cluster_labelled_nodes(
+            config.kubeconfig,
+            labels + ("node-role.kubernetes.io/control-plane", )
+        )
```
martinmo, line 221: Instead of this, I would suggest to split the `main` function and have another function, e.g., `check_nodes`, that takes a `nodes` a parameter. This function can then easily be unit tested with static data instead of the dynamic "real" data from `get_k8s_cluster_labelled_nodes`.

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -99,6 +79,7 @@ def print_usage():
 The following arguments can be set:
     -c/--config PATH/TO/CONFIG         - Path to the config file of the test script
     -k/--kubeconfig PATH/TO/KUBECONFIG - Path to the kubeconfig of the server we want to check
+    -t/--test PATH/TO/YAML             - Input a formatted yaml file to test the script functionality
```
martinmo, line 82: This is a good approach, but see my other comment about splitting the main function for easier testing on the unit testing level.

With this `--test` parameter, you would need to test the script as a whole, e.g., via `subprocess.run()` and check the stdout/stderr output.

martinmo (changes requested):
Good idea with the test data. I have some suggestions about this to make it easier to unit test, see the inline comments.

Apart from that: I think it makes more sense to remove external logging configuration and the `--config` flag. It is not used anywhere and the short form `-c` also conflicts with the short form of `--os-cloud` in the IaaS scripts.

The logging config can be simplified like it is done here:

https://github.com/SovereignCloudStack/standards/blob/5e9dfd596f32e39c77c36a5ada48f0dbcfe855d8/Tests/iaas/flavor-naming/cli.py#L128-L130

Also, when I run without any arguments, the script fails.

```
$ ./kaas/k8s-node-distribution/k8s-node-distribution-check.py
The config file under ./config.yaml couldn't be found.
Traceback (most recent call last):
  File "/home/martinmorgenstern/workspace/scs-standards/Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py", line 127, in initialize_config
    with open(config.config_path, "r") as f:
FileNotFoundError: [Errno 2] No such file or directory: './config.yaml'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/martinmorgenstern/workspace/scs-standards/Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py", line 131, in initialize_config
    exit(1)
  File "/usr/lib/python3.10/_sitebuiltins.py", line 26, in __call__
    raise SystemExit(code)
SystemExit: 1

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/martinmorgenstern/workspace/scs-standards/Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py", line 244, in <module>
    return_code = asyncio.run(main(sys.argv[1:]))
  File "/usr/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "/home/martinmorgenstern/workspace/scs-standards/Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py", line 197, in main
    config = initialize_config(parse_arguments(argv))
  File "/home/martinmorgenstern/workspace/scs-standards/Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py", line 135, in initialize_config
    setup_logging(config.logging)
  File "/home/martinmorgenstern/workspace/scs-standards/Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py", line 112, in setup_logging
    logging.config.dictConfig(config_log)
  File "/usr/lib/python3.10/logging/config.py", line 811, in dictConfig
    dictConfigClass(config).configure()
  File "/usr/lib/python3.10/logging/config.py", line 375, in __init__
    self.config = ConvertingDict(config)
TypeError: 'NoneType' object is not iterable
```

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
cah-hbaum: Did some changes here https://github.com/SovereignCloudStack/standards/pull/558/commits/aec75ef610ce03dc848181b9a85fefc6354b6ed0

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
cah-hbaum: Did some changes here https://github.com/SovereignCloudStack/standards/pull/558/commits/aec75ef610ce03dc848181b9a85fefc6354b6ed0

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -108,7 +89,7 @@ def parse_arguments(argv):
     config = Config()
 
     try:
-        opts, args = getopt.gnu_getopt(argv, "c:k:h", ["config", "kubeconfig", "help"])
+        opts, args = getopt.gnu_getopt(argv, "c:k:t:h", ["config", "kubeconfig", "test", "help"])
```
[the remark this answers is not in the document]
cah-hbaum: Did some changes here https://github.com/SovereignCloudStack/standards/pull/558/commits/aec75ef610ce03dc848181b9a85fefc6354b6ed0

martinmo (changes requested):
The standard states that 

> At least one control plane instance MUST be run in each "failure zone"

I played around with the `test-example.yaml` you provided in `ipython` and created a scenario where all control plane nodes are in the same zone and region. In this case, the script would still exit with 0 and just print

```
WARNING: There seems to be no distribution across multiple regions or labels aren't set correctly across nodes.
WARNING: There seems to be no distribution across multiple zones or labels aren't set correctly across nodes.
INFO: The master nodes are distributed across 3 host-ids.
INFO: The worker nodes are distributed across 2 regions.
```

Shouldn't it produce an `ERROR`? Or do we assume the physical host is the failure zone?

Furthermore, I would like to see more scenarios like the one in the yaml file, and have them tested automatically in the CI against the expected check outcome, with `pytest`.

The config yaml template file can be removed.

cah-hbaum:
> The standard states that
> 
> > At least one control plane instance MUST be run in each "failure zone"
> 
> I played around with the `test-example.yaml` you provided in `ipython` and created a scenario where all control plane nodes are in the same zone and region. In this case, the script would still exit with 0 and just print
> 
> ```
> WARNING: There seems to be no distribution across multiple regions or labels aren't set correctly across nodes.
> WARNING: There seems to be no distribution across multiple zones or labels aren't set correctly across nodes.
> INFO: The master nodes are distributed across 3 host-ids.
> INFO: The worker nodes are distributed across 2 regions.
> ```
> 
> Shouldn't it produce an `ERROR`? Or do we assume the physical host is the failure zone?

The thing that was a bit problematic while writing the standard was that different CSPs probably have different ideas about their failure zone, especially having small CSPs in mind. So it could very well be possible that the physical hosts are dividing the cluster into failure zones.
Thats why the test doesn't throw an error here.

Personally I think thats perfectly reasonable, but if you have another opinion on that, we can discuss this further. (But that would probably also mean to change the standard again)

> Furthermore, I would like to see more scenarios like the one in the yaml file, and have them tested automatically in the CI against the expected check outcome, with `pytest`.

I'm gonna add the other files I added with some additional examples.

> The config yaml template file can be removed.
You're right.

martinmo:
> The thing that was a bit problematic while writing the standard was that different CSPs probably have different ideas about their failure zone, especially having small CSPs in mind. So it could very well be possible that the physical hosts are dividing the cluster into failure zones.
> Thats why the test doesn't throw an error here.
>
> Personally I think thats perfectly reasonable, but if you have another opinion on that, we can discuss this further. (But that would probably also mean to change the standard again)

Thanks for the clarification, I agree :+1:

cah-hbaum:
Added some more example files.

martinmo (changes requested):
Second review round:

As we discussed today in person, I've committed a pytest script that uses the scenarios in your YAML files and I added a few more (in my opinion) interesting scenarios.

One test currently fails (intentionally) because the conformance test issues a warning twice if the `host-id` label is missing where I *think* it should only do that once (that's one of the new scenarios I added). What do you think?

Furthermore, I propose to replace all occurences of `master` with `control-plane` (or just `control`), because this fits better with the current terminology in Kubernetes (`master` is outdated).

Small typo, which is not part of the diff: in line 149, it should read "the label … doesn't …" instead of "… don't …".

cah-hbaum:
> One test currently fails (intentionally) because the conformance test issues a warning twice if the host-id label is missing
> where I think it should only do that once (that's one of the new scenarios I added). What do you think?

Agree with that, that duplication shouldn't happen. I removed that in the coming commit.

I also did a general update to test testscript in order to enforce the standard requirement for labels to be provided (MUST).

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -185,68 +150,67 @@ def compare_labels(node_list, labels, node_type="master"):
...
                 label_data[key].append(node[key])
             except KeyError:
-                logger.warning(f"The label for {key.split('/')[1]}s don't seem to be set for all nodes.")
+                raise LabelException(f"The label for {key.split('/')[1]}s doesn't seem to be set for all nodes.")
 
     for label in labels:
-        if len(label_data[label]) < len(node_list):
-            logger.warning(f"The label for {label.split('/')[1]}s doesn't seem to be set for all nodes.")
+        # if len(label_data[label]) < len(node_list):
+        #     logger.warning(f"The label for {label.split('/')[1]}s doesn't seem to be set for all nodes.")
```
martinmo, line 157:
```suggestion
```

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -236,17 +179,34 @@ async def main(argv):
         if len(labelled_master_nodes) >= 1:
             worker_nodes = [node for node in nodes if "node-role.kubernetes.io/control-plane" not in node]
             # Compare the labels of both types, since we have enough of them with labels
-            compare_labels(labelled_master_nodes, labels, "master")
-            compare_labels(worker_nodes, labels, "worker")
+            compare_labels(labelled_master_nodes, "master")
```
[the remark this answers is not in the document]
cah-hbaum: Changed that part in the latest commit.

martinmo (approved):

mbuechse (approved):
Just one minor PEP-8 thing.

mbuechse (changes requested):
Ah sorry, I would like to request two changes:

- rename the script under test so that it can be imported directly
- put the test data into a single yaml file, load that once, and hand out the sections using a fixture

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -37,29 +37,17 @@
...
-        "handlers": ["console"]
-    }
-}
+
+# It is important to note, that the order of these labels matters for this test.
+# Since we want to check if nodes are distributed, we want to do this from bigger
+# infrastructure parts to smaller ones. So we first look if nodes are distributed
+# across regions, then zones and then hosts. If one of these requirements is fulfilled,
+# we don't need to check anymore, since a distribution was already detected.
+labels = (
```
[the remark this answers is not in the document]
cah-hbaum: Changed it to `LABELS`

Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py
```diff
@@ -76,10 +64,12 @@ class DistributionException(BaseException):
     """Exception raised if the distribution seems to be not enough"""
 
 
+class LabelException(BaseException):
+    """Exception raised if a label isn't set"""
+
+
 class Config:
```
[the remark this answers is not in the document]
cah-hbaum: I'm gonna keep that for now.

cah-hbaum:
> Ah sorry, I would like to request two changes:
>
> * rename the script under test so that it can be imported directly
> * put the test data into a single yaml file, load that once, and hand out the sections using a fixture

[c60f895](https://github.com/SovereignCloudStack/standards/pull/558/commits/c60f89560cd24d0c64e644f46dd19fddba4c4ce0) should have the requested changes.

cah-hbaum:
> @cah-hbaum Thanks! Sorry for the long delay on my end. I think you need to update https://github.com/SovereignCloudStack/standards/blob/main/Tests/scs-compatible-kaas.yaml as well, no? Apart from that, I would prefer if the yaml file were loaded only once, which should be possible with pytest, if you use fixtures, or if you just do it on the module level. But it's not a deal breaker.

I updated the `scs-compatible-kaas.yaml` and introduced fixtures (I hope thats how you meant it).

mbuechse (approved):
LGTM ;)

cah-hbaum:
Waiting on https://github.com/SovereignCloudStack/standards/pull/524, will be merged afterwards.

PR #567, opened 2024-04-16
Add standard for volume backup functionality

markus-hentsch:


markus-hentsch:
Proposal written. Test script implemented.

Cleanup function of the test script needs some polish according to some erronous behavior I observed:

- wait for resources to be in available state before attempting deletion
- wait for backups to finish being cleaned up before starting volume deletion (dependency)
- make errors in cleanup fatal for test execution (resoning: deletion should work just as fine as creation when API works correctly and we need a clean base to test on)

Standards/scs-XXXX-v1-volume-backup-service.md
```diff
@@ -0,0 +1,71 @@
...
+With this, users can leverage functionalities like incremental backups and benefit from optimized performance of the backup process due to the tight integration with the volume service.
+
+## Decision
+
+This standard decides to go with the second option and makes the volume backup feature mandatory in the following way:
+
+In an SCS cloud, the volume backup functionality MUST be configured properly and its API as defined per `/v3/{project_id}/backups` MUST be offered to customers.
+If using Cinder, a suitable [backup driver](https://docs.openstack.org/cinder/latest/configuration/block-storage/backup-drivers.html) MUST be set up.
+
+The volume backup target storage SHOULD be a separate storage system from the one used for volumes themselves.
```
josephineSei, line 59: Would there be a benefit of using volume backups over snapshots or the simple replication, when it is stored in the same backend?

And how do we address small deployments with only one rack (e.g. in edge cloud scenarios)? Would a second separate storage be reasonable for such deployments?

josephineSei (commented):
WE do have the problem that we should have both in mind: large deployments and edge cloud scenarios with only one rack somewhere. I am a bit undecided how to deal with a possible second storage backend in the latter case.

Standards/scs-XXXX-v1-volume-backup-service.md
markus-hentsch: > Would there be a benefit of using volume backups over snapshots or the simple replication, when it is stored in the same backend?

`openstack volume backup restore` has a `--force` flag that enables users to overwrite existing volumes with the backup, which you cannot do with a snapshot afaik. However, there is not much benefit to it since the target volume must be in "available" status and must not be attached to a server even if the server is shutdown. Also, the backups are not bound to the existence of the source volume and the source volume can be deleted while backups being preserved in contrast to snapshots.
I don't think it makes *that* much of a difference in real-world scenarios though.

markus-hentsch: > And how do we address small deployments with only one rack (e.g. in edge cloud scenarios)? Would a second separate storage be reasonable for such deployments?

Good point. I guess this could be a valid reason for this "SHOULD" to be disregarded by a CSP.

markus-hentsch:
> Proposal written. Test script implemented.
> 
> Cleanup function of the test script needs some polish according to some erronous behavior I observed:
> 
>     * wait for resources to be in available state before attempting deletion
> 
>     * wait for backups to finish being cleaned up before starting volume deletion (dependency)
> 
>     * make errors in cleanup fatal for test execution (resoning: deletion should work just as fine as creation when API works correctly and we need a clean base to test on)

Done and tested.

Standards/scs-XXXX-v1-volume-backup-service.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Related Documents
+
+- [OpenStack Block Storage v3 Backup API reference](https://docs.openstack.org/api-ref/block-storage/v3/index.html#backups-backups)
+
+## Conformance Tests
+
+Conformance tests include using the `/v3/{project_id}/backups` Block Storage API endpoint to create a volume and a backup of it as a non-admin user and subsequently restore the backup on a new volume while verifying the success of each operation.
```
anjastrunk, line 67: This does only check MUST-requirementa of this standard. But we have a SHOULD-requirement to: Operating a separate storage backend for backup. As we can not test this automatically, check for additional storage backend is not part of python script. However, we should mention that separate storage backends are not validated in conformance tests in section "Conformance Tests" in order to be transparent for SCS customers. Furthermore, I suggest to create a new issue to collect all manual tests and to decide to proceed with manual audits.

anjastrunk (approved):
LGTM in general. Just one addition requested in section "Conformance Tests".

Standards/scs-XXXX-v1-volume-backup-service.md
josephineSei: I still wonder, how a customer can differentiate between a deployment, that has a different store  for the volume backups and a deployment, that does not fulfill this.
Is that easy to determine for a customer?
I don't want users to accidentally think their volume can be backuped, now everything is fine, when volume and backup are stored in the same storage.

josephineSei (commented):
I think only this one question is left for me.

Standards/scs-XXXX-v1-volume-backup-service.md
markus-hentsch: > Furthermore, I suggest to create a new issue to collect all manual tests and to decide to proceed with manual audits.

Good point. It has been coming up a lot recently, so I took the initiative:

https://github.com/SovereignCloudStack/issues/issues/668

Standards/scs-XXXX-v1-volume-backup-service.md
```diff
@@ -0,0 +1,71 @@
...
+It would make the volume backup feature mandatory for SCS clouds.
+This way users can expect the feature to be available and usable.
+
+With this, users can leverage functionalities like incremental backups and benefit from optimized performance of the backup process due to the tight integration with the volume service.
+
+## Decision
+
+This standard decides to go with the second option and makes the volume backup feature mandatory in the following way:
+
+In an SCS cloud, the volume backup functionality MUST be configured properly and its API as defined per `/v3/{project_id}/backups` MUST be offered to customers.
```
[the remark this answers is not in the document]
markus-hentsch: I think "an" is correct here. The first "S" in "SCS" produces a vowel sound ("ess"), hence the "an", see https://oneminuteenglish.org/an-s-or-a-s-which-is-correct/

berendt:
> I think only this one question is left for me.

Wouldn't it even be a good idea to require backups to be stored on a different storage backend? Then there would always be a "Medienbruch".

So I would prefer that the backup function itself does not necessarily have to be available, but if it is available it always has a dedicated storage backend.

Then you can assume that if the function is available, it also enables real backups.

neuroserve:
Just a quick feedback we got from a customer accustomed to VMWare: The backup functions (provided by Veeam) are integrated in the webfrontend (VCloud Director) there. And that's why they were looking for similar functions in Horizon, too.

markus-hentsch:
> > I think only this one question is left for me.
> 
> Wouldn't it even be a good idea to require backups to be stored on a different storage backend? Then there would always be a "Medienbruch".
> 
> So I would prefer that the backup function itself does not necessarily have to be available, but if it is available it always has a dedicated storage backend.
> 
> Then you can assume that if the function is available, it also enables real backups.

Very good point. I added a new choice to the "Options considered" section. I think it boils down to whether we want to favor feature availability vs. backup reliability.

Personally, I'd actually favor reliability now that I think about it due to the fact the term "backup" does imply some kind of availability guarantees.

neuroserve:
AFAIK the Veeam solution is quite flexible and can use "any" S3-Backend (for example). The backup destination is _not_ tied to the infrastructure, where the VCloud Director is located.

Standards/scs-XXXX-v1-volume-backup-service.md
```diff
@@ -0,0 +1,89 @@
...
+
+This approach would focus on feature availability rather than backup reliability.
+
+#### Focus on backup reliability, make separate backend mandatory
+
+As an alternative, the volume backup feature availability could be made optional but in case a CSP chooses to offer it, the standard would mandate a separate storage backend to be used for volume backups.
+This way, failures of the volume storage backend would not directly impact the availability and safety of volume backups, making them actually live up to their name.
+
+In contrast to the above, this approach would focus on backup reliability rather than feature availability.
+
```
josephineSei, line 69: Speaking from a users perspective: I want to have a backup, that I can definitely use, when the original is not there anymore.
For volumes this would mean, that the volume backup should be in a different backend, in case of a backend failure or even an AZ outage.

On the contrary this could lead to deployments with only one Ceph backend, which spans over three fire zones and is configured with internal replication between those zones, would not be allowed to offer the volume backup feature.

I think we should discuss this with CSPs and in the IaaS call.

markus-hentsch: > I think we should discuss this with CSPs and in the IaaS call.

Agreed. Added it to the next IaaS call.

Standards/scs-XXXX-v1-volume-backup-service.md
markus-hentsch: > I still wonder, how a customer can differentiate between a deployment, that has a different store for the volume backups and a deployment, that does not fulfill this.
> Is that easy to determine for a customer?

Sadly not. This is a `cinder.conf` thing only and not visible to the API user or customer. Either the standard will enforce it so that the guarantee is there if the CSP is compliant or we need some kind of self-description.

> I don't want users to accidentally think their volume can be backuped, now everything is fine, when volume and backup are stored in the same storage.

That's the same feature availability vs. backup reliability topic which came up recently I guess. Added this to the next IaaS call agenda.

Standards/scs-XXXX-v1-volume-backup-service.md
markus-hentsch: I added a note to the Conformance Tests section.

markus-hentsch:
The [discussion in the IaaS call on 2024-07-03](https://github.com/SovereignCloudStack/minutes/blob/main/iaas/20240703.md#cinder-volume-backup-markus-hentsch) did not produce a clear winner among the options (feature availability vs. backup reliability).
However, a third option was proposed:

> - option c)
>     - mandate backup feature and API availability AND make it transparent to users ("discoverability") how much additional reliability is provided
>         - place of discoverability is not obvious (no place for metadata in the cinder backup API)
>             - api extensions?
>             - Gaia-X ~~self-descriptions~~VCs
>             - new discoverability service that we work on introducing for flavors (-> public cloud SIG draft spec) [spec](https://review.opendev.org/c/openstack/publiccloud-sig/+/909387)

... which also favors feature availability while trying to make the potential shortcomings in backup reliability transparent for the user in case a separate storage backend is not configured for backups by the CSP.
This however requires some kind of self-description functionality as stated, which we do not have yet.

There has been no further feedback since then. The current iteration of the standard draft implements the feature availability option, which also provides the basis for the newly suggested option described above. For now, I will keep it that way.

Standards/scs-XXXX-v1-volume-backup-service.md
```diff
@@ -0,0 +1,94 @@
...
+OpenStack offers a variety of resources where users are able to transfer and store data in the infrastructure.
+A prime example of these resources are volumes which are attached to virtual machines as virtual block storage devices.
+As such they carry potentially large amounts of user data which is constantly changing at runtime.
+It is important for users to have the ability to create backups of this data in a reliable and effifcient manner.
+
+## Terminology
+
+| Term | Meaning |
+|---|---|
+| CSP | Cloud Service Provider, provider managing the OpenStack infrastructure |
```
josephineSei, line 19: Images and Volumes are referred throughout this document. Maybe they should be also explained here, so people always can look up the differences.

Standards/scs-XXXX-v1-volume-backup-service.md
```diff
@@ -0,0 +1,94 @@
...
+In contrast to the above, this approach would focus on backup reliability rather than feature availability.
+
+## Standard
+
+This standard decides to go with the second option and makes the volume backup feature mandatory in the following way:
+
+In an SCS cloud, the volume backup functionality MUST be configured properly and its API as defined per `/v3/{project_id}/backups` MUST be offered to customers.
+If using Cinder, a suitable [backup driver](https://docs.openstack.org/cinder/latest/configuration/block-storage/backup-drivers.html) MUST be set up.
+
+The volume backup target storage SHOULD be a separate storage system from the one used for volumes themselves.
```
josephineSei, line 77: As it is not possible to know as a user, whether a volume backup storage is different from the normal volume storage. We should encourage CSPs to give that information to customers. This could also be done via gaiax credentials, couldn't it? Maybe we need another issue for this part.

josephineSei (commented):
This is looking good. Most comments are for the reference of OpenStack-sepcific terms, where we might be more broad to also integrate other implementations of the APIs.

Standards/scs-XXXX-v1-volume-backup-service.md
markus-hentsch: Good point. Added.

Standards/scs-XXXX-v1-volume-backup-service.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Terminology
+
+| Term | Meaning |
+|---|---|
+| CSP | Cloud Service Provider, provider managing the OpenStack infrastructure |
+
+## Motivation
+
+The [volume backup functionality of OpenStack](https://docs.openstack.org/cinder/latest/admin/volume-backups.html) is a feature that is not available in all OpenStack clouds per default.
```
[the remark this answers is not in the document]
markus-hentsch: Rephrased.

Standards/scs-XXXX-v1-volume-backup-service.md
```diff
@@ -0,0 +1,94 @@
...
+The feature requires a backend to be prepared and configured correctly before it can be used.
+In Cinder, this is a separate configuration to the general storage backend of the volume service and is not mandatory.
```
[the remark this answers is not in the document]
markus-hentsch: I rephrased this a bit more in the latest commit because if you remove the direct mention of OpenStack/Cinder here, then the statement cannot be as strict since other implementations may behave differently. Thus, I weakened the statement a bit accordingly.

Standards/scs-XXXX-v1-volume-backup-service.md
```diff
@@ -0,0 +1,94 @@
...
+Thus, an arbitrary OpenStack cloud may or may not offer this feature.
```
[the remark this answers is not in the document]
markus-hentsch: Done.

Standards/scs-XXXX-v1-volume-backup-service.md
```diff
@@ -0,0 +1,94 @@
...
+This standard aims to make this functionality the default in SCS clouds so that customers can expect the feature to be usable.
+
+## Design Considerations
+
+The standard should make sure that the feature is available and usable but should not limit the exact implementation (e.g. choice of backend driver) any further than necessary.
+
+### Options considered
+
+#### Only recommend volume backup feature, use images as alternative
+
```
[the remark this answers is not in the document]
markus-hentsch: Done.

Standards/scs-XXXX-v1-volume-backup-service.md
markus-hentsch: This would require a deep dive into the [Gaia-X Ontology](https://gaia-x.gitlab.io/gaia-x-community/gaia-x-self-descriptions/service/service.html) to figure out whether this can be expressed using Gaia-X Credentials (formerly Self-Descriptions) appropriately.

As long as we don't have a standardized and proven way of representing arbitrary self-description information about an SCS cloud and its services in Gaia-X Credentials, I'm hesitant to add any such suggestions to the standard yet.

josephineSei: Yeah, I thought so - Maybe we can discuss in the next Standardization SIG meeting or tomorrow, if and how we can use the gaia-x credentials for tests.

josephineSei (approved):
This looks good to me.
Someone from a CSP or from the IaaS team should also approve.

PR #572, opened 2024-04-19
Add standard for provider networks

kgube:
This is a WIP for a standard covering provider networks and public IP allocation.
https://github.com/SovereignCloudStack/issues/issues/166
https://github.com/SovereignCloudStack/issues/issues/167


Standards/scs-xxxx-v1-networking-standard.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Options considered
+
+#### API Extensions
+
+The OpenStack Networking API is more modular than many other OpenStack APIs, adding new functionality as optional API extensions, rather than a linear sequence of micro-versions. Which of those extensions are available from Neutron depends on the enabled plugins and back-end drivers.
+
+Mandating a minimum set of enabled extensions will ensure that users have a consistent feature-set available across SCS clouds, but should not unnecessarily restrict the choice of back-end drivers for CSPs.
+
+[...]
```
kgube, line 30: Which extensions should we consider as mandatory?
We could determine the extensions necessary for the other requirements of this standard, e.g.:
* subnetpools
* subnet_allocation
* ...

But then we could also just state this as a requirement without listing them all, so there wouldn't be the problem of potentially missing some.

Standards/scs-xxxx-v1-networking-standard.md
```diff
@@ -0,0 +1,88 @@
...
+
+#### Policies
+
+[...]
```
kgube, line 34: One approach might be to mandate policies to be equal or stricter than the neutron default policies, (with possible exemptions for CSP-internal projects/users). Is something like this already part of the standards? should it be?

Standards/scs-xxxx-v1-networking-standard.md
```diff
@@ -0,0 +1,88 @@
...
+The external network **MUST** have an IPv6 subnet.
+CSPs **MUST** provide a subnet pool for the allocation of at least one public /64 IPv6 prefix per project.
+
+The external network **SHOULD** have an IPv4 subnet.
+If CSPs provide an IPv4 subnet, then CSPs **MUST** at least one public Floating IP per project.
+They **SHOULD** also provide a subnet pool for the allocation of IPv4 prefixes to project networks.
+
+CSPs **MUST** provide dynamic routing for all project-allocated IP-prefixes.
+
+All subnets of external networks **MUST** be configured with `--no-dhcp`.
```
kgube, line 88: This is probably only relevant for shared external networks, which allow direct attachment of VMs

kgube:
We should probably just focus the standard on external provider networks/subnets, because that is ultimately what facilitates public IP networking, as https://github.com/SovereignCloudStack/issues/issues/167 and https://github.com/SovereignCloudStack/issues/issues/166 are calling for.

I have been trying to find a way to meaningfully incorporate the other suggestions from #522 (and the comments), like OVN/L3-router-HA and API-extensions/plugins and port security and policies, but they are either orthogonal to the issues (like the HA-topic), or follow from the provider-network requirements (like port security and, at least partially, the API-extensions), or should be discussed somewhere else more broadly (also the HA topic, and policies, specifically codification of/deviation from default policies).

Standards/scs-xxxx-v1-provider-network-standard.md
```diff
@@ -0,0 +1,200 @@
...
+| Prefix | IP address prefix of a given bit-length N, written _/N_. Divides addresses into a network and a host part, a shorter prefix allows more hosts but takes up more address space. |
+| NAT | Network Address Translation: mapping (usually public) IPv4 addresses onto a different (usually private) address space. May allows multiple hosts to share a public address by multiplexing TCP/UDP ports. |
+| RBAC | Role-based Access Control: A mechanism in the Network API to give projects limited access to resources owned by other projects. Typically used by CSPs to create provider networks. |
+| Shared Network | Virtual network that is shared between projects in a way that allows direct attachment of servers. |
+| External Network | Virtual network that is shared between projects in a way that only allows virtual routers to use it as external gateway. Typically used by CSPs to provide access to networks outside of the cloud environment. |
+| Provider Network | Any shared or external network that is managed by the CSP. |
```
horazont, line 37: OpenStack provider networks have a different property, namely that they're tied to non-virtualised networks. I think that property is relevant enough to include it in this definition.

(Or are we using the term "provider network" in a different sense here?)

Standards/scs-xxxx-v1-provider-network-standard.md
```diff
@@ -0,0 +1,200 @@
...
+This section will provide some general background on OpenStack provider networks before considering specific options to standardize. 
+
+### Provider Network Access Control
+
+In OpenStack, ownership of resources is generally tracked through projects, and, as per default policy, only members of a project have access to its resources
+This is also true for CSP-managed resources, such as provider networks, which have to be created in a CSP-internal project, and are initially only accessible in this project.
+
+The Network API's Role Based Access Control (RBAC) extension can then be used to share it with other projects.
+RBAC rules for networks support the two actions `access_as_external` and `access_as_shared`, and can be created automatically on `openstack network create` with the options `--external` and `--share`.
+* `access_as_external` allows networks to be used as external gateway for virtual routers in the target projects. Such networks are in the following referred to as _external networks_.
```
horazont, line 56: Another notable thing `access_as_external` allows is creation of floating IPs

Standards/scs-xxxx-v1-provider-network-standard.md
```diff
@@ -0,0 +1,200 @@
...
+The documentation at [^pd] still marks prefix delegation as an experimental feature in Neutron and notes low test coverage.
+
+For IPv4, OpenStack virtual routers support source NAT, allowing all servers in the internal subnet to access the external network with the gateway IP of the virtual router.
+They also support destination NAT in the form of floating IPs, addresses from the external network that can be mapped onto specific servers in the internal subnet to make them externally accessible.
+Floating IPs are allocated from a CSP-managed pool, and can be controlled by a per-project quota.
+There is also a set of API extensions that allow more fine grained port-forwarding, mapping different TCP or UDP ports of a floating IP to different internal IP addresses [^pf].
+
+### Port Security and Spoofing
+
+OpenStack networks have the flag `port_security_enabled`, that is set to true by default and can only be changed by it's owner.
```
horazont, line 83: TIL `port_security_enabled` as a network-level flag, but it only controls the *default* for the port-level flag of the same name which has the functionality you describe.

I suggest a rewording:

> OpenStack ports have the flag `port_security_enabled`, that is set to true by default and can only be changed by the owner *of the corresponding network*. The default value of that flag is controlled by a `port_security_enabled` flag on the network.

Standards/scs-xxxx-v1-provider-network-standard.md
```diff
@@ -0,0 +1,200 @@
...
+In Neutron, besides enabling security groups for ports in this network, it also enables a number of built-in spoofing protections.
```
horazont, line 84: Likewise:

> Besides allowing to use security groups for a port, setting `port_security_enabled` to true will also enable a number of built-in spoofing protections.

Standards/scs-xxxx-v1-provider-network-standard.md
```diff
@@ -0,0 +1,200 @@
...
+With the FWaaS API extensions, they can also assign firewall rules to the virtual router, to control which traffic can pass between internal and provider networks.
+
+As described above, there are multiple methods for allocating public IP addresses to project-internal networks.
+For IPv6, the currently best option seems to be subnet allocation from a CSP-managed subnet pool, because support for Prefix Delegation is still experimental.
+For IPv4, NAT and floating IPs are generally preferred over subnet allocation because of scarcity of IPv4 address space.
+
+#### NAT and Floating IPs
+
+Subnets can only have a size that is a power of two, and will thus usually be oversized for the project.
+They also need to reserve one address for the gateway and, if DHCP is used, one or more addresses for DHCP service ports.
```
horazont, line 137: > one or more addresses for DHCP service ports.

Is that still true with OVN? (I don't actually know, but given how OVN messes with DNS, I wouldn't be surprised if it wasn't.)

horazont (commented):
Overall, this looks good to me. There are a few issues and open questions for me, which I noted below.

Standards/scs-xxxx-v1-provider-network-standard.md
kgube: This seems to still be true for OVN, though the owner of the port is "network:distributed" instead of "network:dhcp".
I do suspect that they have dropped support for dhcp HA though, because that would be kinda pointless.

Standards/scs-xxxx-v1-provider-network-standard.md
kgube: In many places, including some [Openstack documentation](https://docs.openstack.org/install-guide/launch-instance-networks-provider.html), I found the term provider network to be used (interchangeably with "external network") to mean CSP-managed networks that are tied to non-virtualized networks, because that is usually what they are used for.

I was looking at this from more of an RBAC-perspective though, where any network, regardless of it's `provider:network_type` and `provider:physical_network` can have an `access_as_external` rule, and where ownership and RBAC rules ultimately dictate how a network can be used in a project. In that context, it made more sense to me to define provider networks literally as "networks managed by the provider".
This will still cover the classical use-case of bridging into non-virtualized networks, but also use-cases like a CSP-managed shared network.

It's still a deviation from common Openstack terminology of course, and I can try rewording it to stay closer to that.

kgube:
The was a short discussion in the community call today on the question of mandatory internet access vs. air-gapped infrastructure that was raised in the IaaS call yesterday.
The feedback was that we also want to support standardization of private and air-gaped clouds, and thus we should not mandate internet-access.

I will try to reword the standard to reflect this.

Standards/scs-xxxx-v1-provider-network-standard.md
horazont: Okay thanks.

Standards/scs-xxxx-v1-provider-network-standard.md
horazont: > In that context, it made more sense to me to define provider networks literally as "networks managed by the provider".

I agree. My main problem with the definition as written above is that the "managed by the CSP" gets overpowered by the "shared or external" attributes, which make me read that definition as if the key point was that the network was "shared or external", not "managed by the CSP".

Maybe: "A network managed by the CSP, typically connected to non-virtualized infrastructure and commonly made external or shared."

kgube: The network being shared or external is kinda essential though, as that is what makes it a "provider network" to tenants, otherwise it's just invisible to them and might as well be non-existent.

I have changed it now to "A CSP-managed virtual network made available to projects as either shared or external, typically connected to non-virtualized infrastructure."

Standards/scs-xxxx-v1-provider-network-standard.md
kgube: I have added a mention of that.

Standards/scs-xxxx-v1-provider-network-standard.md
kgube: done

Standards/scs-xxxx-v1-provider-network-standard.md
kgube: done

kgube:
Finding a common ground regarding provider networks between public, private, and specifically air-gaped clouds is hard, even with individually scoped rules. There is some overlap, but the requirements are very different.

I have now given up on that and made some changes to the draft to limit the standard to clouds that provide public IPs.
This still has some issues, though, like the awkwardness of mandating public IPv6 support when the overall support of public IP addresses is not mandatory.

matfechner:
@kgube overall it looks good to me, maybe it makes sense, besides  neutron-dynamic-routing,  to refer also to ovn-bgp-agent. because OVN will perhaps take over this task in the future

matfechner (approved):
LGTM

Standards/scs-xxxx-v1-provider-network-standard.md
```diff
@@ -0,0 +1,213 @@
...
+track: IaaS
+---
+
+## Introduction
+
+Many use-cases of IaaS require virtual servers to be able to connect to network resources outside of the cloud environment, often to the internet.
+Openstack supports this by allowing CSPs to map non-virtualized networks onto specific virtual networks, such that virtual routers and servers can connect to them.
+
+Such networks will usually be created in a provider-managed project and then shared to user projects using the RBAC-feature of the network API.
+Because they have to be set up by the cloud provider, networks of this type are generally referred to as _provider networks_, though that term can also be used in a broader senseto refer to other types of CSP-managed virtual networks.
```
josephineSei, line 14:
```suggestion
Because they have to be set up by the cloud provider, networks of this type are generally referred to as _provider networks_, though that term can also be used in a broader sense to refer to other types of CSP-managed virtual networks.
```

Standards/scs-xxxx-v1-provider-network-standard.md
```diff
@@ -0,0 +1,213 @@
...
+| Virtual Router | OpenStack resource that can be used to route and bridge between virtual networks and provide other features such as NAT |
+| Subnet | Subdivision of available IP address space using address prefixes. In OpenStack also an abstraction for controlling IP address allocation in a virtual network. |
+| DHCP | Dynamic Host Configuration Protocol: Used to automatically configure hosts in a network with IP addresses, default routes, and other information such as DNS servers. |
+| Prefix | IP address prefix of a given bit-length N, written as _ADDRESS/N_. Divides addresses into a network and a host part, a shorter prefix allows more hosts but takes up more address space. |
+| NAT | Network Address Translation: mapping (usually public) IPv4 addresses onto a different (usually private) address space. May allows multiple hosts to share a public address by multiplexing TCP/UDP ports. |
+| RBAC | Role-based Access Control: A mechanism in the Network API to give projects limited access to resources owned by other projects. Typically used by CSPs to create provider networks. |
```
josephineSei, line 33: This maybe a little bit nitpicking:
When I think of RBAC in OpenStack, I first think about user roles and access to projects. While we also have access to projects here, it has nothing to do with user roles and is a very Neutron specific thing.
Could we maybe change the wording a bit? Maybe to "network RBAC" like in the openstack CLI command?

Also this would have to be adapted overall in this standard.

Standards/scs-xxxx-v1-provider-network-standard.md
```diff
@@ -0,0 +1,213 @@
...
+If they do offer public IPv4 addresses, they **MUST** provide at least one public Floating IP per project, but **MAY** also provide a subnet pool for the allocation of public IPv4 prefixes to project networks.
+
+CSPs **MUST** externally route any public IP addresses allocated from subnets of the standard provider network.
+CSPs **MUST** provide dynamic routing for all project-allocated public IP-prefixes via the standard provider network.
+
+By default, users **SHOULD** be prohibited by policy from creating RBAC rules for networks in their projects, to prevent the creation of faux provider networks.
+
+## Conformance Tests
+
+(TBD, current requirements should mostly be testable by API. Testing external routing is more tricky and will require external testing infrastructure of some sort)
```
josephineSei, line 205: I will review again, when tests are implemented.

kgube:
While working on the conformance tests I stumbled on Openstacks network auto-allocation feature:
https://docs.openstack.org/neutron/latest/admin/config-auto-allocation.html

I'm not sure how I missed that previously, but I am going to update the standard to conform to the auto-allocation requirements.

Standards/scs-xxxx-v1-provider-network-standard.md
kgube: I changed it to "Networking RBAC" now, because Networking is the name of the API, and there is less chance for confusion when talking about RBAC rules for networks.

Standards/scs-xxxx-v1-provider-network-standard.md
```diff
@@ -0,0 +1,233 @@
...
+If CSPs offer public IP addresses at all, they **MUST** provide a shared subnet pool for the allocation of at least one public /64 IPv6 prefix per project.
+This subnet pool **MUST** have the `is_default` flag set to True, and it **MUST** be the only IPv6 subnet pool to do so.
+If CSPs offer public IP addresses, they **SHOULD** also offer public IPv4 addresses.
+If they do offer public IPv4 addresses, they **MUST** provide at least one public Floating IP per project, but **MAY** also provide a shared subnet pool for the allocation of public IPv4 prefixes to project networks.
+If CSPs offer a subnet pool for the allocation of public IPv4 prefixes, it **MUST** have the `is_default` flag set to True, and it **MUST** be the only IPv4 subnet pool to do so.
+
+CSPs **MUST** externally route any public IP addresses allocated from subnets of the standard provider network.
+CSPs **MUST** provide dynamic routing for all project-allocated public IP-prefixes via the standard provider network.
+
+By default, users **SHOULD** be prohibited by policy from creating Networking RBAC rules, to prevent the creation of faux provider networks.
```
josephineSei, line 220: Can you add an implementation note that shows the necessary policy change?

kgube:
There haven't been much updates here, even though I have continuously been trying to build compliance tests for this.
I have been restructuring the code back and forth, without ever really reaching a runnable state (and I don't like pushing code that is not at least able to run). Some aspects of the standard that are just hard to test, some of the requirements are too loosely defined, some are kinda circular, and some I am just personally unhappy with.

I will try to identify those issues and approach them more systematically:

#### The scope of the standard.
   * Currently, the requirements are limited to "CSPs [that] offer public IP addresses to projects". However, whether a CSP is _supposed_ to be offering public IP addresses is impossible to determine for the compliance tests.
   * The current wording would also include private clouds that are offering public IPs for some selected projects, and would force them to offer public IPs to all of their projects.
   * Standardized IP allocation would also be useful for private clouds, offering private IPs reachable via VPN, which are currently excluded from the standard. This would also aid the development of compliance tests, as they could be run against a private test environment.

I now think that the better approach is to not generally mandate public IP addresses, but to define _external IP addresses_ as addresses reachable by a user with access to the openstack API, such that e.g. automated tooling can talk directly to both API and VMs. CSPs offering public cloud services can then specifically be required to provide public IPs as external IPs. This will make the standard more useful and allow easier compliance testing, at least for the basic requirements.
The requirement for public IP addresses could be enabled by an optional CLI flag or environment variable for the test script. 

#### Support for multiple provider networks and/or multiple shared subnet pools per IP version.
   * The standard does not forbid this (though its not recommended), and I'm not sure we should forbid this. However, compliance-testing would mean potentially trying multiple subnet-pool/provider-network combinations until we find one that meets our requirements.
   * This is mostly a hypothetical scenario, but if the standard allows it, the tests should be able to cover it. To test this scenario, we need to build the tests around this scenario, it is not something we can easily add later.
   * I tried adding a requirement for setting the _is_default_ flags of the network auto-allocation feature on the mandated resources to circumvent this, but when discussing this in last week's IaaS call got the suggestion to leave that as a recommendation until the SCS reference implementation has support for this and providers have gotten a chance to test it.

There may be other parts of the standard that are not yet part of the reference implementation and may need a grace period (IPv6 dynamic routing?). It is much more useful to keep the requirement of the _is_default_ flag in the standard and treat the whole standard as a recommendation until it has been properly verified. This is especially true because all of the requirements for auto-allocation (at least for IPv6) are already mandated by the standard, the only addition really is the flag.
So in the interest of getting done with the tests, that is how I will proceed.

#### Compliance tests are limited to the project perspective.
   *  The Standard makes some global requirements, like the single standard provider network available to all projects, but the compliance tests can only check it from the perspective of a single project and don't know if it is actually available to all projects.
   * This is not a big problem, in fact it probably makes no difference to users if their standard provider network is different from the one of other users, as long as it performs the same function. But It might be cleaner if that was reflected in the wording of the standard.

This is not a huge issue and does not actually affect the tests themselves (because this is the only way to do them, in this regard).
I will approach this if I have time for it, and only when I'm done with the compliance tests.

#### Limited feedback on the standard.
   * Not really an issue of the standard itself, but something that has made me very cautious in working on the tests has been the limited feedback from CSPs on the standard.
   * Some of the decisions have been contentious in previous discussions, such as making IPv6 mandatory and IPv4 only recommended. It is not unlikely that parts of the requirements will change later, and the compliance tests with them.
  
There isn't really much I can do about this. I will try to build the compliance tests somewhat modular, but not more than aids general readability. There is no point in considering potential future test scenarios if we aren't there yet, and refactoring will be easier than preparing, in this case.

Standards/scs-xxxx-v1-provider-network-standard.md
josephineSei: I clicked the link and unfortunately this is not good documented by OpenStack itself. It my be better to attach an implementataion note and not just write in block text but have the policy change presented in a clear readable way. Otherwise people might overread it.. I have something in mind like:

### Policy adjustment for network rbac

OpenStack itself allows all user, regardless of role to change the accessability of network resources (e.g. networks, routers, security groups, etc). This allows malicious users to present networking resources to other client as coming from the CSP. To forbid this the following policy adjustments SHOULD be implemented in the `policy.yaml` file:

```
"create_rbac_policy": "rule:admin_only"
```

Standards/scs-xxxx-v1-provider-network-standard.md
```diff
@@ -199,25 +201,42 @@ So, just specifying mandatory features rather than a certain set of extensions m
...
+This subnet pool **MUST** have the `is_default` flag set to `True`, and it **MUST** be the only shared IPv6 subnet pool in the project with `is_default` set to `True`.
+
+### IPv4 Allocation
 
-If CSPs offer public IP addresses at all, they **MUST** provide a shared subnet pool for the allocation of at least one public /64 IPv6 prefix per project.
-This subnet pool **MUST** have the `is_default` flag set to True, and it **MUST** be the only IPv6 subnet pool to do so.
-If CSPs offer public IP addresses, they **SHOULD** also offer public IPv4 addresses.
-If they do offer public IPv4 addresses, they **MUST** provide at least one public Floating IP per project, but **MAY** also provide a shared subnet pool for the allocation of public IPv4 prefixes to project networks.
-If CSPs offer a subnet pool for the allocation of public IPv4 prefixes, it **MUST** have the `is_default` flag set to True, and it **MUST** be the only IPv4 subnet pool to do so.
+If the CSP offers external IPv4 addresses, they **MUST** provide at least one external Floating IP per project that can be allocated from the standard provider network.
```
josephineSei, line 232: Agree to name it external instead of pulic to include clouds which are only used internally / privately.

josephineSei (commented):
This part looks good to me so far. So its just the test and maybe implementation notes missing (see my comment on that)

Standards/scs-xxxx-v1-provider-network-standard.md
kgube: Alright, I put this in an implementation notes supplement now.

josephineSei (commented):
LGTM only tests are missing.

fkr:
I've reached out to a set of CSPs. One gave qualified feedback as in: the only limiting factor is the mandatory IPv6 prefix. 
My suggestion is we merge this now as Draft and advocate for its adoption among the CSPs.

fkr:
More feedback from another CSP: seems solid so far.

PR #577, opened 2024-04-24
Add draft for VM clock synchronisation recommendations

kgube:


scoopex:
Had a discussion with @kgube about the motivation , goals and the contents of this DR.

Input/framework conditions that may be useful:

- Its good to add reference to software-systems which might be used by cusomers that use shared quorum algorithms and reference to the relevance of good system time
  (Zookeper, RabbitMQ, ETCD, Consul, Hazelcast, Ceph)
- SCS environments itself should be operated with at least 3 central and CSP-local NTP sources (for Ceph, RabbitMQ, ...)
- Whether overcommit or that a VM is not “scheduled” plays a role for the quality of the time synchronization with the virtualization used must not matter to the user
- The CSP offers at least three local and not rate limited NTP servers that have at least 5 statically defined upstream stratum servers or local time sources with high quality
- We can define a minimum quality that is based on the requirements of common systems and provides some reserve to keep popular systems running without problems (offset, jitter, frequency drift, ...)
- The CSP ensures that a time with a minimum quality can be maintained in VMs with a reference setup
    - defined chrony setup/configuration that uses the min. 3 CSP NTP servers
    - this should be possible with all flavors (in some virtualization technologies the size of the virtual machine has impact to the scheduling of it and related to that to its time sychronization
    - the health check service activates several VMs with a single defined flavor distributed across the CSP landscape (e.g. 3) that run permanently and checks their quality to evaluate the compliance
- Subordinate, but exciting would be a idea how to provide the flavor images with a standardized setup by default which can be used independent from the CSP (e.g. by using a standardized setup mechanism,  or standardized references to the servers)

Standards/scs-XXXX-v1-time-sync-decisions.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Consequences
+
+A standardized NTP setup will allow SCS users and third party image providers to develop images that support local time synchronization across SCS clouds.
+
+As discussed in the NTP section above, there are a number of limitations in OpenStack and related projects that a standardized NTP setup has to work with:
+
+* OpenStack currently has no support for providing NTP servers to VMs under a fixed link-local address, like AWS and GCP are doing.
+  Such a feature could probably be implemented by re-using the metadata service IP (`169.254.169.254`) and the mechanisms to inject it into subnets.
+  If this feature becomes available in the future, it will be an attractive target for standardization, but until then it seems more sensible to focus on servers available through provider networks.
```
artificial-intelligence, line 114: It was afaik encouraged during the discussion to actually work on this upstream.

This will not be implemented upstream by just waiting for it.

At least we should mention this in the decision record imho.

artificial-intelligence (changes requested):

kgube:
I discussed the potential upstream topic with Neutron Team, and [created an RFE issue for it](https://bugs.launchpad.net/neutron/+bug/2084894).

The topic will also be discussed during the PTG, it is [currently scheduled for the 2014-10-24 15:00 - 16:00 UTC timeslot](https://etherpad.opendev.org/p/oct2024-ptg-neutron).

Standards/scs-XXXX-v1-time-sync-decisions.md
kgube: Sorry, I forgot to reply. I have since updated the DR accordingly.

kgube:
I could not attend the PTG, unfortunately, but the Topic was discussed and there were some questions on the scope of the feature that were forwarded to the [RFE ticket](https://bugs.launchpad.net/neutron/+bug/2084894), which I answered.
In particular, both ovn and dnsmasq allow global dhcp-options, so provided that the link-local NTP server address is the same in all subnets (which would be a design goal), we can configure it as a global option and there is no dynamic port-specific DHCP-config necessary.

If we want to proceed with pursuing this feature, it would probably best to track it in a separate issue.
The next step would be to take the RFE Ticket to a [Neutron Driver meeting](https://wiki.openstack.org/wiki/Meetings/NeutronDrivers), get affirmation of the scope of the feature from the team, and ask for guidance on how to proceed with the implementation.

artificial-intelligence (commented):
not sure if jitter times and such should be mandated but I currently don't have much time reviewing and researching this topic in depth, so I don't want to hold this up.

artificial-intelligence (approved):

fkr:
> @kgube thanks for fixing the linter issues, sadly I also have no idea what the scs/check post error means. Given that this also seems to fail similarly for entirely different PRs, I would assume it is not related to yours, but rather a general pipeline issue.

Is zuul properly working? Am just on my phone right now, but it seems that zuul is not properly responding - eg the webfrontend does not even open for me.

@garloff?

Issue #583, opened 2024-05-03
[BUG] flavors.scs.community should keep CPU arch if non x86 in short name.


garloff:
The flavor name generator on https://flavors.scs.community/ does handily also suggest a short name for a generated flavor name that strips out some details that are not relevant for most customers.
It e.g. strips out, whether we have an intel or AMD CPU and what the CPU generation is.
For non-x86 CPUs, it strips out one detail too much though: The flavor name should encode that we have arm or riscv CPUs rather than x86-64. So we should keep it in the short name, as this does make a difference. Images will not work on both.
The standard says that a not-specified CPU vendor means generic x86-64, so we need to indicate it if we're not x86-64.

Note: This could be debated -- unlike most of the other properties, there is metadata for the architecture to ensure we don't try to boot an x86-64 image on a aarch64 flavor. But I prefer to select the non-x86 CPU architecture by explicitly saying so in chosing the flavor.

mbuechse:
Solved with #584

PR #584, opened 2024-05-03
Keep CPU architecture in short name (if non-x86).

garloff:
This should address #583


Tests/iaas/flavor-naming/flavor_names.py
```diff
@@ -234,6 +234,13 @@ def shorten(self):
...
         if self.hype is None and self.hwvirt is None and self.cpubrand is None:
             return self
+        # For non-x86-64, don't strip out CPU brand for short name, as it contains the architecture
+        if self.cpubrand and self.cpubrand.cpuvendor != 'i' and self.cpubrand.cpuvendor != 'z':
+            cpubrand = self.cpubrand
+            # Delete non-relevant details for the short name, only keep brand
+            cpubrand.cpugen = 0
+            cpubrand.perf = ""
+            return Flavorname(cpuram=self.cpuram, disk=self.disk, cpubrand=cpubrand, gpu=self.gpu, ib=self.ib)
         return Flavorname(cpuram=self.cpuram, disk=self.disk, gpu=self.gpu, ib=self.ib)
```
mbuechse, line 244: Don't mutate the existing cpubrand object.

```suggestion
        cpubrand = None
        if self.cpubrand and self.cpubrand.cpuvendor not in ('i', 'z'):
            # Delete non-relevant details for the short name, only keep brand
            cpubrand = CPUBrand()
            cpubrand.cpuvendor = self.cpubrand.cpuvendor
            cpubrand.cpugen = 0
            cpubrand.perf = ""
        return Flavorname(cpuram=self.cpuram, disk=self.disk, cpubrand=cpubrand, gpu=self.gpu, ib=self.ib)
```

mbuechse (commented):
I think we shouldn't mutate the existing flavor name. Therefore a new instance of `CPUBrand` should be created and used. See my suggestion.

garloff:
Hi Matthias,
I further improved upon your proposal by adding an explicit constructor to the `class CPUBrand`.
Let me know what you think.

Tests/iaas/flavor-naming/flavor_names.py
```diff
@@ -192,6 +192,11 @@ class CPUBrand:
     })
     perf = TblAttr("Performance", {"": "Std Perf", "h": "High Perf", "hh": "Very High Perf", "hhh": "Very Very High Perf"})
 
+    def __init__(self, cpuvendor="i", cpugen=0, perf=""):
```
mbuechse, line 195: It's a bit strange that this is the only class of the bunch that has a constructor, which might raise some eyebrow, and it is only used at one occurrence, but I guess these points aren't sufficient to warrant any real objection.

Tests/iaas/flavor-naming/flavor_names.py
```diff
@@ -234,6 +239,11 @@ def shorten(self):
         """return canonically shortened name as recommended in the standard"""
         if self.hype is None and self.hwvirt is None and self.cpubrand is None:
             return self
+        # For non-x86-64, don't strip out CPU brand for short name, as it contains the architecture
+        if self.cpubrand and self.cpubrand.cpuvendor != 'i' and self.cpubrand.cpuvendor != 'z':
```
mbuechse, line 243: I still prefer 

```suggestion
        if self.cpubrand and self.cpubrand.cpuvendor not in ('i', 'z'):
```

I would also have preferred the variant with the single call of the constructor, but that's mainly my taste.

Tests/iaas/flavor-naming/flavor_names.py
garloff: The fact we don't even need constructors for most is the exception I think, not that we have one here.
The c'tor does what I would expect from a c'tor, so defining it and using it makes the code where the logic happens more readable IMVHO.

garloff:
> Sorry. I meant start a review and posted single comments instead. My bad. I think this can be merged, once you took note of my comments.

No worries. I merged you `in ('a', 'i')` suggestion, that's more pythonic ...

PR #585, opened 2024-05-03
Update Domain Manager standard to keep enforce_scope and enforce_new_defaults disabled

markus-hentsch:
As long as the Domain Manager persona is not fully integrated upstream[^1][^2], we need to implement it using policy adjustments only.

This makes it incompatible with the new `enforce_scope` and `enforce_new_defaults` options of oslo.policy in Keystone.
The options are still disabled per default currently but are planned to be the new default in the future.

Since it is currently unknown when the upstream contribution work will conclude, we might see the options becoming the new default before we get the persona upstream.
So for the standard to be future-proof, we should mandate to keep the conflicting options disabled.

This will not change existing infrastructures as it matches current defaults.

[^1]: https://bugs.launchpad.net/keystone/+bug/2045974

[^2]: https://review.opendev.org/c/openstack/keystone-specs/+/903172

### DIFF
--- a/Standards/scs-0302-v1-domain-manager-role.md
+++ b/Standards/scs-0302-v1-domain-manager-role.md
@@ -98,6 +98,8 @@ This means that by creating a new role and extending Keystone's API policy confi
 
 ### Limitations
 
+#### Domain Visibility
+
 The approach described in this standard imposes the following limitations:
 
 1. as a result of the "`identity:list_domains`" rule (see below), Domain Managers are able to see all domains[^5] via "`openstack domain list`" and can inspect the metadata of other domains with "`openstack domain show`"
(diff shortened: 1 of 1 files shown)


markus-hentsch:
Note that the Role Standard (#590) will most likely end up mandating to disable those options for all services in general due to their conflict with Heat[^1] which SCS is in the process of officially supporting as an optional component[^2].

As such, the addition to this standard seems less impactful all things considered.

[^1]: https://governance.openstack.org/tc/goals/selected/consistent-and-secure-rbac.html#the-issues-we-are-facing-with-scope-concept

[^2]: https://github.com/SovereignCloudStack/standards/pull/587

markus-hentsch:
> Note that the Role Standard (https://github.com/SovereignCloudStack/standards/pull/590) will most likely end up mandating to disable those options for all services in general due to their conflict with Heat which SCS is in the process of officially supporting as an optional component.

This is not true anymore. Things have changed:

- `enforce_scope` and `enforce_new_defaults` do not clash with the SCS Domain Manager implementation anymore as of Keystone 2024.2 release
- incompatibities of Heat with those options have been fixed

When the Domain Manager standard moves from Draft to Stable, either 2024.2 will be available or the Domain Manager persona is even already implemented upstream. This PR is obsolete now, closing.

PR #589, opened 2024-05-09
Multiline entry for PCO Regions.

garloff:


garloff:
Hmm, this breaks markdownlint.
Should we disable MD033?
But maybe there are good reasons to forbid inline HTML?
(Not that I think inline HTML is a good idea in general, but occasionally, it might be the only solution.)

garloff:
Sidenote about the other CI fail: The TLS cert for the regio cloud HM is expired. This is addressed in PR #588.

maxwolfs:
RE: MD033: Let's do it the way we handlet it within the docs page: https://github.com/SovereignCloudStack/docs/pull/193/files by disabling MD033 only for this file.

berendt:
> Sidenote about the other CI fail: The TLS cert for the regio cloud HM is expired. This is addressed in PR #588.

You can also keep it in. I finally have it on the list to repair by the beginning of next week (just In time for the summit..).

garloff:
> Looks like this is already merged now. I will fix the HM on our site over the weekend and will re-add the link with a separate PR.

We unfortunately have this list twice -- here in the README.md and in docs/standards/certification/overview.md
The latter was merged already. We should consolidate this in the future ... https://github.com/SovereignCloudStack/docs/issues/166

frosty-geek (approved):
LGTM

PR #378, opened 2023-11-10
Draft: Roles Standard

markus-hentsch:
Adds a standard to define all RBAC roles and permissions in the SCS IaaS layer.

Resolves https://github.com/SovereignCloudStack/issues/issues/396

PR #587, opened 2024-05-07
Create standard for mandatory and supported IaaS services

josephineSei:
closes https://github.com/SovereignCloudStack/issues/issues/528

PR #590, opened 2024-05-14
Draft: Roles Standard

### LINKED ISSUE #396: Define list of useful SCS-standardized roles in OpenStack
As a IaaS user of SCS-compatible clouds, I want a number of standard roles to be available on every SCS-compatible IaaS environment which serve my typical needs and which are the same (especially from a security & privacy analysis point of view) on all these clouds.

These could be an "admin" role (not available to users, just operators), "domain-manager", "project-member", "read-only", "auditor". These would be global (all services) inside a project (or domain for domain-manager). Maybe read-only vs autitor distrinction is not useful ...
Work has been done upstream (RBAC work in Yoga) on this.

Tasks:
* Research what exists
* Discuss what is needed (keep it simple, stay close to upstream if possible)
* Implement it

## Definition of Ready:

- [ ] User Story is small enough to be finished within one sprint
- [ ] User Story is clear and understood by the whole team
- [ ] Acceptance criteria are defined
- [ ] Acceptance criteria are clear and understood by the whole team

## Definition of Done:

- [ ] All acceptance criteria are met
- [ ] Changes have been reviewed
- [ ] CI tests have run successfully
- [ ] Documentation has been updated
- [ ] Release Notes have been updated

markus-hentsch:
Adds a standard to define all RBAC roles and permissions in the SCS IaaS layer.

Resolves https://github.com/SovereignCloudStack/issues/issues/396

Supersedes #378


berendt:
Does it makes sense not to only link the domain-manager role standard but also to list the domain-manager role inside this standard?

markus-hentsch:
> Does it makes sense not to only link the domain-manager role standard but also to list the domain-manager role inside this standard?

It is listed as "manager" under the core roles. It isn't called domain-manager anymore, see https://docs.scs.community/standards/scs-0302-v1-domain-manager-role/#change-the-naming-of-the-domain-manager-role-to-align-with-upstream

Standards/scs-03XX-v1-standard-roles.md
```diff
@@ -0,0 +1,250 @@
...
+This in turn renders encryption features like the volume encryption of OpenStack's volume service unusable for customers unless the corresponding users are assigned the Barbican-specific "creator" role in projects additionally.
+This creates unnecessary management overhead on the CSP side and ambiguity for users since the role is only useful in Barbican but its name does not communicate this.
+
+To improve user experience and make the encryption features easily accessible, this standard should adjust the Key Manager API policies to extend permissions referencing the Barbican-specific "creator" role by the "member" role.
+This offers users easy access to the Key Manager API and aligns the permission set with the future rework (as per `enforce_new_defaults`), because it will later replace the "creator" role by the "member" role entirely.
+
+The "creator" role will be kept for compatibility reasons concerning service integration.
+For example, the block storage service Cinder usually has a technical user in Keystone possessing the "creator" role in the "service" project.
```
markus-hentsch, line 111: FTR, I considered replacing the "creator" role by "member" entirely. However, I decided against it for the following reasons:

- both DevStack and RedHat setups[^1] seem to assign the "creator" role to a "cinder" user in the "service" project, which is registered in Cinder's `[keystone_authtoken]` config section
- mandating to switch Cinder's service user to "member" role here could open up permissions in other services' APIs potentially which otherwise wouldn't react to the "creator" role (which is Barbican-specific)
  - without `enforce_scope` (which this standard *cannot* use), the scoping would not protect us here

*However*, I'm not exactly sure in which workflow Cinder needs its own Barbican-enabled role, i.e. what the reasoning for this is. Usually, the user's auth token is passed around during requests so it would simply take the role permissions from there. However, it is most likely there for a reason, so I'm leaving it in because I don't know the context and don't want to break things.

[^1]: https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/13/html/manage_secrets_with_openstack_key_manager/encrypting_cinder_volumes#overview_of_the_migration_steps

markus-hentsch:
> > Does it makes sense not to only link the domain-manager role standard but also to list the domain-manager role inside this standard?
> 
> It is listed as "manager" under the core roles. It isn't called domain-manager anymore, see https://docs.scs.community/standards/scs-0302-v1-domain-manager-role/#change-the-naming-of-the-domain-manager-role-to-align-with-upstream

FTR, I have now moved the role from core rolees to service-specific roles after some more research. For the rationale see the new 'Classification of the "manager" role' section under the Design Considerations.

Standards/scs-03XX-v1-standard-roles.md
josephineSei: I guess Cinder will need it in case of a deletion of an encrypted volume (to delete the key) or to create a backup or snapshot (to copy the key). Maybe there is an edge case, where e.g. a backup is not triggered by someone from the project (hence no chance to aquire the key) or a deletion of a key is done so late, that the user conetxt is already omitted, so cinder does not have the token of the user anymore.

Standards/scs-03XX-v1-standard-roles.md
```diff
@@ -0,0 +1,271 @@
...
+
+This primarily concerns the new scoping and defaults in `oslo.policy`:
+
+```ini
+[oslo_policy]
+enforce_new_defaults = True
+enforce_scope = True
+```
+
+As of the time of writing this standard, those options currently default to `False`[^3] for all OpenStack services.
```
josephineSei, line 146: I thought they default to True for at least Nova, Glance and Neutron: https://etherpad.opendev.org/p/rbac-goal-tracking ?

Even though I wonder how these services work together with services with the old settings...

Standards/scs-03XX-v1-standard-roles.md
```diff
@@ -0,0 +1,271 @@
...
+| Octavia | load-balancer_global_observer | CSP | access to read-only APIs including resources owned by others |
+| Octavia | load-balancer_member | customer | access to read and write APIs |
+| Octavia | load-balancer_quota_admin | CSP | admin access to quota APIs only, including quota resources owned by others |
+| Octavia | load-balancer_admin | CSP | admin access to all LB APIs including resources owned by others |
+| Swift | ResellerAdmin | Ceilometer (internal) | assigned to technical users of Ceilometer to integrate with Swift for access privileges in the object storage API to store statistics for metering |
+| Heat | heat_stack_user | internal | assigned to technical user accounts resulting from other resources' creation in Heat templates |
+
+### API Policies
+
+TODO: what does the CSP need to adhere to when it comes to API policy configuration?
```
josephineSei, line 223: Maybe we can advise that the policies at least should not be loosened. I don't think we can advise any more than maybe a few edge cases.

Standards/scs-03XX-v1-standard-roles.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Scope Enforcement Compatibility
+
+The API policy library used by OpenStack (oslo.policy) introduced two new [configuration options](https://docs.openstack.org/oslo.policy/latest/configuration/#oslo-policy) during the ongoing RBAC rework of OpenStack[^2]:
+
+- `enforce_scope`
+- `enforce_new_defaults`
+
+Using those new defaults and scope-enforcing options [will currently break](https://governance.openstack.org/tc/goals/selected/consistent-and-secure-rbac.html#the-issues-we-are-facing-with-scope-concept) orchestration tooling such as **OpenStack Heat** and Tacker.
+Due to OpenStack Heat being a service supported by the SCS project, those conflicting options cannot be mandated by a SCS standard.
```
josephineSei, line 49: After reading the whole standard, I can say this is a bad time to try to write a standard for roles for OpenStack APIs due to the slow an sometimes controversial work on secure RBAC upstream. Nevertheless we have to do it and maybe we should reconsider the way we go here:
1. This is based on the list of mandatory and supported services.
2. Only one of the supported services is the reason, that we cannot use the new and ongoing work of secure RBAC <- Is this correct?
3. We have to made adjustments to the old somehow deprecated way (in Barbican and Octavia) which will at some point hopefully in near future outdated, because of 2.
4. Upstream will not go back to the old way, but will only focus on the new secure RBAC


If we reconsider supporting Heat, do we have any other things that would require us to use the old version?
If there would be nothing holding us from using the new defaults, would those just work out of the box (with setting the two config parameters to true)?

I hesistant to to put so much work into something, that is not supported anymore by upstream OpenStack, but is already under re-development. If there would be a more easy way to directly make use of the new default roles, would that also safe CSPs time and headache when adjusting policies?

josephineSei (commented):
Overall this is a very unfortunate time to write a role standard.

So maybe we should discuss this in a bigger group, if we somehow can make it easier for you. (E.g. kick heat from the supported list?)

Standards/scs-03XX-v1-standard-roles.md
markus-hentsch: Thank you for bringing this up! I was not aware of this and only looked up the defaults in the oslo.policy library[^1]. Furthermore, on the Secure RBAC page they stated "we decided to postpone the scope implementation"[^2] so I didn't expect something like this.

Turns out, Nova **overrides the oslo.policy defaults with its own defaults for the library**[^3]. Likely, Glance and Neutron do the same.

I need to check that and maybe need to rewrite parts of the standard once again ...

[^1]: https://github.com/openstack/oslo.policy/blob/c7083ec2c591e0f921d3532135e235695df0778f/oslo_policy/opts.py#L38

[^2]: https://governance.openstack.org/tc/goals/selected/consistent-and-secure-rbac.html#the-issues-we-are-facing-with-scope-concept

[^3]: https://github.com/openstack/nova/blob/3a53d715cd505258c1680c442b64c5bf7e182352/nova/policy.py#L44-L52

Standards/scs-03XX-v1-standard-roles.md
```diff
@@ -0,0 +1,271 @@
...
+| CSP | Cloud Service Provider, provider managing the OpenStack infrastructure |
+| IaaS | Infrastructure-as-a-Service |
+| IAM | Identity and Access Management |
+| RBAC | Role-Based Access Control[^1] established by OpenStack Keystone |
+
+[^1]: [OpenStack Documentation: Role-Based Access Control Overview](https://static.opendev.org/docs/patrole/latest/rbac-overview.html)
+
+## Motivation
+
+The permission settings of OpenStack RBAC roles are configured in service-specific deployment configuration files (usually the respective "`policy.yaml`") in a rather static way and have to be carefully managed.
```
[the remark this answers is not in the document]
markus-hentsch: Thanks for pointing that out. You are right, the phrasing did suggest that the configuration files are the primary source but it's actually the in-code defaults.
I have rephrased this paragraph to better reflect that.

Standards/scs-03XX-v1-standard-roles.md
```diff
@@ -0,0 +1,271 @@
+---
+title: SCS Standard Roles
+type: Standard
+status: Draft
+track: IAM
+---
+
+## Introduction
+
+SCS aims to provide a standardized role model for RBAC roles across all supported OpenStack API services that applies sensible and consistent permission sets based on a set list of roles defined by a standard.
```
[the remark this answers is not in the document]
markus-hentsch: I rephrased this part.

Standards/scs-03XX-v1-standard-roles.md
```diff
@@ -0,0 +1,203 @@
...
+```ini
+[oslo_policy]
+enforce_new_defaults = True
+enforce_scope = True
+policy_file = policy.yaml
+```
+
+#### API Policies
+
+TODO: what does the CSP need to adhere to when it comes to API policy configuration?
```
josephineSei, line 173: Maybe this is more kind of an implementation note?
Because policy adjustments might occur more often than a change of the role model itself.

And the policy I am thinking about is the network rbac - which might be changed in the external network standard to only allow admin role to use it: https://github.com/SovereignCloudStack/standards/pull/572

josephineSei (commented):
I think this is going into a good direction.

Standards/scs-03XX-v1-standard-roles.md
markus-hentsch: This is the last part I need to figure out for the standard: what to put here.

We will definitely need to make sure that CSPs apply the special policy templates offered by Octavia[^1] so that the LBaaS API will be aligned with the reader, member, admin model.

Other than that, we're pretty much in line with the in-code defaults now that we can use Secure RBAC. I need to double check but I think there won't be much more needed to be added here in terms of strict guidelines.

[^1]: https://docs.openstack.org/octavia/2024.1/configuration/policy.html#openstack-default-roles-policy-override-file

Standards/scs-03XX-v1-standard-roles.md
markus-hentsch: Standard is now updated to reflect the actual state.

Standards/scs-03XX-v1-standard-roles.md
markus-hentsch: Right. I added remarks about not permitting any deviation of the default role model in a detrimental manner and addressed the Octavia edge case.

Standards/scs-03XX-v1-standard-roles.md
markus-hentsch: > 2. Only one of the supported services is the reason, that we cannot use the new and ongoing work of secure RBAC <- Is this correct?

Not anymore, see [this](https://github.com/SovereignCloudStack/issues/issues/396#issuecomment-2200097660) and [this](https://github.com/SovereignCloudStack/issues/issues/396#issuecomment-2253052813).

> 3. We have to made adjustments to the old somehow deprecated way (in Barbican and Octavia) which will at some point hopefully in near future outdated, because of 2.

Only Octavia needs special care in this regard.
Barbican already switches to the new role model as soon as `enforce_new_defaults` is enabled[^1].

> 4. Upstream will not go back to the old way, but will only focus on the new secure RBAC

And we should do the same with standard I think.
The standard should not bother supporting some deprecated leftovers. I'd rather postpone the standard a bit until we get an OpenStack release (maybe 2024.2?) where the new SRBAC just works and we can keep the standard really simple and mandate to use the new defaults.

[^1]: https://github.com/openstack/barbican/blob/d99764e6112cedc6ca168a29fb0279dec2d94288/barbican/common/policies/secrets.py#L85-L88

Standards/scs-03XX-v1-standard-roles.md
markus-hentsch: I added Octavia there now plus some general do's and dont's. It's pretty brief and I don't think it justifies an implementation note split currently.

> And the policy I am thinking about is the network rbac - which might be changed in the external network standard to only allow admin role to use it: https://github.com/SovereignCloudStack/standards/pull/572

@josephineSei I'm wondering whether this should be part of the corresponding network standard or this role standard. I think it's more a detail of how the networking stuff works and is to be used, not a general role model or scoping thing. So I'm leaning more towards addressing this in the network standards if there is a network standard that would fit.

Standards/scs-03XX-v1-standard-roles.md
```diff
@@ -0,0 +1,214 @@
...
+  As of the 2024.1 release of Octavia, this template is provided as [keystone_default_roles_scoped-policy.yaml](https://github.com/openstack/octavia/blob/stable/2024.1/etc/policy/keystone_default_roles_scoped-policy.yaml).
+
+## Related Documents
+
+### SCS Mandatory and Supported IaaS Services
+
+**Description:** SCS standard that lists mandatory and supported OpenStack APIs for SCS clouds.
+This list is decisive for the standard on roles as all applicable services need to be taken into consideration.
+
+**Link:** TBD <!-- https://github.com/SovereignCloudStack/standards/pull/587 -->
```
markus-hentsch, line 193: Needs to be replaced by a genuine link once #587 is merged.

markus-hentsch:
FTR, I removed the part about the role adjustment for Barbican and linked the corresponding implementation notes of the Key Manager standard in the related documents section instead, since this is part of that standard now.

markus-hentsch:
FTR, I think this PR will need to stay in draft state as long as upstream has remaining incompatibilities with `enforce_new_defaults` in any of the SCS-mandatory/supported services (esp. Cinder and Barbican), see https://github.com/SovereignCloudStack/issues/issues/396#issuecomment-2361401271

This standard must be able to rely fully on the newer SRBAC model (reader, member, admin) without any service-specific exceptions, in my opinion.

jklare:
@markus-hentsch thank you for the clarification of the next steps. Given that there seems to be limited interest of the community presently to work on this, I will close this PR for now. Once we find somebody who wants to continue the work here, I think your pointers will be helpful and we can reopen the PR.

Issue #592, opened 2024-05-16
Introduce mechanism to parametrize standards.


mbuechse:
Sometimes a standard can be neatly divided into a static part on the one hand and dynamic parts on the other, such as:

- [scs-0104-v1](https://docs.scs.community/standards/scs-0104-v1-standard-images), where the actual list of standard images is a yaml file that is not part of the standard text,
- the [upcoming standard](https://github.com/SovereignCloudStack/standards/pull/548) on connection security might refer to [Mozilla recommendations](https://wiki.mozilla.org/Security/Server_Side_TLS#JSON_version_of_the_recommendations), which get published as versioned json files.

The check scripts that are specified in the [certificate-scope yaml](https://docs.scs.community/standards/scs-0003-v1-sovereign-cloud-standards-yaml) may of course accept command-line parameters of all sorts, but those are not part of the [official docs](https://docs.scs.community/standards/scs-compatible-iaas). If we introduce the concept of parameters for standards, their values could easily be documented.

PR #595, opened 2024-05-17
Introduce mechanism to parametrize standards

mbuechse:
solves #592


fkr (approved):
Very good idea. LGTM!

Issue #509, opened 2024-03-12
Standardize List of allowed Barbican Plugins


josephineSei:
Barbican is used to store secrets for encrypted resources in OpenStack. To store those secrets different plugins / backends can be used. For security issues, we should not allow all possible plugins.
E.g. there is the simply crypto plugin, that simply stores the Master KEK in the config file, and all other secrets in the database, which makes it easy to access all secrets.

Most likely testing this will be only possible through audits, as this is something configuration specific and/or deployment specific.

## Definition of Done:

- [ ] Proposal has been written with name of the form `scs-xxxx-v1-slug.md` (only substitute `slug`)
- [ ] Proposal has the fields `status`, `type`, `track` set
- [ ] Proposal has been voted upon in the corresponding team
- [ ] Status has been changed into `Draft`, file renamed: `xxxx` replaced by document number
- [ ] If applicable: test script has been written (this item may be moved into a separate issue so long as the state is `Draft`)

josephineSei:
At first I have to note, that it is better to use encryption with a not-optimal backend than to not use encryption at all.

There is the possibility to configure multiple plugins: https://docs.openstack.org/barbican/latest/configuration/plugin_backends.html
-> this adds a lot of complexity to this issue, as we have to do a deep dive into the correct choosing of backend.

The Plugins are described here: https://docs.openstack.org/barbican/latest/install/barbican-backend.html
Plus: There is the possibility to use a software HSM instead of a hardware one.

| Plugin | Master KEK store | alternative | possible security risk |
| ---- | --- | --- | --- |
| simple crypto | config file | config and Barbican processes in enclave[^1] | vulnerable controller Host |
| PKCS#11 | HSM | - | HSM access |
| KMIP | KMIP device | - | access info in config file |
| Dogtag | HSM | software-based NSS security database | access info in config file |
| Vault | Vault | - |  access info in config file |

We should evaluate the usage of the different Plugins through the CSPs.

The problem is also the testing of this issue: We would need to audit the config file used for Barbican in the deployment to know exactly which plugin is used. But doing so would compromise the security of Barbican, as information from this file should not be available to anyone outside the few operators of the CSP.

My conclusion is, that in regards of the lack of possible testing, we could only give some recommendation to CSPs. And we should only do so, if we find, that we want to exclude one of the plugins OR if we want to rather give advice on how to rise the protection level of the config files / controller nodes / wherever Barbican is executed.

[...]

PR #599, opened 2024-05-23
Key-Manager Standard

josephineSei:
closes https://github.com/SovereignCloudStack/standards/issues/509


josephineSei:
If we decide to move Barbican from Supported to Mandatory in the lists of mandatory and supported openstack services, we may not need an extra test script here, but should move it into the other test-file.

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -0,0 +1,73 @@
...
+title: Key-Manager Standard
+type: Standard
+status: Draft
+track: IaaS
+---
+
+## Introduction
+
+To encrypt user data like volumes or in the future also Images and ephemeral storage for VMs, the key is needed to be present in the infrastructure.
+Therefore a key-manager is needed to store the keys and apply autorization policies on requests.
```
artificial-intelligence, line 11: typo:
```suggestion
Therefore a key-manager is needed to store the keys and apply authorization policies on requests.
```

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -0,0 +1,73 @@
...
+| API | Application Programming Interface, often referring to the REST API interfaces provided by OpenStack and related services |
+| CSP | Cloud Service Provider, provider managing the OpenStack infrastructure |
+| IaaS | Infrastructure-as-a-Service |
+| HSM | Hardware Security Module |
+| KEK | Key Encryption Key |
+
+## Motivation
+
+When user data is encrypted the keys need to be stored securely in the Infrastructure to be present, when a possible encryption or decryption needs to take place.
+Key-managers like Barbican provide this functionality on IaaS-Level.
```
artificial-intelligence, line 28: nit:
```suggestion
Key-managers like Barbican provide this functionality on the IaaS-Level.
```

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -0,0 +1,73 @@
...
+
+While discussing what this standard should aim for it was discovered that some CSPs don't use Barbican or another key-manager at all and do not provide the feature to encrypt user data to their customers.
+This should change, but the exact change comes with financial burden, when choosing a plugin in Barbican to store the Master-KEK.
+To minimize the burden and enable more CSPs to step up and provide encryption, this standard will only make recommendations about plugins.
+
+### Options considered
+
+#### _Option 1_
+
+Barbican as the OpenStack key-manager implementation offers different plugins to store the Master-KEK.
```
artificial-intelligence, line 47: I feel this sentence is a little bit redundant with the section in lines 32-35, maybe it can be removed?

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -0,0 +1,73 @@
...
+
+Looking into the plugins and possible attack vectors one design decision in the plugins is very important: where and how to store the Master-KEK.
+Because the Plugins might use different technologies, but most of them increase the security level by not storing the Master-KEK in plain text on the physical machine Barbican is running on.
+This mechanism as a whole, is something that CSPs should aim to do.
+
+## Standard
+
+To increase the level of security and overall user data encryption CSPs SHOULD implement the key-manager API (e.g. implemented by barbican) with a security level of storing Keys encrypted and storing the KEK in another place than the Keys.
+
+If possible CSPs SHOULD NOT store the Master-KEK in plain-text on the physical host the key-manager is running on.
```
artificial-intelligence, line 62: I think it would be helpful here to explicitly call out the Barbican plugins which do store keys unencrypted on local disc, so CSPs can avoid these, or do the reverse and at least list some plugins which are deemed "secure" (enough).

artificial-intelligence (changes requested):
This standard touches on a topic which comes up time and time again: Should we standardize something very implementation specific?

While the standard doesn't mandate anything in it's current form I could easily see this change in the future, because of the focus on Barbican here.

I'm conflicted if this is a problem or rather a feature though.

Still, even if currently no CSPs implement a Key-Manager I would like to hear opinions of other CSPs if they will implement this, if the standard doesn't really enforce anything.

Personally I don't see much value in a standard that doesn't enforce anything. For a draft and further discussion it is of course fine.
We need to start somewhere, after all.

Standards/scs-XXXX-vN-key-manager-standard.md
josephineSei: I removed it and rephrased the whole paragrph a little bit

Standards/scs-XXXX-vN-key-manager-standard.md
josephineSei: The problem here is:
the simple crypto plugin does store the masterKEK unencrypted in the Barbican config. So we can name it as a negative example. But I do not think we should deem this plugin as completely insecure.
E.g. it can be managed to let Barbican run in enclaves like SGX, to protect the config.
And having Barbican with simpley crypto is still better than not having any option to encrypt user data at all - so I am a bit hesitant to give a "not do this"-example.

We could add a sentence stating, that using HSMs would be one option to fulfill this, but this could give CSPs the impression, that we will put a high financial burden on them.

I am skeptical of both options and would like to have more input from other reviewers on this

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -0,0 +1,72 @@
...
+| KEK | Key Encryption Key |
+
+## Motivation
+
+When user data is encrypted the keys need to be stored securely in the Infrastructure to be present, when a possible encryption or decryption needs to take place.
+Key-managers like Barbican provide this functionality on the IaaS-Level.
+Using such a key-manger and thus provide to customers the possibility to encrypt user data is not yet implemented everywhere in IaaS deployments.
+A first step towards more security is to encourage CSPs to use this service and provide the feature of using encrypted resources to customers.
+It is also important to take a closer look into the key-manager and to apply aim for an appropiate level of security there.
+The Key-Manager is responsible for storing the keys securely, which can be done by encrypting them with a KEK and storing them in a database and store the KEK either also encrypted in the database or to store it somewhere else.
```
gtema, line 32: The Key-Manager service manages keys in a secure manner. This can be achieved differently and is not primarily in scope of this standard.

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Terminology
+
+| Term | Meaning |
+|---|---|
+| API | Application Programming Interface, often referring to the REST API interfaces provided by OpenStack and related services |
+| CSP | Cloud Service Provider, provider managing the OpenStack infrastructure |
+| IaaS | Infrastructure-as-a-Service |
+| HSM | Hardware Security Module |
+| KEK | Key Encryption Key |
```
gtema, line 23: I suggest to add Barbican into the Terminology to reduce repeating in the spec itself.

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Motivation
+
+When user data is encrypted the keys need to be stored securely in the Infrastructure to be present, when a possible encryption or decryption needs to take place.
+Key-managers like Barbican provide this functionality on the IaaS-Level.
+Using such a key-manger and thus provide to customers the possibility to encrypt user data is not yet implemented everywhere in IaaS deployments.
+A first step towards more security is to encourage CSPs to use this service and provide the feature of using encrypted resources to customers.
+It is also important to take a closer look into the key-manager and to apply aim for an appropiate level of security there.
+The Key-Manager is responsible for storing the keys securely, which can be done by encrypting them with a KEK and storing them in a database and store the KEK either also encrypted in the database or to store it somewhere else.
+Barbican as the OpenStack implementation of a key-manager is relying on a Master-KEK, which encrypts project-specific KEKs, which encrypt the Keys within a project.
```
gtema, line 33: Barbican stores keys encrypted with the project specific KEK, including the KEK itself, in the database. Master KEK, used to encrypt project specific KEK is not stored in the database and is stored differently depending on the backend storage plugin used.

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -0,0 +1,72 @@
...
+
+#### _Option 1_
+
+It was considered to only recommend a certain set of plugins or backends for the key-manager, but this may be very prone to change if Barbican adds a new plugin.
+As the SCS only wants to mandate the API that can be abstracted through the Castellan library in OpenStack, integrating any other key-manager implementation is not uncommon, so this standard needs to consider other possible key-managers as well.
+Due to these reasons this Option was disregarded.
+
+#### _Option 2_
+
+Looking into the plugins and possible attack vectors one design decision in the plugins is very important: where and how to store the Master-KEK.
```
gtema, line 53: Looking at the available Barbican plugins ...

Standards/scs-XXXX-vN-key-manager-standard.md
gtema: I would add another sentence like: "Barbican supports deploying out-of-tree drivers what enables operators to satisfy their specific needs." With this sentence I would like to give a signal to the CSPs that they have quite lot possibilities in their hands to offer better security to their customers.

Tests/iaas/key-manager/check-for-key-manager.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+
+    for svc in services:
+        svc_type = svc['type']
+        if svc_type == "key-manager":
+            # key-manager is present
+            return 0
+
+    # we did not find the key-manager service
+    logger.warning("There is no key-manager endpoin in the cloud.")
```
gtema, line 49: endpoint

gtema (changes requested):

Standards/scs-XXXX-wN-key-manager-implementation-testing.md
```diff
@@ -0,0 +1,27 @@
...
+There are also software HSMs available, that should be tested for their integration into the Barbican workflow.
+
+## Automated Tests
+
+The check for the presence of a Key Manager is done with an test script, that checks the presence of a key manager service in the catalog endpoint of openstack.
+This check can eventually be moved to the checks for the mandatory an supported service/API list, in case of a promotion of the key-manager to the mandatory list.
+
+## Manual Tests
+
+There need to be a manual test that searches the physical host of the Key-Manager host for the present of a Master-KEK in Plain-Text, e.g. in the Barbican config file.
```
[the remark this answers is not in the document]
josephineSei: It is possible, to run certain parts of Barbicna in an enclave that also holds the complete config file. In that case, the Master KEK will not be found.

So one option as a manual test would be to search for a plain text config file and check, whether simple_crypto is used as a plugin and the MasterKEK is written into the config.

Maybe in future there are other ways to secure the MasterKEK even within the simple_crypto plugin. We should not make assumptions about that part of configuration of this plugin here, other than that we want to encourage CSPs to protect the MasterKEK.

Standards/scs-XXXX-vN-key-manager-standard.md
josephineSei: @gtema this is more of an implementation note and not part of the standard. Should I still add this somewhere in this document or just reference the implementation notes somewhere?

Maybe an:
```
:::tip

Barbican supports deploying out-of-tree drivers what enables operators to satisfy their specific needs.

:::
```
should be added?

gtema: yupp, adding a tip sound good (I agree that we can place it in the impl. part). I just want to ensure, that CSPs are not frightened by the standard and know that it is possible to have some custom stuff running what would satisfy the standard.

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -7,32 +7,34 @@ track: IaaS
 
 ## Introduction
 
-To encrypt user data like volumes or in the future also Images and ephemeral storage for VMs, the key is needed to be present in the infrastructure.
-Therefore a key-manager is needed to store the keys and apply autorization policies on requests.
-The OpenStack key-manager implemenation that can bu utilized for this is named Barbican.
+To encrypt user data like volumes or in the future also Images and ephemeral storage for VMs, the key is need to be known in the infrastructure.
+To provide the key to those operations without includign the user every time a key-manager within the infrastructure can be utilized to store the keys and apply authorization policies on requests.
```
artificial-intelligence, line 11:
```suggestion
To provide the key to those operations without including the user every time a key-manager within the infrastructure can be utilized to store the keys and apply authorization policies on requests.
```

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -7,32 +7,34 @@ track: IaaS
...
 ## Motivation
 
-When user data is encrypted the keys need to be stored securely in the Infrastructure to be present, when a possible encryption or decryption needs to take place.
-Key-managers like Barbican provide this functionality on IaaS-Level.
-Using such a key-manger and thus provide to customers the possibility to encrypt user data is not yet implemented everywhere in IaaS deployments.
-A first step towards more security is to encourage CSPs to use this service and provide the feature of using encrypted resources to customers.
+User data encryption requires an encryption key to be known during encryption and decryption processes.
+Key-managers like Barbican provide this functionality on the IaaS-Level.
+Not every IaaS deployment currently offers user data encryption as part of their standard offering.
+A first step towards more security is to encourage CSPs to provide a better data security by offering data encryption to the customers.
```
artificial-intelligence, line 31: nit:
```suggestion
A first step towards more security is to encourage CSPs to provide better data security by offering data encryption to the customers.
```

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -7,32 +7,34 @@ track: IaaS
...
 It is also important to take a closer look into the key-manager and to apply aim for an appropiate level of security there.
-The Key-Manager is responsible for storing the keys securely, which can be done by encrypting them with a KEK and storing them in a database and store the KEK either also encrypted in the database or to store it somewhere else.
-Barbican as the OpenStack implementation of a key-manager is relying on a Master-KEK, which encrypts project-specific KEKs, which encrypt the Keys within a project.
-All keys except for the Master-KEK are stored in a database.
-This Master-KEK is stored differently for each plugin and needs to be protected.
+The Key-Manager service manages keys in a secure manner.
+This can be achieved differently and is not primarily in scope of this standard.
+Barbican stores keys encrypted with the project specific KEK, including the KEK itself, in the database.
+The Master KEK, used to encrypt the project specific KEKs is not stored in the database and is stored differently depending on the backend storage plugin used.
+This standard also abstracts from used plugins and want to ensure that the Master-KEK is protected, too.
```
artificial-intelligence, line 37: nit:
```suggestion
This standard also abstracts the used plugins and wants to ensure that the Master-KEK is protected, too.
```

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -44,14 +46,13 @@ To minimize the burden and enable more CSPs to step up and provide encryption, t
 
 #### _Option 1_
 
-Barbican as the OpenStack key-manager implementation offers different plugins to store the Master-KEK.
-It was considered to only recommend a certain set of plugins, but this may be very prone to change if Barbican adds a new plugin and this standard would need to consider other possible key-managers as well.
-As the SCS only wants to mandate the API that can be abstracted through the Castellan library in OpenStack, integrating any other key-manager implementation is not uncommon.
-So this Option was disregarded.
+It was considered to only recommend a certain set of plugins or backends for the key-manager, but this may be very prone to change if Barbican adds a new plugin.
+As the SCS only wants to mandate the API that can be abstracted through the Castellan library in OpenStack, integrating any other key-manager implementation is not uncommon, so this standard needs to consider other possible key-managers as well.
+Due to these reasons this Option was disregarded.
```
artificial-intelligence, line 51: nit:
```suggestion
Due to these reasons this option was disregarded.
```

Standards/scs-XXXX-wN-key-manager-implementation-testing.md
```diff
@@ -0,0 +1,32 @@
...
+There are also software HSMs available, that should be tested for their integration into the Barbican workflow.
+
+Other Plugins in Barbican are the KMIP plugin and Vault[^1].
+They are storing the keys differently and CSPs need to make sure, that the access to the keys is configured securely.
+
+[^1]:[Barbican Plugins](https://docs.openstack.org/barbican/latest/install/barbican-backend.html)
+
+## Automated Tests
+
+The check for the presence of a Key Manager is done with an test script, that checks the presence of a key manager service in the catalog endpoint of openstack.
```
artificial-intelligence, line 27: nit:
```suggestion
The check for the presence of a Key Manager is done with a test script, that checks the presence of a Key-Manager service in the catalog endpoint of Openstack.
```

Standards/scs-XXXX-wN-key-manager-implementation-testing.md
```diff
@@ -0,0 +1,32 @@
...
+This check can eventually be moved to the checks for the mandatory an supported service/API list, in case of a promotion of the key-manager to the mandatory list.
+
+## Manual Tests
+
+There need to be a manual test that searches the physical host of the Key-Manager host for the present of a Master-KEK in Plain-Text, e.g. in the Barbican config file.
```
artificial-intelligence, line 32: nit:
```suggestion
There needs to be a manual test that searches the physical host of the Key-Manager host for the presence of a Master-KEK in Plain-Text, e.g. in the Barbican config file.
```

artificial-intelligence (changes requested):
Just some spelling stuff.

Also there are different spellings of "key-manager", "Key Manager" and "Key-Manager" throughout the text. I don't know if this is a category name in this context, but I guess it should be spelled consistently for easier grep-ability :wink: 

Contentwise this looks mostly good to me, I still would like to have more opinions of different CSPs in here, but I guess I can't force people to comment.

artificial-intelligence (approved):
LGTM

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -0,0 +1,74 @@
+---
+title: Key-Manager Standard
+type: Standard
+status: Draft
+track: IaaS
+---
+
+## Introduction
+
+To encrypt user data like volumes or in the future also Images and ephemeral storage for VMs, the key is need to be known in the infrastructure.
```
gtema, line 10: "the key is need" is syntactically wrong

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -0,0 +1,74 @@
...
+To provide the key to those operations without including the user every time a Key Manager within the infrastructure can be utilized to store the keys and apply authorization policies on requests.
```
gtema, line 11: I needed to re-read this sentence 3 times before I understood the semantics. Please simplify

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -0,0 +1,74 @@
...
+| IaaS | Infrastructure-as-a-Service |
+| HSM | Hardware Security Module |
+| KEK | Key Encryption Key |
+
+## Motivation
+
+User data encryption requires an encryption key to be known during encryption and decryption processes.
+Key Managers like Barbican provide this functionality on the IaaS-Level.
+Not every IaaS deployment currently offers user data encryption as part of their standard offering.
+A first step towards more security is to encourage CSPs to provide better data security by offering data encryption to the customers.
```
gtema, line 31: "[snip] step towards more security is to encourage [snip] to provide better data security [snip]". The sentence is really awkward.

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -0,0 +1,74 @@
...
+It is also important to take a closer look into the Key Manager and to apply aim for an appropiate level of security there.
```
gtema, line 32: "apply aim" is not a proper word combination

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -0,0 +1,74 @@
...
+#### _Option 1_
+
+It was considered to only recommend a certain set of plugins or backends for the Key Manager, but this may be very prone to change if e.g. Barbican adds a new plugin.
+As the SCS only wants to mandate the API that can be abstracted through the Castellan library in OpenStack, integrating any other Key Manager implementation is not uncommon, so this standard needs to consider other possible Key Managers as well.
+Due to these reasons this option was disregarded.
+
+#### _Option 2_
+
+Looking into the available Barbican plugins and possible attack vectors one design decision in the plugins is very important: where and how to store the Master-KEK.
+Because the Plugins might use different technologies, but most of them increase the security level by not storing the Master-KEK in plain text on the physical machine Barbican is running on.
```
gtema, line 56: "Because [A], but [B]." - I can't understand the sentence.

Standards/scs-XXXX-vN-key-manager-standard.md
```diff
@@ -0,0 +1,74 @@
...
+This mechanism as a whole, is something that CSPs should aim to do.
+
+## Standard
+
+To increase the level of security and overall user data encryption CSPs SHOULD implement the Key Manager API (e.g. implemented by Barbican) with a security level of storing Keys encrypted and storing the Master-KEK in another place than the Keys.
```
gtema, line 61: Please rephrase the sentence. I do not think that KeyManager API has any "security level" with not storing key encrypted.

Standards/scs-XXXX-wN-key-manager-implementation-testing.md
```diff
@@ -0,0 +1,32 @@
...
+type: Supplement
+track: IaaS
+status: Proposal
+supplements:
+  - scs-XXXX-v1-vN-key-manager-standard.md
+---
+
+## Implementation
+
+A Key Manager service can have different backends, for Barbican these are called Plugins.
```
gtema, line 12: Semantically those are 2 different sentences in Englisch.

Standards/scs-XXXX-wN-key-manager-implementation-testing.md
```diff
@@ -0,0 +1,32 @@
...
+The standard plugin is `simple_crypto`, which has the Master-KEK written in the Barbican config file.
+To secure that Master-KEK when the `simple_crypto` plugin is used it is advised to protect the Barbican config e.g. through running Barbican in an enclave.
```
gtema, line 14: "to [A] when [B] it is advised to [C]" => "when [B] [A] can be achieved by [C]"

Standards/scs-XXXX-wN-key-manager-implementation-testing.md
```diff
@@ -0,0 +1,32 @@
...
+[^1]:[Barbican Plugins](https://docs.openstack.org/barbican/latest/install/barbican-backend.html)
+
+## Automated Tests
+
+The check for the presence of a Key Manager is done with a test script, that checks the presence of a Key Manager service in the catalog endpoint of Openstack.
+This check can eventually be moved to the checks for the mandatory an supported service/API list, in case of a promotion of the Key Manager to the mandatory list.
+
+## Manual Tests
+
+There needs to be a manual test that searches the physical host of the Key Manager host for the presence of a Master-KEK in Plain-Text, e.g. in the Barbican config file.
```
gtema, line 32: Depending on the used plugin the config file may look very differently. One possibility is that the plugin invokes command that returns the key, but the configuration of the command itself contains the key in plain text (or base64). 
I think it makes not much sense to state that there must be a test to check presense of the plain text master key. Instead I would explicitly state that it is out of scope and is in the hands of the CSP to ensure master kek is protected correspondingly.

gtema (changes requested):
@josephineSei, sadly I still have issues with the document "language". It is very hard to read the document and some sentences need to be re-read few times. This should not be the case with standards.

Standards/scs-XXXX-vN-key-manager-standard.md
josephineSei: I broke it up into two sentences.

Standards/scs-XXXX-wN-key-manager-implementation-testing.md
josephineSei: I think an audit is needed somehow, but how about an audit of the architecture?

Standards/scs-XXXX-wN-key-manager-implementation-testing.md
josephineSei: I rephrased it.

Tests/iaas/key-manager/check-for-key-manager.py
```diff
@@ -0,0 +1,81 @@
...
+    # parse cloud name for lookup in clouds.yaml
+    cloud = os.environ.get("OS_CLOUD", None)
+    if args.os_cloud:
+        cloud = args.os_cloud
+    assert cloud, (
+        "You need to have the OS_CLOUD environment variable set to your cloud "
+        "name or pass it via --os-cloud"
+    )
+
+    return check_presence_of_key_manager(cloud)
```
markus-hentsch, line 77: Now that we include the requirement of giving `member`s access to the Key Manager API if present, we should move the corresponding test from the Roles Standard over to here: https://github.com/SovereignCloudStack/standards/blob/65d72d4a8d7cb8598133ddb4fd7563d75cc039f1/Tests/iam/iaas-roles/standard-roles-check.py#L136-L189

josephineSei:
Testing whether a user with the member role is allowed to interact with Barbican (That is, that either new SRBAC is used or the policy is changed to integrate the member role) is quite difficult:
In old deployments, or deployments where all services are using the old RBAC policies it is possible to reconnect only with the member role and check if the user is allowed to list secrets.
In nowadays and future deployments, this will not work anymore.
OpenStack does not offer any API to see what roles a user itself has, nor does it allow to only interact as having one specific role.

So we need to put the burden of a correct configuration into the CSPs hand: We need a user who only has the member role in the project, that is specified when executing the test.

If a CSP also gives out a reader role or a creator role, this will lead to a false positive: The test will succeed, but the standard may not be fulfilled.

We should document this behavior. But when the user-role-project triple is configured correctly, we can test the policy.

Tests/iaas/key-manager/check-for-key-manager.py
artificial-intelligence: > I started integrating the code until I stumbled about this:
> 
> ```
> keystoneauth1.exceptions.http.Unauthorized: Error authenticating with application credential: Application credentials cannot request a scope. (HTTP 401) (Request-ID: req-afca1cf7-0c4f-49d9-b716-44c7bb4f6f72)
> ```
> 
> Now we are investigating, whether there are other ways to ensure that a logged in user only has the member role within the specified project.

@josephineSei mhm, weird, I didn't test it myself, but this should work?

see: https://opendev.org/openstack/keystone/commit/52da4d0e129048d1b808bdde07364cde698cf475

artificial-intelligence:
> Testing whether a user with the member role is allowed to interact with Barbican (That is, that either new SRBAC is used or the policy is changed to integrate the member role) is quite difficult: In old deployments, or deployments where all services are using the old RBAC policies it is possible to reconnect only with the member role and check if the user is allowed to list secrets. In nowadays and future deployments, this will not work anymore. OpenStack does not offer any API to see what roles a user itself has, nor does it allow to only interact as having one specific role. [...]

can't we simply test this by e.g. calling a known admin only api endpoint and expecting it to fail? this isn't sufficient of course (this API could be disabled even for admins via a special crafted policy I guess). But for "normal" deployments, we should receive a HTTP 403 forbidden error code.

It's not bulletproof, but maybe better than nothing?

gtema:
> Testing whether a user with the member role is allowed to interact with Barbican (That is, that either new SRBAC is used or the policy is changed to integrate the member role) is quite difficult: In old deployments, or deployments where all services are using the old RBAC policies it is possible to reconnect only with the member role and check if the user is allowed to list secrets. In nowadays and future deployments, this will not work anymore. OpenStack does not offer any API to see what roles a user itself has, nor does it allow to only interact as having one specific role.
> 

??? I do not get it. When you authorize you get back a token response. It contains all roles user have in the current scope (and for that you do not need to do any other call). In addition to that there is https://docs.openstack.org/api-ref/identity/v3/index.html#id627 but that requires you to have privileges for that (admin)

Tests/iaas/key-manager/check-for-key-manager.py
josephineSei: I have a devstack that already partly includes the new secure RBAC roles. They are enabled in Keystone (and Nova and Glance I think), but not in Barbican. Maybe that is the reason for this behavior. I can take another look into this.

josephineSei:
In the [sdk documentation](https://docs.openstack.org/openstacksdk/latest/user/proxies/identity_v3.html) I found this:
```
validate_user_has_project_role(project, user, role)

    Validates that a user has a role on a project

    Parameters:

            project – Either the ID of a project or a [Project](https://docs.openstack.org/openstacksdk/latest/user/resources/identity/v3/project.html#openstack.identity.v3.project.Project) instance.

            user – Either the ID of a user or a [User](https://docs.openstack.org/openstacksdk/latest/user/resources/identity/v3/user.html#openstack.identity.v3.user.User) instance.

            role – Either the ID of a role or a [Role](https://docs.openstack.org/openstacksdk/latest/user/resources/identity/v3/role.html#openstack.identity.v3.role.Role) instance.
```

Which would be really nice to use. But I need a role id (which does not seem to work - I have the member role but get a "False" from this method, when providing the id of the member role). Even if it would work, as a simple user without the admin role I am not able to list roles and their ids. 

Additionally in the OpenStackSDK, the connection has the properties: `current_project_id` and `current_user_id` - but nothing about the roles, that are inherited.

gtema:
> 
> Additionally in the OpenStackSDK, the connection has the properties: `current_project_id` and `current_user_id` - but nothing about the roles, that are inherited.

The token contains list of **all** roles (inherited roles are resolved and thus explicitly included). But there is a bug in Keystone (https://bugs.launchpad.net/keystone/+bug/2030061) for which fix was finally decided and merged on Friday where application credentials created with/without explicit specifying the AppCred roles are having this resolving not working properly (AppCreds created with i.e. requesting "member" role were not getting "reader" role included). This is not relevant here directly, but may be important to know when you use AppCreds for checks.

Tests/iaas/key-manager/check-for-key-manager.py
josephineSei: I now check the roles of the current user, when there is an admin or manager role present (or any custom role), the check for the creation of a secret will be skipped and a warning logged.

bitkeks (approved):
lgtm. 

very Barbican-centrific, but that's another topic.
"Simple Crypto Plugin" should be forbidden in case we make the Key Manager mandatory (and Barbican is the reference)

artificial-intelligence (changes requested):
one of the tests seems at least unclear or at worst incomplete to me. see the comment for details.

Tests/iaas/key-manager/check-for-key-manager.py
```diff
@@ -26,6 +27,53 @@ def connect(cloud_name: str) -> openstack.connection.Connection:
...
+        }
+    }
+
+    has_member_role = False
+    request = conn.session.request(auth_data['auth_url'] + '/v3/auth/tokens',
+                                   'POST',
+                                   json={'auth': auth_dict})
+    for role in json.loads(request.content)["token"]["roles"]:
+        role_name = role["name"]
+        if role_name == "admin" or role_name == "manager":
```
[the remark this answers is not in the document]
josephineSei: It is intended that the user has at maximum the member role. (I will change the docstring).

Yes if a CSP changes roles for users (actively) and all the corresponding policies to allow more than the member role does, and a test user has the member role and the custom role, we will get a false positive (check will pass) on this test. But the only other way (application credentials) failed within my deployment with the new secure RBAC. Unfortunately there is  currently no way for a simple user to ask OpenStack, which roles they have in a certain project.

Within the new secure RBAC we will only have the roles: reader, member, manager and admin available for users. I do not think, that these will change from OpenStack side soon, because the the work for secure RBAC is going on now for years and after adding the manager role, there was no change to role names.

josephineSei:
I looked of the standard today and adjusted some spelling mistakes left.
@gtema and @artificial-intelligence could you please check if there are still open points from your side? Thank you

gtema (approved):

Standards/scs-XXXX-vN-key-manager-standard.md
artificial-intelligence: from my pov this conversation is resolved, @gtema ?

artificial-intelligence (approved):

Standards/scs-XXXX-vN-key-manager-standard.md
flyersa: You need barbican anyway alone as a requirement for octavia if you do ssl offloading with amphoras too. Also other Keymanager appliances plugin into "barbican" as drivers/plugins

josephineSei:
I asked today in the IaaS call for reviews from CSPs - because they are encouraged to integrate a key-manager sooner or later. I will leave this PR open until the end of this week, so feedback can still be written here.

Issue #591, opened 2024-05-16
Introduce a formal grace period for certificate-scope versions.


mbuechse:
Let's just say sometimes it may be desireable for some certificate to be extended, either publicly or discreetly.

mbuechse:
This can be implemented in multiple ways.

- add new date `obsoleted_at` in between `stabilized_at` and `deprecated_at` (and use obsoleted as grace)
- say that for N weeks after deprecation, we may still show the status for this version, but mark it in some way (put an asterisk, or use yellow instead of green etc.)
- more ideas?

mbuechse:
This should be resolved with #668.

PR #605, opened 2024-05-27
Introduce mechanism for a formal grace period

mbuechse:


Standards/scs-0003-v1-sovereign-cloud-standards-yaml.md
```diff
@@ -115,18 +115,34 @@ then a certificate of that prerequisite scope has to be presented before the cer
...
+This means that more than one version may be eligible at a certain point in time. Tooling SHOULD default
 to use the newest allowable version (the one with the most recent `stabilized_at` date) then.
 
-Note: We intend to keep only one version in effect, except for a grace period of 4 to 6 weeks, when two versions
-are effective at the same time.
+A version is deemed _obsolete_ at a given date if
+
+- `obsoleted_at` is before or at the given date.
+
+Any test result that refers to an obsolete, yet effective version MUST be accompanied by a notice that
```
anjastrunk, line 139: Do you mean "conformance test result"?

Standards/scs-0003-v1-sovereign-cloud-standards-yaml.md
```diff
@@ -115,18 +115,34 @@ then a certificate of that prerequisite scope has to be presented before the cer
 | --------------- | ------------- | ------------------------------------------------------------------------------- | ------------ |
 | `version`       | String        | Mandatory: Version of the particular list of standards                          | _v3_         |
 | `standards`     | Array of maps | Mandatory: List of standard descriptors for this particular layer               |              |
-| `stabilized_at` | Date          | ISO formatted date indicating the date after this version is considered stable. | _2022-11-09_ |
-| `deprecated_at`  | Date          | ISO formatted date indicating the date on which this version is expired.        | _2023-04-09_ |
+| `stabilized_at` | Date          | (ISO formatted) from then on, this version is considered stable (see remarks)   | _2022-11-09_ |
+| `obsoleted_at`  | Date          | (ISO formatted) from then on, this version is considered obsolete (see remarks) | _2023-03-01_ |
```
anjastrunk, line 119: Standard [scs-0001-v1-sovereign-cloud-standards](https://github.com/SovereignCloudStack/standards/blob/issue/591/Standards/scs-0001-v1-sovereign-cloud-standards.md) does not mention a state, called "obsolete". Instead, standard mentions "Draft", "Stable" and "Deprecated". It it makes sense, to introduce "obsolete" in [scs-0001-v1-sovereign-cloud-standards](https://github.com/SovereignCloudStack/standards/blob/issue/591/Standards/scs-0001-v1-sovereign-cloud-standards.md) in favor of consistency.

anjastrunk (changes requested):
See my comments

Standards/scs-0003-v1-sovereign-cloud-standards-yaml.md
mbuechse: Yes. I will add this word.

Standards/scs-0003-v1-sovereign-cloud-standards-yaml.md
mbuechse: I strongly disagree. This is a purely formalistic argument. There is (as of now) just no need for this additional state in scs-0001, it would make matters unnecessarily more complicated for no practical gain, and we have already decided against it in SIG Std/Cert.

anjastrunk: Reading [scs-0001-v1-sovereign-cloud-standards](https://github.com/SovereignCloudStack/standards/blob/issue/591/Standards/scs-0001-v1-sovereign-cloud-standards.md) more carefully, I figured out, that we already have an obsolete state. scs-0001-v1-sovereign-cloud-standards](https://github.com/SovereignCloudStack/standards/blob/issue/591/Standards/scs-0001-v1-sovereign-cloud-standards.md) says:

> Should a document become obsolete, it can be deprecated.
>
> Obsoletions SHOULD be announced ahead of their execution by setting the deprecated_at field to a future date and moving the status to Deprecated. This signals current and future implementors that the subject matter of the document is not considered necessary or state of the art anymore.

This means, the definition of obsolete and deprecated is a follows as:

- obsolete (implicitly defined):  A standard is obsolete, if there is a deprecated date that is after the current date
- deprecated (implicitly defined): A standard is deprecated, if there is a deprecated state that is equals or before the current date.

So I agree, we do not need an explicitly obsolete **state** and I really doubt, if we need an 'obsolete' **date** in SCS Certification YAML at all, as it is already given implicitly, with a deprecated date in the future. Instead of introduction of an obsolete **date**, I suggest to define the meaning of a future, present and past deprecated date more precisely.

anjastrunk (changes requested):
Clarify, if we really need an additional 'obsolete' date.

Standards/scs-0003-v1-sovereign-cloud-standards-yaml.md
mbuechse: I am sorry, but that is not true.

First of all, scs-0001 and scs-0003 are independent of one another. Of course, we nevertheless try not to introduce confusing differences.

Second, scs-0001 does not have two states, it merely has bad wording. Hannes changed `obsolete` to `deprecated` and he seems to have missed a few occurrences.

mbuechse:
This PR shows but one possible way of implementing the goal of a grace period. See [this comment](https://github.com/SovereignCloudStack/standards/issues/591#issuecomment-2144875858)

mbuechse:
Closing this in favor of #668

PR #608, opened 2024-06-04
Update auth_url endpoint and app credential id for poc-wgcloud

martinmo:
The endpoint changed, we now use `l1a` instead of `l1` in the URL.


garloff (approved):
This works, thanks.

garloff:
This also resolves #614

PR #609, opened 2024-06-05
GitHub actions based DCO check

martinmo:
As a (maybe temporary) workaround for the currently malfunctioning DCO check.

Downside: this action has to be added to every repository in the org – in other words, this PR fixes only the `standards` repo.

The DCO app "required status check" must be disabled in the ["Branches" settings of this repo](https://github.com/SovereignCloudStack/standards/settings/branches).

### DIFF
--- a/.github/workflows/dco-check.yml
+++ b/.github/workflows/dco-check.yml
@@ -11,10 +11,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
 
     - name: Set up Python 3.11
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
       with:
         python-version: '3.11'
         cache: 'pip'
--- a/.github/workflows/dco-check.yml
+++ b/.github/workflows/dco-check.yml
@@ -1,0 +1,27 @@
+# Based on the example described in:
+#   https://github.com/christophebedard/dco-check/blob/master/README.md#github
+# plus version pinning and caching.
+
+name: DCO check
+
+on: push
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


martinmo:
I've tested it with a non-conformant commit, which can be seen in the Actions log: https://github.com/SovereignCloudStack/standards/actions/workflows/dco-check.yml (this branch originally contained an empty commit without the `Signed-off-by` tag).

berendt:
Is it likely that it will be broken for longer? Otherwise we have to check the sign-off manually for the moment. I think that's better than having to touch all repos or having a snowflake on individual repos. Just my 2 cents.

martinmo:
> Is it likely that it will be broken for longer? Otherwise we have to check the sign-off manually for the moment. I think that's better than having to touch all repos or having a snowflake on individual repos. Just my 2 cents.

That's a good question! The DCO app doesn't seem to get a lot of attention these days, and there have been several outages in the past before the one this week, e.g., https://github.com/dcoapp/app/issues/162.

This PR is just a proposal.

martinmo:
PS. Wrt to attention, the last release of the dco-check package which I use here is also 2 years ago :(

berendt:
There is a DCO job for Zuul. Perhaps it would be better to use this?

https://zuul-ci.org/docs/zuul-jobs/latest/general-jobs.html#job-dco-license

berendt:
This is the related issue: https://github.com/dcoapp/app/issues/211

Looks like the bot is up & running again.

garloff:
In general, I am in favor of merging this into just this one repository.
So we have a replacement for a broken github DCO check which is used here in real life.
Based on what happens with github's own DCO, we have thus an alternative in stock.

garloff:
Hmm, as DCO works again, we don't need this right now.
Should we close this and reopen, if the github folks continue to do a bad job and we see failure again?

martinmo:
I'll close it, we can revive it when needed.

PR #610, opened 2024-06-05
Update appcred ID for artcodix

garloff:
A new appcred that allows for creating a VM to do the entropy test.


chrisschwa (approved):
LGTM

garloff:
I have meanwhile also updated the secret, so the merge should result in a working check.
Thanks, @mbuechse and @chrisschwa !

PR #600, opened 2024-05-23
Update cert table to use v4 of scs-compatible iaas, for all the older…

mbuechse:
… versions have expired

PR #613, opened 2024-06-05
Use IaaS v4 badges for all clouds

berendt:
According to https://docs.scs.community/standards/scs-compatible-iaas v4 is effective and v3 is deprecated as of 2024-04-30. We should show the current status in the table in the README.


mbuechse:
Duplicate of https://github.com/SovereignCloudStack/standards/pull/600

PR #623, opened 2024-06-11
Upgrade dependencies

mbuechse:
Hopefully subsumes
https://github.com/SovereignCloudStack/standards/pull/621
https://github.com/SovereignCloudStack/standards/pull/617
https://github.com/SovereignCloudStack/standards/pull/616

### DIFF
--- a/Tests/requirements.txt
+++ b/Tests/requirements.txt
@@ -4,21 +4,19 @@
 #
 #    pip-compile requirements.in
 #
-aiohttp==3.9.4
+aiohttp==3.9.5
     # via
     #   -r requirements.in
     #   kubernetes-asyncio
 aiosignal==1.3.1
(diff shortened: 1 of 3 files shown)


martinmo:
That's a big jump for the openstacksdk, from 0.103.x to 3.1.x. And I wonder why the Python unit tests did not run. I thought with https://github.com/SovereignCloudStack/standards/blob/main/.github/workflows/test-python.yml#L6-L10 this should happen.

martinmo:
Ok, only the initial push triggered the pipeline with the python tests, which ran successfully. The second push only contained changes for npm and that's why python tests don't run. Probably, that's the difference between the `push` and `pull_request` triggers in the github workflow file.

martinmo (approved):
Let's try it.

PR #626, opened 2024-06-11
Fix templates: new documents start out in Proposal state

mbuechse:


Standards/scs-XXXX-vN-decision-record-template.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: _Descriptive title_
 type: Decision Record
-status: Draft
+status: Proposal
```
garloff, line 4: I think we still call it `Draft` for Standards, correct?
I'd prefer to keep the terms aligned ...

mbuechse: No, and that is precisely the point of this PR.

Quoting https://docs.scs.community/standards/scs-0001-v1-sovereign-cloud-standards#proposal-of-a-new-document

> The metadata MUST indicate the intended `track` and `type` of the document, and the `status` MUST be set to `Proposal`

mbuechse: See also here: https://docs.scs.community/standards/scs-0001-v1-sovereign-cloud-standards#process

Proposal -> Draft -> Stable

with a "loose consensus" at each transition -- which is good, because that's how these proposals get exposure

garloff: Ah, I was misremembering the workflow, sorry. Ignore my comment then.

garloff (approved):

PR #630, opened 2024-06-12
Proposal of scs-0103-v2-standard-flavors.md to lift restriction…

### LINKED ISSUE #594: Relax scs-0103-v1, introducing scs-0103-v2
This is a follow-up to #574. The next version of this standard should no longer forbid setting `scs:name-vN`. It should, however, mandate that once such a property is present for some flavor, then this flavor has to have all of them (for all values of N, that is, all major versions of scs-0100). It should furthermore mandate that the names then be correct according to the respective major version standard.

mbuechse:
…on `scs:name-vN`

Note that this is not merely a restriction, because it now requires `scs:name-v3` to be set for all standard flavors.

resolves #594


Standards/scs-0103-v2-standard-flavors.md
```diff
@@ -0,0 +1,175 @@
...
+  environments to ensure uniformity across SCS clouds. Mandatory and recommended flavors are defined with specific
+  configurations of vCPUs, vCPU types, RAM, and root disk sizes, alongside extra specs like scs:name-vN, scs:cpu-type,
+  and scs:diskN-type to detail the flavor's specifications. This standard facilitates guaranteed availability and
+  consistency of flavors, simplifying the deployment process for DevOps teams.
+---
+
+## Introduction
+
+This document is a revision of [scs-0103-v1](scs-0103-v1-standard-flavors.md) that lifts the following
+restriction regarding the property `scs:name-vN`: this property may now be used on any flavor, rather than
```
garloff, line 18:
```suggestion
restriction regarding the property `scs:name-vN`: this property may now be used on any SCS-flavor, rather than
```

garloff: Thinking about it some more:
* Maybe we have existing clouds where providers do have existing non-SCS flavors and want to add an `scs:name-vN` property to just indicate how the name would look like according to SCS systematic flavor naming to increase compatibility with SCS. We probably don't want to discourage this.
* Thus ignore my suggestion ...

garloff (approved):
And I like the revision.

garloff:
I approved it and think we should merge.
@berendt: Any concerns?
@mbuechse, @fkr - anything you want to bring into team/SIG discussions before we do so?

garloff:
Side Questions:
* We used to generate the input for the flavor manager to avoid any inconsistencies caused by manually created yaml files ... This tool was in Tests/iaas/flavor_naming/flavor-manager-input.py
Shouldn't we update it to generate the extra_specs pieces? Including the new scs:namev3 information?
* I am not sure I understand the need for scs:namev3 -- it will always be identical to namev2 as far as I read the current standards. So I wonder why it was introduced with this PR.

mbuechse:
@garloff 

* No, the current script is https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/standard-flavors/flavor-manager-input.py
* The reason is that many people don't know prima facie that these are identical, and they shouldn't have to, should they?

mbuechse:
@garloff With the revised GPU part in the flavor naming, we will have a version 4 of the standard, and then **each standard flavor** will have the same name for v2, v3, and v4. And who knows what versions we will get in the future, and they probably won't change the names of the standard flavors either (because that would be bad). So maybe we need a different solution altogether!

mbuechse:
I suggest to use https://github.com/SovereignCloudStack/standards/pull/677 instead.

PR #636, opened 2024-06-14
Improve badges: show version number, link to Github action

mbuechse:
as discussed in SIG Std/Cert of 2024-06-13
https://input.scs.community/2024-scs-sig-standardization?both#Update-table-of-compliant-clouds


gtema (commented):
What if we name the label "compliance" instead if "compliant"? I am a bit unhappy with "compliant: failing" from the wording point of view.
Otherwise looks great

gtema (approved):
cool, I like it

PR #637, opened 2024-06-14
Bugfix: an erroneous br tag got sneaked into the README

mbuechse:
it was found in the docs repo by Docusaurus, but apparently no one cared to fix it here, so I copied the mistake back into the docs repo... now that it's fixed over there, it's time to finally fix it here as well


gtema (approved):
in it's own change is ok, the check is unclear to me now

Issue #426, opened 2023-12-19
[Other] Evaluate costs of making a cluster SCS compliant 


anjastrunk:
Epic issue to evaluate costs to make existing, non-compliant OpenStack and K8S cluster SCS compliant. In doing so, we set up a non-compliant cluster with the open source life Lifecycle Management Tool for OpenStack and K8S [Yaook](https://yaook.cloud/).

This issue acts as epic.

**General** 

- [ ]  https://github.com/SovereignCloudStack/standards/issues/414

**IaaS**

- [ ]  https://github.com/SovereignCloudStack/standards/issues/415
- [ ] https://github.com/SovereignCloudStack/standards/issues/416
- [ ] https://github.com/SovereignCloudStack/standards/issues/417
- [ ] https://github.com/SovereignCloudStack/standards/issues/418
- [ ] https://github.com/SovereignCloudStack/standards/issues/419

**KaaS**

- [ ] https://github.com/SovereignCloudStack/standards/issues/420
- [ ] https://github.com/SovereignCloudStack/standards/issues/421
- [ ] https://github.com/SovereignCloudStack/standards/issues/422
- [ ] https://github.com/SovereignCloudStack/standards/issues/423
- [ ] https://github.com/SovereignCloudStack/standards/issues/424
- [ ] https://github.com/SovereignCloudStack/standards/issues/425

anjastrunk:
> I created a document to show what is necessary to make a Yaook cluster (OpenStack) compliant with SCS. It basically combines the standards into an easy to read document (and make it a bit easier). Some information get lost, but I also reference to the original documents, so every reader knows where to go if questions arise.
> 
> I'm not entirely sure where to post this document to keep the SCS-Team in the loop on this, but I think I will just create a branch that will never be merged. The most current version can always be found in our internal infrastructure, since the updates here will maybe be behind 1-2 days.
> 
> Edit: https://github.com/SovereignCloudStack/standards/compare/do-not-merge/scs-compliant-yaook contains the documents.

Good idea. General outcomes on how to make arbitrary OpenStack/KaaS cluster SCS compliant, should be documented in the corresponding issue, as listed in this epic. Furthermore, we should write blog articles, whenever a topic is value-added for CSPs and SCS community.

cah-hbaum:
I started to setup an independent K8s cluster that will be used to test the K8s standards. The cluster is already up and running, next step is to apply the standards.

cah-hbaum:
[...]

Issue #569, opened 2024-04-18
[Feature Request] Require up-to-date cluster-api images for OpenStack


mxmxchere:
The problem that brings us here is how to update and keep track with upstream kubernetes versions when using the cluster-stacks approach on OpenStack. As the [SCS K8S Version Policy](https://docs.scs.community/standards/scs-0210-v2-k8s-version-policy#motivation) formulates, SCS is quite eager to keep track with the upstream work and does not want to fall behind:

> We want to achieve an up-to-date policy, meaning that providers should be mostly in sync with the upstream and don't fall behind the official Kubernetes releases. 

We came up with two solutions that are described in today's [container minutes](https://github.com/SovereignCloudStack/minutes/blob/main/container/20240418.md#make-use-of-the-clusterspectopologyversion-field-b1-5min)

In short the two Options are:
A.) Mandate the CSP to keep cluster-api Images up-to-date, then the kaas-user can choose the kubernetes version by setting it in the cluster-resource.

B.) Use CSCTL to creates a new cluster-stack for each new kubernetes version, release it (automation is still needed here), change cluster-Stack-CRD (for each tenant) to reconcile the new clusterstack (creating a new clusterclass), let the CSPO upload the new image to OpenStack (for each tenant). Then the user can update the cluster resource (either the clusterclass, which in turn uses a new hardcoded image and/or the kubernetes version either automatically or via a webhook.)

If i understand [scs-0104-v1-standard-images](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0104-v1-standard-images.md#image-specification-class-of-images) correctly we currently can not require a class of images:

> If the status is mandatory, then at least one image MUST be present whose name matches the regular expression given via name_scheme.

We would need to change "at least one image" to "all images". Afterwards we could mandate the class of Cluster-API images to be updated.

Any thoughts / ideas / feedback appreciated @jschoone @garloff

berendt:
For A: We provide up-to-date images here: https://github.com/osism/k8s-capi-images and also added a OSISM CLI command (osism manage image clusterapi) to have a way a CSP can provide the images in no time. Optional image definitions for the openstack-image-manager are also available.

jschoone:
Hi @mbuechse, I put this on the [agenda](https://input.scs.community/2024-scs-sig-standardization?both) of the next SIG Standardization. Does it fit?

PR #643, opened 2024-06-19
Extend implementation notes

cah-hbaum:
Add and extend some implementation notes derived from the findings of https://github.com/SovereignCloudStack/standards/issues/426.


cah-hbaum:
Not sure if the link checker is good in this state or how links should be entered.
I wrote in some links to a file that will be create WITH this PR. I already wrote in the `main` branch, since it WILL be there soon, but it isn't yet, so this gets marked as a dead link...

Which would be the correct workflow here? (maybe @martinmo, I think he worked on this)

Standards/scs-0102-w1-image-metadata-implementation-testing.md
```diff
@@ -0,0 +1,52 @@
...
+Additionally, a user can also decide to test their private images, although this isn't a necessity.
+
+### Errors and warnings
+
+The test provides increases its return value for every error found during execution.
+Additionally, logs are provided to output further information:
+
+- Error   for invalid property values or missing properties
+- Warning for missing images or not recommended values
+- Info    for violations of recommendations
```
mbuechse, line 43: The sections on errors and warnings are not meant to describe the implementation, but rather to serve as a specification for it. They ought to describe any deviations from the general guidelines (https://docs.scs.community/contributor-docs/development/tests/rfc2119-keyword-test-guide).

Said deviations are possible whenever the standard is particularly hard to test.

mbuechse (commented):
Thanks for the good work! The supplements should be a bit more succinct; they are not stand-alone documents, and they shouldn't duplicate content from the main document. Thay should, however, be actionable, and I think this has been achieved. Unfortunately, for one of the documents, there is a conflict with another pending (albeit approved PR), and I have some reservations about the new spec file.

cah-hbaum:
Most things should be adressed with https://github.com/SovereignCloudStack/standards/pull/643/commits/804b5c34367a830f0a646a96aa3afd36f43344ad

Tests/iaas/SCS-Spec.Images.yaml
```diff
@@ -0,0 +1,107 @@
+---
```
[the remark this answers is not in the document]
cah-hbaum: I understand where you come from, but we already provide a set of standard images in the form https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/scs-0104-v1-images.yaml.
So IMO, its not that much of a deal to also provide this other file, since its practically is derived from the first one.
And we also don't say that this is the only solution, but its quite a good guidance.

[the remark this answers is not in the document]
mbuechse: The set of standard images you're referring to is merely fixing the image source, so people know what to expect when they encounter a certain name. The spec file, on the other hand, contains quite a lot of specific details that I (as a standards person) have absolutely no expertise about. So I'm not convinced.

Standards/scs-0104-w1-standard-images-implementation-testing.md
```diff
@@ -0,0 +1,47 @@
+---
+title: "SCS Standard Images: Implementation and Testing Notes"
```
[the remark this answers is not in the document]
mbuechse: That PR has been merged, so once you rebase this branch, you can merge the two files. Mind you that the file is called
https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0104-w1-standard-images-implementation.md (without `-testing`)

Standards/scs-0101-w1-entropy-implementation-testing.md
```diff
@@ -7,18 +7,26 @@ supplements:
   - scs-0101-v1-entropy.md
 ---
 
-## Implementation
+## Implementation notes
 
 We presume that almost nothing has to be done (or indeed can be done), as
-long as the CPUs and VM images are reasonably recent; only the flavor and
-image attributes have to be set:
+long as the CPUs and VM images are reasonably recent.
```
artificial-intelligence, line 13: I think this is overly vague for an implementation note. What does "recent" mean here?

I would prefer if we could list the minimum kernel and qemu/libvirt/openstack versions required here, don't you think? At least the kernel version in the guest vm should be explicitly called out imho as there might be reasons for some guests to run older kernels.

Standards/scs-0102-w1-image-metadata-implementation-testing.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Images sample
+
+Some checks need to be performed on a live instance. All publicly available images on this instance
+will be checked for either only the mandatory properties or possibly also the recommended ones.
+Additionally, a user can also decide to test their private images, although this isn't a necessity.
+
+### Implementation
+
+The script [`image-md-check.py`](https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/image-metadata/image-md-check.py)
```
artificial-intelligence, line 31: I think we should use a (somewhat) stable permalink here:
```suggestion
The script [`image-md-check.py`](https://github.com/SovereignCloudStack/standards/blob/59683d8fc592d5d27152be7dab32290d4c84316b/Tests/iaas/image-metadata/image-md-check.py)
```

Standards/scs-0210-w1-k8s-version-policy-implementation-testing.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Notes
+
+The test for the [K8s Version Policy Standard](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0210-v2-k8s-version-policy.md)
+can't be used like most other tests provided in the SCS standards repository.
+It aims on testing a KaaS offering of a CSP with the creation of the most recent
+Kubernetes version provided by the CSP. Since only this part is tested,
+the test doesn't have any validity for a single cluster, since the updating and
+versioning should be decided by the user, which could want to leave a server
+on a version just for stability reasons.
```
artificial-intelligence, line 31: I don't think we should suggest to users that it is okay to leave a server on a certain version or that they could achieve increased stability this way.

Outside of very niche usecases the opposite is true: pinning versions for stability will often lead to instability due to bugfixes not getting deployed.

also, afauj technically, the script https://github.com/SovereignCloudStack/standards/blob/main/Tests/kaas/k8s-version-policy/k8s_version_policy.py can test different versions, not just the latest?

artificial-intelligence (changes requested):

Standards/scs-0101-w1-entropy-implementation-testing.md
mbuechse: True. We should indicate what CPUs and kernel versions are required. I don't know about any qemu/libvirt/openstack requirements.

Standards/scs-0102-w1-image-metadata-implementation-testing.md
mbuechse: I disagree. The links are supposed to be kept alive using our continuous-integration system. I don't want them to point to some super-old version.

artificial-intelligence: The CI can't guarantee this as long as the CI is not immutable itself, which it isn't.
Any practice from the last 15 years writing documentation tells me that links can and will rot away.

Regardless, I don't think this is an important enough issues, to warrant a long winded discussion, as everything can be retrieved via some git archeology, if really needed.

Tests/iaas/SCS-Spec.Images.yaml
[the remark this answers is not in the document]
martinmo: I agree with Hannes here: to make it easier for a CSP to become compliant, it would be really nice to have a complete, working YAML that can just be fed to the `openstack-image-manager`. At the moment, you'll need to assemble parts of it yourself (I exercised this recently):

- go to https://github.com/osism/openstack-image-manager/tree/main/etc/images
- find and copy ubuntu and debian files
- filter unnecessary things from the ubuntu files (e.g., versions 18.04 and older, "Minimal" variants)
- for the CAPI images, it is even more difficult, because the `kubernetes.yml` in that directory is not up-to-date
  - in the OSISM stack, the `osism` cli tool is used for this purpose (`osism manage image clusterapi` as mentioned [here](https://github.com/SovereignCloudStack/standards/issues/569#issuecomment-2063919012))
  - it creates an up-to-date YAML on the fly and feeds it to the image manager (https://github.com/osism/python-osism/blob/main/osism/commands/manage.py)

However, I disagree with the file name choice: it should reflect that it is meant for v1 of scs-0104 and to be fed to the image manager.

martinmo:
Because Hannes is blocked, I'll adopt this PR

Tests/iaas/SCS-Spec.Images.yaml
```diff
@@ -0,0 +1,107 @@
...
+      image_description: "https://cloud.debian.org/images/cloud/bullseye/latest/"
+      url: https://cloud.debian.org/images/cloud/bullseye/latest/debian-11-generic-amd64.qcow2
+      checksum: "sha512:8bc1f39a91ede83a15248de7c5b69ad6eae2094db45c7e6c22cf3a6fb774878a37f87575b829c19358389ec1f98dc4d5f12ae996977057ce2e9c29d88b871d57"
+      build_date: 2024-04-29
+    - version: '12'
+      os_version: '12'
+      image_description: "https://cloud.debian.org/images/cloud/bookworm/latest/"
+      url: https://cloud.debian.org/images/cloud/bookworm/latest/debian-12-generic-amd64.qcow2
+      checksum: "sha512:cae81bd82490bb68aaa554f4573fecd09d19b9bf5feccdcd2c42eabba7ac507cda8b7fa53983e1a843c6334bbb684142897dda2990d2fe159261bca3308c4333"
+      build_date: 2024-04-29
```
martinmo, line 107: I believe using `versions` this way is wrong and the files in https://github.com/osism/openstack-image-manager/blob/main/etc/images/debian.yml are organized in another way, I'll fix this.

Tests/iaas/SCS-Spec.Images.yaml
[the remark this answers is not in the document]
mbuechse: Yes, people will have to do this. However, we can't keep this file up to date, which is what OSISM (semi-) automatically does in their repo. That we used the `versions` wrong is just the evidence that we shouldn't meddle in this.

Standards/scs-0101-w1-entropy-implementation-testing.md
mbuechse: 65edea5720ccce61b0d37414e114e2848ffb7d33

Standards/scs-0210-w1-k8s-version-policy-implementation-testing.md
mbuechse: c2a860a7cccd5684bbfea0b69fbbf6b26e0e6bae

artificial-intelligence (approved):
LGTM as far as I can see on mobile

PR #645, opened 2024-06-22
Feat/flavor0103 add extras

garloff:
This tool can add scs:name-v1, name-v2, cpu-type and disk0-type to existing SCS- flavors in an automated fashion.
These properties are required by https://docs.scs.community/standards/scs-0103-v1-standard-flavors#properties-extra-specs
To test, just call it without admin privileges :-), optionally add `-d` or `--debug`.


garloff:
Note 1: If you create your flavors with the [flavor-manager](https://github.com/osism/openstack-flavor-manager), you won't need this.
Note 2: I have not checked if flavor-manager does have a mode where it will just add the missing properties (extra_specs) to existing mandatory (and recommended) flavors without changing anything else -- if it does, you may not need this little tool.
Note 3: Even then, you may have additional SCS flavors where you may want to set the extra_specs according to the logic. This used to be illegal and was recently changed. I expect we will change this a little bit:
- Move the extra_specs properties to scs-0100, because it describes a scheme for these settings
- Introduce an addition files that says: scs:status: mandatory or recommended (and leave this field out for non-prescribed SCS flavors according to scs-0103).

garloff:
@PSwatchmen: Running this on SCS1 should make it SCS IaaS-v4 compatible.

fkr:
Why exactly do we need this if we have the flavor manager? If the flavor manager does not suffice shouldn't we just add the missing pieces to the flavor manager instead of adding another tooling?

garloff:
@fkr: The tools are very different:
Flavor-Manager:
- flavor-manager consumes lists with all details ([OSISM](https://github.com/osism/openstack-flavor-manager/blob/main/flavors.yaml) or [SCS](https://raw.githubusercontent.com/SovereignCloudStack/standards/main/Tests/iaas/SCS-Spec.MandatoryFlavors.verbose.yaml))
- These lists would be buggy, if we did not generate them with [tooling](https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/flavor-naming/flavor-manager-input.py)
- flavor manager does not know about the meaning of names or extra_specs
- it will thus only ever handle the mandatory and recommended flavors that we generated for it

This tool:
- It uses the library [flavor_names.py](https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/flavor-naming/flavor_names.py) that is also used for our [compliance checks}(https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/flavor-naming/flavor-names-openstack.py), the [interactive tool](https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/flavor-naming/flavor-name-check.py) and the [online tool]()https://flavors.scs.community/).
- It thus knows to interpret names and properties and knows how to check these for consistency and how to fill in missing information.
- It can do this for any flavor that follows the SCS- rules, not just the mandatory and recommended flavors from 0103.

frosty-geek:
I agree that this is either a "side-car" tool of the flavor manager and thus should live in said git repo, or even better should be able to "update" those flavors that it is configured to create (obviously there is a way to do this)

mbuechse:
The properties in question are not required, except for the standard flavors, and I would highly recommend using the OSISM flavor manager. It can indeed just set the `extra_specs` for those flavors that already exist, and while the spec file used by the flavor manager could be erroneous due to the flavor manager being agnostic to our flavor naming, the spec file provided in this repo is being checked by our continuous-integration pipeline.

So this new tool is mainly useful for flavors beyond the list of standard flavors, which is perfectly okay. But it only sets the main part and the disk part of the flavor name. This does seem a bit restricted to me. Besides, we might be getting a new revision of the standard https://docs.scs.community/standards/iaas/scs-0103 which will also include the property `scs:name-v3` -- see [this PR](https://github.com/SovereignCloudStack/standards/pull/630). BTW, this new revision expressly states that new versions of scs-0100 shall deal with the name property in the future.

garloff:
> The properties in question are not required, except for the standard flavors, and I would highly recommend using the OSISM flavor manager. 

Agreed.

> It can indeed just set the `extra_specs` for those flavors that already exist, and while the spec file used by the flavor manager could be erroneous due to the flavor manager being agnostic to our flavor naming, the spec file provided in this repo is being checked by our continuous-integration pipeline.

Well we generate the input to the flavor manager, so this should not be error-prone. 

> So this new tool is mainly useful for flavors beyond the list of standard flavors, which is perfectly okay. But it only sets the main part and the disk part of the flavor name. This does seem a bit restricted to me. 

It does not have more information, because OpenStack does not expose much in standardized ways.
So this would need to be passed by the user.

> Besides, we might be getting a new revision of the standard https://docs.scs.community/standards/iaas/scs-0103 which will also include the property `scs:name-v3` -- see [this PR](https://github.com/SovereignCloudStack/standards/pull/630). BTW, this new revision expressly states that new versions of scs-0100 shall deal with the name property in the future.

Ack.

garloff:
I just checked with `osism manage flavors`: It does add the missing properties to existing flavors without doing any further changes to the flavors. So this little helper here is not needed for any of the recommended or mandatory flavors.

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
```diff
@@ -0,0 +1,345 @@
...
+        disk = Disk()
+        disk.disksize = flavor.disk
+        disk.disktype = disk0_type
+        flv = Flavorname(cpuram, disk)
+    else:
+        flv = Flavorname(cpuram)
+    return outname(flv)
+
+
+def check_name_extra(flavor, ver, match, flname):
```
mbuechse, line 115: I have a hard time following the intricate logic of this function. Also, there have been considerable changes to scs-0103-v1. The extra_spec `scs:name-vN` no longer requires a name for version N of the standard on flavor naming; it merely requires a name that is valid for ANY version of the standard.

mbuechse: Oh, and I think it's a very bad idea to destructively update the `flavor.extra_specs` here and then rely on some other party for actually performing an update on the server. We need to collect the changes in a separate `dict` which can then be used to compute the necessary API calls, or, preferably, the command lines that the operator can then execute.

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
```diff
@@ -0,0 +1,345 @@
...
+    if not cloud:
+        print("ERROR: Need to pass -c|--os-cloud|OS_CLOUD env", file=sys.stderr)
+        usage(3)
+
+    conn = openstack.connect(cloud)
+    conn.authorize()
+    compute = conn.compute
+
+    flavors = compute.flavors()
+    for flavor in flavors:
```
mbuechse, line 277: insert a separate step before this loop that filters `flavors` depending on whether all SCS flavors should be processed or a fixed list of flavors supplied via name by the user

mbuechse (commented):
I'm afraid this needs a lot of work, for two reasons:

- the logic is quite intricate and could need some simplification (I included an idea here or there)
- the standard scs-0103-v1 has changed in the meantime, and this should be reflected here

I think the following would be good:

- collect all flavor names from the flavor (be it the actual name or any name given in the extra_specs) and convert to `Flavorname` instances
- check that each of theses instances is consistent with the actual flavor
- MAYBE try to check whether these instance are consistent with one another (because they can have varying degree of detail)
- in case a flavor doesn't have any names in the `extra_specs`: generate one name for v1 of the standard and one for v2 of the standard
- otherwise, check that at least one of the names in the `extra_specs` is valid for a stable version of said standard

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
```diff
@@ -0,0 +1,345 @@
...
+
+Usage: flavor-add-extra-specs.py [options] [FLAVORS]
+Options:
+    -h|--help:  Print usage information
+    -d|--debug: Output verbose debugging info
+    -q|--quiet: Don't output notes on changes performed
+    -t|--disk0-type TYPE:   Assumes disk TYPE for flavors w/ unspec disk0-type
+    -p|--cpu-type TYPE:     Assumes CPU TYPE for flavors w/o SCS name
+    -c|--os-cloud CLOUD:    Cloud to work on (default: OS_CLOUD env)
+    -n|--no-changes:        Do not perform any change
```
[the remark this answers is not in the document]
garloff: Well, rather than performing the operations, you provide input for c'n'p for the changes.
This has some advantages:
- Operator can review the changes -- for understanding or because of lack of trust
- Operator could even add additional things if she wanted to
(Sorry, I don't see the elimination of an option with a global variable here as relevant. I know pylint gives me a lower rating ...)
It also has disadvantages:
- The tool can not just be called in order to accomplish a job. You force it to be interactive
- For lots of changes (several screens full) or for environments with limited c'n'p capabilities (think people that are not using Linux as their main desktop), this can be very inconvenient.

So, not convinced.

What I could imagine however doing: Reverse the default.
By default we don't perform the changes.
And with `-y|--yes-i-really-want-this` things would be performed. 

If you prefer the current `INFO` output to be stated as openstack flavor set ... commands, that should be feasible as well.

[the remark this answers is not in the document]
mbuechse: Reversing the option is fine with me! Couldn't we generate the output and then perform it as well (using the shell)?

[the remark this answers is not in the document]
mbuechse: Regarding the option: I do think we have an unhealthy proliferation of options with our tools. Operators usually don't want to read manuals. However, if the tool is used without the "do it" option, it could output a hint that this option exists.

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
garloff: > Oh, and I think it's a very bad idea to destructively update the `flavor.extra_specs` here and then rely on some other party for actually performing an update on the server. We need to collect the changes in a separate `dict` which can then be used to compute the necessary API calls, or, preferably, the command lines that the operator can then execute.

If we decide we want to output `openstack flavor set ...` commands, then collecting the changes seems a must.
I have no particularly strong opinion on it, except that I do NOT want to force operators to do cut'n'paste.

I don't see why updating the `flavor.extra_specs` fields in-place would be bad. We could however do the API call here right away rather than returning the need for it. There may have been thoughts on more complicated logic that lead to the split and looking at the code, I'm glad that does not seem to be needed.

mbuechse: The problem with updating the official dict in place is that we can not trace afterwards what is only local and what is on the server. Therefore I would collect the changes in a dedicated data structure. A dict (which I suggested above) could work, but in light of the current discussion, I propose to use the _Command pattern_, that is, to introduce an abstract class that represents a change of the flavor, and it provides methods (a) for building a command-line representation of this change, and (b) for executing the change on the openstack connection.

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
```diff
@@ -0,0 +1,344 @@
...
+        # we don't need to update the name.
+        try:
+            parsed = parser_v2(name)
+        except ValueError as exc:
+            try:
+                parsed = parser_v1(name)
+            except ValueError as exc2:
+                print(f"ERROR parsing {spec} {name}: {exc!r} {exc2!r}",
+                      file=sys.stderr)
+                need_name_set = False
```
mbuechse, line 143: I think it might be worthwile to move the class `ParsingStrategy` from [cli.py](https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/flavor-naming/cli.py#L15) to flavor_names.py and to use it here as `ParsingStrategy(parsers=(parser_v2, parser_v1, ))`.

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
garloff: We need flavor instances, not just a list of name.
So something like this would please?
```python
def filter_flavors(conn, specified):
  flavors = []
  for flv in conn.compute.flavors():
    if specified:
      if flv.name in specified:
        flavors.append(flv)
    else:
      if flv.name[0:4] == "SCS-":
        flavors.append(flv)
  return flavors
```

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
garloff: OK, sound fine. I was unaware of its existence :-O

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
mbuechse: I thought more along the lines of:

```python
    if flvs:
        flavors = [flv for flv in flavors if flv.name in flvs]
    else:
        flavors = [flv for flv in flavors if flv.name.startswith('SCS-')
```

or maybe

```python
    predicate = (lambda fn: fn in flvs) if flvs else (lambda fn: fn.startswith('SCS-')
    flavors = [flv for flv in flavors if predicate(flv.name)]
```

mbuechse: Just FYI, I think the second variant could also be written

```python
    predicate = flvs.__contains__ if flvs else (lambda fn: fn.startswith('SCS-'))
    flavors = [flv for flv in flavors if predicate(flv.name)]
```

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
mbuechse: I guess I would even change the class a little bit... We could add an attribute `vstr` and change the `parse` to `__call__` so that it has the same protocol as the class `Parser` from `flavor_names.py`. This would constitute a kind of _Composite pattern_ then ;)

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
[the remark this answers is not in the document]
garloff: Thinking about collecting changes to only do one `openstack flavor set` call per flavor:
* We would probably remember the changes that need to happen in our own data structure with something like a verb (`DELETE`, `ADD`, `SET` or `DELETE`, `POST`, `PUT` if you prefer REST notation), an object and a property. So we'd end up with `[("DELETE", "scs:name-v3", None), ("POST", "scs:cpu-type", "shared-core"), ("PUT", "scs:name-v1", "SCS-1L:1"), ("POST", "scs:name-v2", "SCS-1L-1")]` for example.
* This could then be translated to an `openstack image set` call with the complexity, that there is no syntax to remove single properties, but just one to delete all, so in case there is any `DELETE` operation, we would have to re-add all existing properties, even those that do not change. This make it harder to review for the operator and removes some of the value of this feature ... So we'd probably output the list of changes for review rather than the OpenStack command ...
* The amount of API calls would be reduced, as we'd of course still use the list to consolidate calls. Unlike the CLI tool (which always takes an extra second to startup and get a token), doing API calls from the python SDK is not a performance issue, so this optimization does not seem important. 

I would think we leave this improvement to a v2 for the tool.
While it does make things more beautiful and clearer, it does also add some complexity on its own and I'm unsure whether it's worth it.

garloff:
> I think the following would be good:
> 
>     * collect all flavor names from the flavor (be it the actual name or any name given in the extra_specs) and convert to `Flavorname` instances
> 
>     * check that each of theses instances is consistent with the actual flavor

You mean the settings that can be discovered from OpenStack, right?
Yes, this makes sense.

>     * MAYBE try to check whether these instance are consistent with one another (because they can have varying degree of detail)

In case the flavors have information beyond what OpenStack exposes normally, we can indeed only check for contradictions. Maybe we don't need to aim for perfection here? I would at least check `cpu-type` and disk0-type`. Maybe CPU Architecture.

>     * in case a flavor doesn't have any names in the `extra_specs`: generate one name for v1 of the standard and one for v2 of the standard

Ack.
 
>     * otherwise, check that at least one of the names in the `extra_specs` is valid for a stable version of said standard

I would (by default) generate a set of names:
* Full v1 name (`scs:name-v1`)
* Full v2 name (`scs:name-v2`)
* Shortened v1 name (`scs:name-v3`) IF it's different from the full v1 name.
* Shortened v2 name (`scs:name-v4`) IF it's different from the full v2 name.

So we'd end up with 2 or 4 generated names. For the mandatory and recommended flavors, they will be precisely the same as the ones currently provided by the flavor-manager, which is good, IMVHO.

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
garloff: If you are into functional programming, you could do something like (untested!):
```python
class matcher:
  def __init__(self, lst):
    self.lst = lst
  def flv_match(self, flavor):
    return (self.lst and flavor.name in self.lst) or (not self.lst and flavor.name[0:4] == "SCS-")

def filter_flavors(conn, specified):
  mymatch = matcher(specified)
  return filter(mymatch.flv_match, conn.compute.flavors())
```
More elegant, less readable ...

garloff: Surprisingly, this works. Wrap a `list()` around it for testing ...

mbuechse: Well, I had to objectives:

- pull the if out of the loop
- get by with 2 to 4 lines, because more is not warranted for this simple task

So far, each of your variants has the if inside the loop, and it has 9 to 10 lines ;)

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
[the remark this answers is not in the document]
mbuechse: > there is no syntax to remove single properties, but just one to delete all, so in case there is any DELETE operation, we would have to re-add all existing properties, even those that do not change

I couldn't believe it, so I had to have a look. And what I saw (cannot be unseen). It's such a horrible sh**show -- the whole range: python-openstackclient, openstacksdk, and not least nova, with its api, its objects and its db. This is so much boilerplate, so much bloat, and just so ... mediocre at best.

So we definitely won't do the translation to `python-openstack-client`. Screw that!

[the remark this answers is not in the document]
mbuechse: So just one of the many peculiarities... there is an API endpoint for creating extra specs. But it is functionally equivalent with the update endpoint, which means:

- It won't fail if any extra specs are already present.
- It won't remove any keys at all.

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
```diff
@@ -270,22 +270,29 @@ def main(argv):
         names_to_check.extend(extra_names_to_check)
 
         # syntax check: compute flavorname instances
+        # Also select best match
+        bestln = 0
```
mbuechse, line 274: I definitely agree that we have to be careful here. My suggestion would be as follows:

```python
collection = set()
# in the following, keep version to facilitate sorting afterwards
for flavorname in flavornames.values():
    name_v2 = outputter(flavorname)
    collection.add(("v1", SyntaxV1.from_v2(name_v2)))
    collection.add(("v2", name_v2))
    short_v2 = outputter(flavorname.shorten())
    # could check whether short_v2 != name_v2, but the set will swallow everything
    collection.add(("v1", SyntaxV1.from_v2(short_v2)))
    collection.add(("v2", short_v2))
names = [item[1] for item in sorted(collection, key=lambda item: item[0], -len(item[1]))]
for idx, name in enumerate(names):
    logger.debug(f"set scs:name-v{idx+1} to {name}")
```

mbuechse: In prose: we derive the canonical name strings for each of the flavornames that we have parsed, and we collect those name strings (each one together with the version to facilitate sorting afterwards) in a set, so that we don't see any duplicates. Then we sort by version, then descending by length of the name. This way, we will arrive at our canonical order.

This approach will preserve all names and only make sure that shortened variants are present as well as both versions of the syntax.

mbuechse: This approach doesn't use a reference at all, but we need one in order to derive the other extra_specs (cpu-type and disk0-type). Frankly I still think that these are hard to set IF the names don't agree. In this case, the user has to disambiguate, or we have to mandate some disambiguation rule; for instance: these extra_specs **may not underpromise** (is my opion and also what the standard requires for the standard SCS flavors), and in this case, we would compute the supremum of the flavornames (the highest promise in each category). Then I would also mandate that at least one of the flavornames MUST not underpromise.

garloff: I would agree that the names must not contradict each other. If they do, we may throw an error and let the operator resolve it rather than second guessing what is intended.
But: As understatement is allowed (and in some situations even expected, if we for example expect provider to support nested virtualization `_hwv` but still recommend to not use the complicated flavor name *as primary name for it*, we do need to be prepared that we have several names, some not stating all capabilities.
Your approach is exactly right: We could look at all names and take the max that is being claimed anywhere as reality. (Only exception is where we have hard facts, such as the ram, vcpu and disk fields in the OpenStack nova anwers. Any claim beyond these is of course a lie that should result in a hard error and maybe -- protected by an option -- be fixed.)
My simple implementation was an approximation: I look at the flavor name with the most details, which will be the longest name. It ignores the case where you have several names that each describe some aspects of the flavor in more detail but no name that has it all.
If we find a good implementation (and it looks like you have the skills to do so while I do not without investing significant time) that does build the superset, that would be great. But if we document that this tool only does the right thing for several flavor names if at least one includes all capabilities, that would not be a serious limitation for its usefulness.

garloff: > This approach doesn't use a reference at all, but we need one in order to derive the other extra_specs (cpu-type and disk0-type). 

I like the implementation, thanks!

> Frankly I still think that these are hard to set IF the names don't agree. In this case, the user has to disambiguate, or we have to mandate some disambiguation rule; for instance: these extra_specs **may not underpromise** (is my opion and also what the standard requires for the standard SCS flavors), and in this case, we would compute the supremum of the flavornames (the highest promise in each category). Then I would also mandate that at least one of the flavornames MUST not underpromise.

Thinking into the future: We will at some point stop mandating flavor names, which makes the names (and properties) in extra_specs our primary source of information to select the right flavor in an (SCS-compatible) operator-independent way. So we will have name-vN properties that do underpromise in order to match what users are searching for. Requiring at least one that contains all capabilities is fine and makes sense from my perspective. If it saves us some complexity in the code, even better!

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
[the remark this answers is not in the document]
garloff: Yeah, it's a mess :-(

garloff:
> So we'd end up with 2 or 4 generated names. For the mandatory and recommended flavors, they will be precisely the same as the ones currently provided by the flavor-manager, which is good, IMVHO.

I see you have done all of this in #703, thanks a lot!
This and all of your other improvements have been merged.

mbuechse (approved):
@garloff Maybe you can test once more? Otherwise, this is go for merge.

garloff:
It survives my testing:
* Filling in missing `extra_specs` for SCS- flavors.
* Filling out all `extra_specs` with option `-A` for a non-SCS flavor that has one long `extra_spec` name set

It fails for a flavor that has no `SCS-` name at all (neither in name nor in extra_specs), even if I specify `-p` CPU type and -`t` disk0-type.
```
  File "/casa/src/SovereignCloudStack/standards/Tests/iaas/flavor-naming/flavor_names.py", line 605, in flavorname_to_dict
    collection['cpus'].add(flavorname.cpuram.cpus)
                           ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'function' object has no attribute 'cpus'
```

Let's fix this in a followup merge!

mbuechse:
@garloff There's your followup right there: https://github.com/SovereignCloudStack/standards/pull/706

PR #648, opened 2024-06-26
Add draft for KaaS LB standard (#169)

cah-hbaum:
Follow-Up PR to https://github.com/SovereignCloudStack/standards/pull/169
Everything still relevant will be discussed in here now.

### DIFF
--- a/Standards/scs-XXXX-vN-kaas_load_balancers.md
+++ b/Standards/scs-XXXX-vN-kaas_load_balancers.md
@@ -1,0 +1,81 @@
+---
+title: Behavior of KaaS load balancers
+type: Standard
+status: Draft
+track: KaaS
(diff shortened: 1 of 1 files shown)


joshmue:
I recall that the type of the record was the topic of discussion in #169 - whether this should be a "Decision Record" or "Standard".

I guess that changing the type in the header section to "Standard" would already bring this document mostly in line with other standard documents like #658 - regarding formal aspects as well as value for CSP's/users.

jklare:
Given that there seems to be limited interest of the community to work on this presently, I will close this for now. Once we find somebody who wants to work on this, we can reopen the PR and continue from here. Thank you for the work so far @cah-hbaum, I hope someone will continue it.

Issue #652, opened 2024-07-03
[Standard] A simple, usable default StorageClass should be available in Kubernetes clusters


martinmo:
At the moment, we have [scs-0211-v1](https://docs.scs.community/standards/scs-0211-v1-kaas-default-storage-class) *SCS KaaS default storage class*, which defines some contraints for a default storage class, *if it exists*. But it doesn't enforce that such an SC exists.

Proposal to be discussed:

1. Require that a default storage class exists and is properly annotated (`storageclass.kubernetes.io/is-default-class`).
2. It must support RWO access mode.
3. Users may delete this SC, it just has to exist in a fresh cluster.
4. How this default StorageClass is implemented, is up to the CSP. For example, it could be provided by something simple as [rancher/local-path-provisioner](https://github.com/rancher/local-path-provisioner).
5. Advanced requirements are out of scope, for example:
    - not being backed by local storage
    - support for `allowVolumeExpansion`
    - HA / redundancy
    - performance

Motivation: lower the entry barrier for users by making things work out of the box, wrt storage. For example, many Helm charts require either a default SC or the user needs to specifiy a SC name.

Risks: users may solely rely on this default SC without checking what guarantees it gives.

(CC @mxmxchere because he proposed this to me via chat and I endorse this.)

## Definition of Done:

Please refer to [scs-0001-v1](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0001-v1-sovereign-cloud-standards.md) for details.

- [ ] Proposal has been written with name of the form `scs-xxxx-v1-slug.md` (only substitute `slug`)
- [ ] Proposal has the fields `status`, `type`, `track` set
- [ ] Proposal has been voted upon in the corresponding team
- [ ] Status has been changed into `Draft`, file renamed: `xxxx` replaced by document number
- [ ] If applicable: test script has been written (this item may be moved into a separate issue so long as the state is `Draft`)

martinmo:
I've edited the above text and added some more details and thoughts to be discussed in upcoming calls.

martinmo:
I brought this up in today's Container Call. The original author of the existing standard participated and he endorsed it as well. In the call, I think we reached consensus on the following:

- there must be a default SC
- the requirement of "not being backed by local storage" should remain
  - e.g., remember that in OpenStack, local storage is ephemeral
  - but users expect persistence, after all it's called PVC and (not EVC)
- there was a quick discussion whether this is too strict from a standards point of view
[...]

PR #658, opened 2024-07-16
[Standard] KaaS default StorageClass v2 (scs-0211-v2)

martinmo:
Resolves: #652


martinmo:
@joshmue can you please have a look at this. @mbuechse kindly agreed to have a look at the formal aspects.

My plan is to proceed with the conformance tests once we have merged this v2 as draft. And then, when conformance tests are ready & merged, we can proceed to stabilize it.

Standards/scs-0211-v2-kaas-default-storage-class.md
```diff
@@ -0,0 +1,48 @@
...
+preconfigured with a sensible default storage class providing persistent storage.
+
+## Decision
+
+A freshly provisioned Kubernetes cluster MUST have a default storage class, i.e., a `StorageClass`
+object that is annotated with `storageclass.kubernetes.io/is-default-class=true` as described in the
+[Kubernetes documentation][k8s-default-sc].
+The name of this storage class is not standardized.
+
+Users MUST be able to change the default storage class.
```
joshmue, line 31: While Kubernetes itself does not prevent that, this is not necessarily easily doable in all Kubernetes distributions. For example, Gardener does not seem to support it: https://github.com/gardener/gardener/issues/1225

The value provided by such feature is also pretty slim, as users can always just reference their own storage class.

Hence, I guess it should be dropped:

```suggestion
```

joshmue (commented):
LGTM, otherwise.

Standards/scs-0211-v2-kaas-default-storage-class.md
martinmo: I agree, thanks for the hint!

Standards/scs-0211-v2-kaas-default-storage-class.md
```diff
@@ -0,0 +1,46 @@
...
+
+- MUST support the `ReadWriteOnce` [access mode][k8s-accessmode].
+- MUST NOT be backed by local or ephemeral storage.
+- MUST NOT be bound to the lifecycle of a Kubernetes node.
+
+The provisioned storage class MAY support volume expansion (`allowVolumeExpansion=true`).
+
+## Conformance Tests
+
+TBD
```
martinmo, line 42: Btw, I just noticed, while working through #643, that we already have a conformance test (for v1) and this TBD was never updated in the v1 standard.

joshmue (approved):

mbuechse (commented):
Please add something about previous versions, as in https://docs.scs.community/standards/scs-0100-v3-flavor-naming
-- first, one sentence in the intro about this version, and then one section further below about previous versions and what has changed compared to those.

Standards/scs-0211-v2-kaas-default-storage-class.md
mbuechse: If you haven't already, can you please open an issue (or, if you're quick, a PR) about this?

martinmo: https://github.com/SovereignCloudStack/standards/pull/661, but unfortunately, while doing that, also I found another inconsistency, see #662

Standards/scs-0211-v2-kaas-default-storage-class.md
```diff
@@ -0,0 +1,59 @@
...
+## Previous standard versions
+
+[Version v1 of this standard](scs-0211-v1-kaas-default-storage-class.md) did not enforce the
+existence of a default storage class in a newly created cluster.
+
+Previously, the backing storage of the default storage class was required to be protected
+against data loss caused by a physical disk or host failure.
+It also contained recommendations (MAY) with regard to redundant storage across hosts
+or availability zones.
+In this revision, these requirements and recommendations have been dropped.
```
joshmue, line 51: This clarification actually confused me a bit, as I was under the impression that protection "against data loss caused by a physical disk or host failure" was intended to be covered by "MUST NOT be backed by local or ephemeral storage".

Regarding edge setups, you concluded in #652:

> - the requirement of "not being backed by local storage" should remain
> ...
> - there was a quick discussion whether this is too strict from a standards point of view
>   - example: edge clouds probably cannot fulfill this -> but this a very special case and might be solved by another hypothetical certificate scope for edge environments

There are not that many options (or use cases) to implement storage that is NOT being backed by local storage, yet NOT being protected against data loss due to disk/host failure.
So this concession to CSP's will probably create little value for CSP's, yet would reduce the value for users who may want to rely on SCS standards to be "sensible/common defaults". I do not know of any major block storage cloud offering that does not provide basic redundancy or at least snapshotting.

martinmo: > This clarification actually confused me a bit, as I was under the impression that protection "against data loss caused by a physical disk or host failure" was intended to be covered by "MUST NOT be backed by local or ephemeral storage".

I removed the "physical disk or host failure" sentence because we will not be able to test this as part of the conformance checks anyway.

My reasoning behind this is: even if you use the Cinder CSI, or let's say, the Rook CSI provisioner, the backing storage could in theory be misconfigured wrt to single host or disk failure tolerance, but there is, AFAIK, no reliable way to check this from inside the K8s cluster (and with minimal privileges).

Of course, our expectation would be that this failure tolerance is configured correctly by the CSP. As a compromise I can add a sentence that documents this expectation.

In my understanding, the important bit of this standard is that a provisioned volume isn't bound to the node's lifecycle, i.e., it doesn't just use something like the `rancher.io/local-path` or the `lvm.csi.metal-stack.io` provisioner.

For the test, I currently plan to check against an allowlist of provisioners (e.g., `cinder.csi.openstack.org`, `*.rbd.csi.ceph.com`, `driver.longhorn.io`, may need extension in the future), because I think it is the most pragmatic approach. (Going one step further, the test could even try to mount the same PVC on two different nodes.)

> So this concession to CSP's will probably create little value for CSP's, yet would reduce the value for users who may want to rely on SCS standards to be "sensible/common defaults". I do not know of any major block storage cloud offering that does not provide basic redundancy or at least snapshotting.

I do not agree completely: there is a value for the user if we prevent that a "simple" provisioner is used (like local-path), e.g., volume is independent of the node, can be used on another node, etc.

joshmue: Regarding testability: Alright, did not know that this that much of an important factor. :+1: 

---

> I do not agree completely: there is a value for the user if we prevent that a "simple" provisioner is used (like local-path), e.g., volume is independent of the node, can be used on another node, etc.

Definitely! Yet, it's kind of a niche use case.

In my experience, one would want to have either...

- some *failure-safe* volume for any kind of application. Maybe a trivial non-HA database installation for non-HA use cases. While availability may be of little concern, data durability often will be, making a storage backend with some hands-off redundancy very much favorable. This is the default on most clouds, which will also work with most use cases involving applications implementing data high availability/durability themselves.  
OR
- some *not necessarily failure-safe* volume (may be local/node-bound/non-redundant, possibly fast) to run applications that take care of data durability and availability on application level.

While offering a default storage class which is not redundant/failure-safe/replicated in any way, only non-local, would make e. g. "A lot of third-party software, such as Helm charts, [which] assume that a default storage class is configured" technically initially installable, it would probably also break assumptions of these charts regarding data durability, undermining this standard's motivation for an "out-of-the-box working experience" (at least some time later if/when a disk eventually fails).

That may be the outcome (yet I do not know of CSP's who are really restrained here), but no matter what, IMHO it should be spelled out more explicitly also in the "Decision" section.

fraugabel:
added the omitted requirements as recommendations under Decisions

fraugabel (approved):
necessary requirements for fail-safe data carriers have been added as recommendations, as there may be use cases that allow these to be ignored. In general, these recommendations can be adhered to and checked by selecting the right provisioner.

PR #677, opened 2024-07-31
introduce scs-0103-v1 v1.1 (backwards compatible, compliance-preserving)

### LINKED ISSUE #594: Relax scs-0103-v1, introducing scs-0103-v2

mbuechse:
resolves #594
alternative to PR #630


Standards/scs-0103-v1-standard-flavors.md
```diff
@@ -43,6 +48,21 @@ The following extra specs are recognized, together with the respective semantics
 
 Whenever ANY of these are present on ANY flavor, the corresponding semantics must be satisfied.
 
+The property `scs:name-vN` is to be interpreted as "name variant N". This name scheme is designed to be
```
artificial-intelligence, line 51:
```suggestion
The extra spec `scs:name-vN` is to be interpreted as "name variant N". This name scheme is designed to be
```

Standards/scs-0103-v1-standard-flavors.md
```diff
@@ -127,14 +147,19 @@ instance life cycle.)
...
-and check whether the flavors are present and their extra specs are correct. Missing
-flavors will be reported on various logging channels: error for mandatory, info for
+and check whether the flavors are present and their extra specs are correct.
+
+Missing flavors will be reported on various logging channels: error for mandatory, warning for
 recommended flavors. Incorrect extra specs will be reported as error in any case.
 The return code will be non-zero if the test could not be performed or if any error was
 reported.
 
+The script does not check whether a name given via `scs:name-vN` is indeed valid according
```
artificial-intelligence, line 160:
```suggestion
The script does not check whether a name given via the extra spec `scs:name-vN` is indeed valid according
```

artificial-intelligence (changes requested):
Maybe we should also add that the tests don't need to be changed for this?

artificial-intelligence (approved):
Thanks for the terminology links, these are super useful imho, because the meaning of extra_specs is very nuanced and different, depending on how you look at it.

LGTM

Standards/scs-0103-v1-standard-flavors.md
```diff
@@ -14,6 +14,19 @@ description: |
...
+now interpreted as "name variant N" instead of "version N of the naming standard". Note that this change
+indeed preserves compliance, i.e., compliance with v1.0 implies compliance with v1.1.
+
+## Terminology
+
+extra_specs
+  Additional properties on an OpenStack flavor, see
+  https://docs.openstack.org/nova/2024.1/user/flavors.html#extra-specs
+  and
+  https://docs.openstack.org/nova/2024.1/configuration/extra-specs.html
```
artificial-intelligence, line 28:
```suggestion
  <https://docs.openstack.org/nova/2024.1/user/flavors.html#extra-specs>
  and
  <https://docs.openstack.org/nova/2024.1/configuration/extra-specs.html>
```

should suffice to please the markdown linter gods

mbuechse: I think this linter is outdated. The newer Docusaurus prefers bare links.

mbuechse: @maxwolfs can you please comment on this?

mbuechse: I have removed the bare links by adding descriptions.

PR #678, opened 2024-08-01
Define maintenance severity behavior

joshuai96:
Ref: SovereignCloudStack/issues#682

### DIFF
--- a/Standards/scs-0402-v1-status-page-openapi-spec-decision.md
+++ b/Standards/scs-0402-v1-status-page-openapi-spec-decision.md
@@ -135,7 +135,7 @@ A severity with the value of 100 MUST always be supplied. This is the highest se
 
 Components list their impacts, which they are affected by, as read only. Only an incident creates an impact on a component. Components MUST only list their currently active impacts.
 
-An optional `now` parameter can be supplied, to set a reference time to show all incidents, active at that time, even when they are inactive currently.
+An optional `at` parameter can be supplied, to set a reference time to show all incidents, active at that time, even when they are inactive currently.
 
 ### Maintenance
 
--- a/Standards/scs-0402-v1-status-page-openapi-spec-decision.md
+++ b/Standards/scs-0402-v1-status-page-openapi-spec-decision.md
@@ -64,6 +64,12 @@ Requests to updating operations SHOULD contain the minimum of the changed fields
 
 The endpoints are named in plural form, even when handling single objects, to keep uniform paths.
 
+### Incidents
+
+Incidents are the main information bearer at the status page. They hold data to describe an incident, that is happening or has happened, all stages the incident progresses through and all updates related to the resolution of the incident.
+
+An incident is considered active, while no endpoint was set.
+
 ### Phase list
 
 The list of phases that an incident can go through has a crucial order. So it MUST be handled as the given list.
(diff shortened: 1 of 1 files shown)


Standards/scs-0402-v1-status-page-openapi-spec-decision.md
```diff
@@ -137,6 +135,12 @@ A severity with the value of 100 MUST always be supplied. This is the highest se
 
 Components list their impacts, which they are affected by, as read only. Only an incident creates an impact on a component. Components MUST only list their currently active impacts.
 
+An optional `now` parameter can be supplied, to set a reference time to show all incidents, active at that time, even when they are inactive currently.
```
snkaupe, line 138: Wrong comma placement.

snkaupe:
```suggestion
An optional `now` parameter can be supplied, to set a reference time to show all incidents, active at that time, even when they are inactive currently.
```

snkaupe: Have the docs ever defined what an incident being "active" means? If not, we should add this definition, just to make it abundantly clear.

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
```diff
@@ -137,6 +135,12 @@ A severity with the value of 100 MUST always be supplied. This is the highest se
...
+
+### Maintenance
+
+An `impact`, having the special `SeverityValue` of 0 indicates a maintenance time slot, as such it MUST include a start and end time.
```
snkaupe, line 142: Comma placement. I'd suggest rephrasing to get rid of it.

snkaupe:
```suggestion
Any `impact` that has the reserved `SeverityValue` of 0 is a maintenance time slot. As such it MUST include a start and end time. However, both are allowed to be set in the future.
```

snkaupe (changes requested):
A fine PR with just a few changes to request.

Standards/scs-0402-v1-status-page-openapi-spec-decision.md
joshmue, line 138: Maybe change the `now` bool parameter to be a timestamp named `at`? This will enable the same use case (and more), while keeping server responses stable, unaffected by system time changes.

joshuai96: No, thats true. 

An active incident is just an incident without a `ended_at` set.

snkaupe: It is my understanding that this is already the intended behaviour. The name `at` is, however, a marked improvement in my eyes over `now`.

joshuai96: changed

joshuai96: Mentioned the incidents and their end condition.

joshmue (approved):

snkaupe (changes requested):
Two little further changes and we should be golden.

I added *resolved* as an alternative to *inactive* concerning incidents due to this apparently being a common way to talk about past incidents. If that is an unwanted change, please feel free to remove it.

snkaupe (approved):
Looks good to me!

Issue #672, opened 2024-07-26
[EPIC] Migrate official compliance tests to Zuul


mbuechse:
PREREQUISITES:

- [x] #660
- [x] #715 
- [x] #716  
- [x] #664 
- [x] #665 
- [x] #738
- [x] #750 

OPTIONAL:

- [x] #671
- [x] #666 
- [ ] add notification mechanism to compliance monitor (or poll by some other service)

TASKS:

- [x] https://github.com/SovereignCloudStack/docs/issues/271
- [ ] mark GitHub actions as obsolete
- [x] make sure templates in compliance monitor are up to date
- [x] include generated table into docs page
- [x] include link to generated table into README.md
- [x] finally close https://github.com/SovereignCloudStack/docs/issues/166

martinmo:
Motivated by https://github.com/SovereignCloudStack/standards/pull/680#issuecomment-2268320692, are parallel checks against the same cloud coordinated in Zuul, so they don't interfere with each other?

berendt:
@martinmo We have worked with semaphores on jobs that basically share the same project and have overlapping names. This prevents jobs from running in parallel on a shared project.

Semaphore definition: https://github.com/regiocloud/e2e-tests/blob/main/.zuul.yaml#L82-L84
Semaphore use: https://github.com/regiocloud/e2e-tests/blob/main/.zuul.yaml#L107-L108

martinmo:
@berendt Thanks for the hint! It seems we do not currently use this here, at least a quick `rg --hidden semaphore` through the standards repo didn't reveal anything like that.

mbuechse:
I don't think we have the case right now (the nightly jobs are not run in parallel as far as I can tell), but it would be good to be on the safe side nevertheless.

mbuechse:
Discussed in today's (extraordinary) meeting of SIG Std/Cert: most new features are actually not needed for the switch and can be postponed.

PR #680, opened 2024-08-02
Run pre-check cleanup also in GH actions

martinmo:
Even if we want to completely switch to Zuul, it's not clear when this happens and in the meantime there is still plenty of opportunity that this is the reason for failed tests. The fix is rather easy, I just copied and slightly adapted the invocation from `playbooks/pre_cloud.yaml`.

### DIFF
--- a/.github/workflows/scs-compliance-check-with-application-credential.yml
+++ b/.github/workflows/scs-compliance-check-with-application-credential.yml
@@ -32,6 +32,8 @@ jobs:
               auth:
                 application_credential_secret: ${{ secrets[inputs.secret_name] }}
           EOF
+      - name: "Clean up any lingering resources from previous run"
+        run: "cd /scs-compliance && ./cleanup.py -c ${{ inputs.cloud }} --prefix _scs-"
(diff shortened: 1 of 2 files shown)


mbuechse (approved):
Thanks for taking the initiative!

martinmo:
Please note: I'll merge this on monday, because I fear that this might break something and I don't want to risk this right before the weekend :-)

martinmo:
After having another look, I came to the conclusion that this can't work reliably: the workflows for one cloud can run in parallel, sometimes 4 at the same time (and not only because of the same schedule, they also run after the Docker image was built). Hence, with good timing, the cleanup will interfere with a run. This can be worked around by running the cleanup only for IaaS v4 checks (it is only needed there).

mbuechse:
I guess it would have been better in principle to define a separate cleanup job and make it a prerequisite of the compliance-test job. Since we want to abandon the Github jobs anyway (#672), we may opt to save us this effort.

martinmo:
I agree with both statements, this is why I just used the quick workaround :+1:

PR #682, opened 2024-08-02
Add Ubuntu 24.04 as mandatory image, drop mandatory for 22.04

martinmo:


Tests/iaas/scs-0104-v1.1-images.yaml
```diff
@@ -0,0 +1,30 @@
...
+  status: mandatory
+- name: "Ubuntu 22.04"
+  source:
+  - https://cloud-images.ubuntu.com/releases/jammy/
+  - https://cloud-images.ubuntu.com/jammy/
+- name: "ubuntu-capi-image"
+  name_scheme: "ubuntu-capi-image v[0-9]\\.[0-9]+(\\.[0-9]+)?"
+  source: https://swift.services.a.regiocloud.tech/swift/v1/AUTH_b182637428444b9aa302bb8d5a5a418c/openstack-k8s-capi-images/ubuntu-2204-kube
+  status: recommended
+- name: "Ubuntu 20.04"
```
berendt, line 15: While touching this list, should we remove this?

martinmo: @berendt I tend to to agree. However: the status quo of the standard and check is as follows: without a `status`, the conformance check will not complain at all if this image is missing. However, it will complain if it is available but from the wrong source.

What we probably want, and what we discussed internally: there needs to be a way phase out old OS images, i.e., new `status` values in the YAML. @mbuechse proposed `discouraged` and `forbidden` (the counterparts for `recommended` and `mandatory`). It needs to be discussed, whether such an image should be removed completely, or visibility changed to `community` or something.

martinmo: To be discussed as part #684

mbuechse: Two questions here:

1. does it hurt to specify the preferred source url for a distro that is EOL? Maybe not.
2. shall we discourage CSPs from providing these EOL distros? Maybe yes.

So yeah, in the end, maybe keep them in the file, but go ahead with #684 and mark them "discouraged".

Tests/iaas/scs-0104-v1.1-images.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - https://cloud.debian.org/images/cloud/bookworm/
+  - https://cdimage.debian.org/cdimage/cloud/bookworm/
+- name: "Debian 11"
+  source:
+  - https://cloud.debian.org/images/cloud/bullseye/
+  - https://cdimage.debian.org/cdimage/cloud/bullseye/
+- name: "Debian 10"
+  source:
+  - https://cloud.debian.org/images/cloud/buster/
+  - https://cdimage.debian.org/cdimage/cloud/buster/
```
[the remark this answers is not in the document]
mbuechse: I think it can be problematic if an environment has Debian 10, but NOT 11 or 12, as seems to be the case with Syseleven, see #696. Then users have a hard time choosing a current version. They can always upload one, I guess, but that is an additional barrier.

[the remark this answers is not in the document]
garloff: I would expect Debian 12 to be mandatory at this point, same for 11 and 10 recommended.

The whole point of requiring a minimal set of public images is to ensure that customers do not have to upload their own stock images and remember to update them once per month ...
This is certainly a good answer to folks that want specific images. But mainstream images are better provided by the cloud operator, as this is more efficient for everyone.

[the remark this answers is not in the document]
mbuechse: Then we shall do that for v5 of SCS-compatible IaaS

mbuechse (commented):
Make the latest Ubuntu LTS mandatory. Make Debian 12 recommended. Find a better filename. Keep v1 at the front, maybe add v5 to the back, as a reference to the certificate scope version?

mbuechse (approved):
as per SIG Std/Cert of 2024-08-22

PR #668, opened 2024-07-23
Increase granularity of badges or certification

mbuechse:
Major revision of certificate-scope yaml
example scs-compatible-iaas.yaml, scs-compatible-kaas.yaml and related tests
including scs-0003-v1 and adaptation of compliance monitor

resolves #660 
also resolves #591

PR #683, opened 2024-08-02
Update the conformance tests readme and Dockerfile

martinmo:
The instructions were completely outdated. Also having a default command in the `Dockerfile` doesn't make sense anymore, because so many parameters have to be passed, so I switched the default to showing the help output.

### DIFF
--- a/Tests/README.md
+++ b/Tests/README.md
@@ -16,30 +16,51 @@ python3 -m venv .venv && source .venv/bin/activate
 pip install -r requirements.txt
 ```
 
+At the moment, there are two cloud layers that can be checked: IaaS and KaaS.
+
+### IaaS checks
+
 With a cloud environment configured in your `~/.config/openstack/clouds.yaml`
 and `secure.yaml`, then run
 
 ```shell
-./scs-compliance-check.py scs-compatible.yaml iaas --os-cloud CLOUDNAME
+./scs-compliance-check.py  -a os_cloud=CLOUDNAME -s CLOUDNAME scs-compatible-iaas.yaml
 ```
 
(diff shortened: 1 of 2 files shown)


mbuechse:
This is already part of #668, albeit less good. I would've preferred a contribution to that PR to save  me the trouble when rebasing it later.

Tests/README.md
```diff
@@ -16,30 +16,51 @@ python3 -m venv .venv && source .venv/bin/activate
...
+```shell
+./scs-compliance-check.py -v -a kubeconfig=path/to/kubeconfig.yaml -s SUBJECT scs-compatible-kaas.yaml
+```
+
+Replace `SUBJECT` with an arbitrary, but meaningful subject name.
+Also, please note that the check will always use the `current-context` of the kubeconfig and will
+fail if it isn't set.
+
+For both types of checks, the exit code indicates success (0) or failure (!= 0).
 You can also request a YAML report using the option `-o OUTPUT.yaml`
```
mbuechse, line 46: this should be moved befre the case distinction, otherwise it may be missed by people who aren't interested in the KaaS casr; maybe introduce a general commandline first (with ellipses for the -a and the spec file)

mbuechse (approved):
I like it, thanks! As I said, I would've preferred to have that as part of #668, where I had already started something similar, but still. Please address my comment and then fire at will.

martinmo:
> This is already part of #668, albeit less good than here. I would've preferred a contribution to that PR to save me the trouble when rebasing it later.

Sorry, that was an oversight on my part. I'll assist in merging the changes from this PR into your PR, of course :-)

PR #685, opened 2024-08-05
Perform scs-check-adr-syntax job in a pod

gtema:
In order to improve speed of the job (reduce wait time for the VM to be
prepared) we can execute it in the K8 pod.

In addition to that get rid of double data copying (what also consumes noticeable amount of time) for the job in question. This is caused by a fact that the base Zuul job takes care of pushing the repository content to the target environment in the most optimal way. Other jobs would profit from similar change independently.

Both changes together bring job duration from very varying 2-5 minutes (excluding waiting for the job to start) down to  stable 40 sec (with 0-10 seconds job start delay).

Signed-off-by: Artem Goncharov <artem.goncharov@gmail.com>

### DIFF
--- a/playbooks/pre.yaml
+++ b/playbooks/pre.yaml
@@ -4,20 +4,7 @@
   roles:
     - role: ensure-pip  # https://zuul-ci.org/docs/zuul-jobs/latest/python-roles.html#role-ensure-pip
   tasks:
-    - name: Copy ADRs on the node
-      ansible.builtin.copy:
-        src: "../Standards"
-        dest: "~/"
-        mode: 0500
-      no_log: false
-
-    - name: Copy Tests on the node
-      ansible.builtin.copy:
-        src: "../Tests"
-        dest: "~/"
-        mode: 0500
-      no_log: false
-
     - name: Install dependencies
       ansible.builtin.pip:
-        requirements: ~/Tests/requirements.txt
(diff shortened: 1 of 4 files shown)


gtema:
recheck

playbooks/pre.yaml
```diff
@@ -4,20 +4,20 @@
   roles:
     - role: ensure-pip  # https://zuul-ci.org/docs/zuul-jobs/latest/python-roles.html#role-ensure-pip
   tasks:
-    - name: Copy ADRs on the node
-      ansible.builtin.copy:
-        src: "../Standards"
-        dest: "~/"
-        mode: 0500
-      no_log: false
+    # - name: Copy ADRs on the node
```
mbuechse, line 7: I don't quite understand this change, but more importantly, I'm quite sure that it will break the other jobs. I get that this PR is a draft, just my 2 cents

gtema: I know, it was an interim step and I kept "current" state as pre-old.yaml with the intention to do similar improvement for other job separately.

mbuechse (commented):
If I get this rightt, there are two problems:

- we do unnecessary copying
- we could use pods

I would like to address those separately, BUT each for all jobs at once (the number of jobs has massively decreased in the meantime).

mbuechse:
Many thanks for this valuable inspiration. I hope I got it right when I tried to apply it in full with #750. Superseded by #750.

PR #689, opened 2024-08-05
Bugfix: gracefully handle incorrectly set min_disk property

mbuechse:
Remedy for
https://github.com/SovereignCloudStack/standards/actions/runs/10252209326/job/28362074808#step:6:51
and
https://github.com/SovereignCloudStack/standards/actions/runs/10252209272/job/28362074519#step:6:51


martinmo (approved):
No complaints. There is only a follow-up: in scs-0102-v1, the property `min_disk_size` (vs. `min_disk`) is considered mandatory (without value constraints) and probably there's a check missing.

PR #690, opened 2024-08-06
Fix typo: use min_disk property instead of min_disk_size

martinmo:
see #689


martinmo:
For the record, in `image-md-check.py`:

https://github.com/SovereignCloudStack/standards/blob/c580700238b2131c64093efcfd383b6b3bd48a75/Tests/iaas/image-metadata/image-md-check.py#L275-L283

Turning these warnings into errors would be nice, but probably there are some reasons or edge cases I don't know about that prevent us from doing this.

mbuechse (approved):
According to decision of [Team IaaS](https://github.com/SovereignCloudStack/minutes/blob/main/iaas/20240807.md#image-metadata-standard-typo-or-grave-mistake-mbuechsemartinmo). Please include into squash commit message.

PR #693, opened 2024-08-13
hotfix: avoid buggy(?) openstack-sdk version

mbuechse:


martinmo (approved):

mbuechse:
Just one remark: I spent a considerable amount of time trying to find out why any version from 3.2.0 onwards is breaking things. I studied the openstack-sdk code, particularly the diff between 3.1.0 and 3.2.0, but I did not succeed! I suppose it would be better to stay away from `openstack.cloud` stuff (we don't include that module, but it provides all the mix-ins to the openstack connection class, and there we use it). For instance, instead of `conn.create_server` we might say `conn.compute.create_server` plus a few other calls for the public IP etc.

mbuechse:
Okay, I did find the problem after all. The culprit is this:

```text
commit a8adbadf0c4cdf1539019177fb1be08e04d98e82
Author: Stephen Finucane <stephenfin@redhat.com>
Date:   Thu Aug 17 12:26:33 2023 +0100

    cloud: Do not reference other cloud mixin's methods
    
    We are using a mixin-like pattern to construct the cloud layer of our
    API. However, this is not a true mixin pattern since each of the
    "mixins" have dependencies on each other and to be honest, it's really
    only there to allow us to split up the otherwise giant 'Connection'
    object. Unfortunately this pattern doesn't really work with type
    checkers or language servers since there's no obvious way to identify
    where the additional resources are coming from until they're combined in
    the 'Connection' class. Fixing this will require multiple steps, but the
    first it to remove all dependencies between the various service mixins.
    As such, rather than depending on e.g. the 'get_volume' cloud method in
    various compute-related cloud methods, we use the 'find_volume' proxy
    method. A later change can then add these various proxy attributes,
    exposing them.
    
    Change-Id: I10d3782899ac519f715d771d83303990a8289f04
    Signed-off-by: Stephen Finucane <stephenfin@redhat.com>
```

particularly, with these changes in `openstack/compute/_floating_ip.py`:

```diff

```diff
@@ -680,7 +678,7 @@ class FloatingIPCloudMixin:
             # old to check whether it belongs to us now, thus refresh
             # the server data and try again. There are some clouds, which
             # explicitely forbids FIP assign call if it is already assigned.
-            server = self.get_server_by_id(server['id'])
+            server = self.compute.get_server(server['id'])
             ext_ip = meta.get_server_ip(
                 server, ext_tag='floating', public=True
             )
@@ -719,7 +717,7 @@ class FloatingIPCloudMixin:
                 "Timeout waiting for the floating IP to be attached.",
                 wait=min(5, timeout),
             ):
-                server = self.get_server_by_id(server_id)
+                server = self.compute.get_server(server_id)
                 ext_ip = meta.get_server_ip(
                     server, ext_tag='floating', public=True
                 )
```

The old call did include a call to `add_server_interfaces` from `openstack/cloud/meta.py`, which the new version doesn't. And that's the whole deal. Let it be known that I personally abhor Mixins to begin with, much like metaclasses and so on. It is just too clever for one's own good.

PR #697, opened 2024-08-15
find recent Debian image even if only codenames are given

### LINKED ISSUE #696: [BUG] entropy-check taking hours
This happens on the syseleven environments. The test is terminated by Github after 6 hrs when it reaches the maximum runtime...

mbuechse:
resolves #696

### DIFF
--- a/Tests/iaas/entropy/entropy-check.py
+++ b/Tests/iaas/entropy/entropy-check.py
@@ -15,7 +15,6 @@ restrictions); for further information, see the log messages on various channels
 from collections import Counter
 import getopt
 import logging
-from operator import attrgetter
 import os
 import re
 import sys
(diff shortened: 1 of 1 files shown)


gtema (approved):
Great would be to have a unittest for the function. I tried to iterate in the head with few common values and noticed that our famous "not_really_required" case of dashes or underscores instead of spaces will not work, "ubuntu22.04" neither. Not critical, but not sure about your general intention of the usage.

mbuechse:
> Great would be to have a unittest for the function. I tried to iterate in the head with few common values and noticed that our famous "not_really_required" case of dashes or underscores instead of spaces will not work, "ubuntu22.04" neither. Not critical, but not sure about your general intention of the usage.

In this case, just make the test work better with the syseleven environments. But you are right, we need some better way of finding a certain image! I think this calls for better discoverability or at least more/better standardized properties?

gtema:
Yes, there are standard properties os_distro and os_version We should make filling of those a standart and not the name convention

mbuechse:
> Yes, there are standard properties `os_distro` and `os_version` We should make filling of those a standart and not the name convention

My bad! I have to say, I must have been blind. I should of course use those properties. I will make a new PR. Stupid me.

PR #698, opened 2024-08-15
bugfix: use os_distro and os_version to determine distro and version …

mbuechse:
…(d'oh)

thanks to @gtema for spotting this

### DIFF
--- a/Tests/iaas/entropy/entropy-check.py
+++ b/Tests/iaas/entropy/entropy-check.py
@@ -453,8 +453,6 @@ def _deduce_version(img):
     return d_idx, deducer("")
 
 
-
-
 def select_deb_image(images):
     """From a list of OpenStack image objects, select a recent Debian derivative."""
     return max(images, key=_deduce_version, default=None)
--- a/Tests/iaas/entropy/entropy-check.py
+++ b/Tests/iaas/entropy/entropy-check.py
@@ -402,38 +402,62 @@ class CountingHandler(logging.Handler):
         self.bylevel[record.levelno] += 1
 
 
-def _deduce_version(name, ubuntu_ver=re.compile(r"\d\d\.\d\d\Z"), debian_ver=re.compile(r"\d+\Z")):
-    """helper for `select_deb_image` to deduce a version even if its only given via codename"""
-    canonicalized = [part.strip() for part in name.lower().split()]
-    if "debian" in canonicalized:
-        # don't even consider "stretch" (9) here
-        codenames = ("buster", "bullseye", "bookworm")
-        for idx, name in enumerate(codenames):
-            if name in canonicalized:
-                return idx + 10
-        for part in canonicalized:
-            if debian_ver.match(part):
-                return int(part)
-    elif "ubuntu" in canonicalized:
-        for part in canonicalized:
-            if ubuntu_ver.match(part):
-                return int(part[:2] + part[3:])
-    return -1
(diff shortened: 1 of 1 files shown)


Tests/iaas/entropy/entropy-check.py
```diff
@@ -402,38 +402,60 @@ def handle(self, record):
...
+    name = name.lower().strip()
+    if debian_ver.match(name):
+        return int(name)
+    return DEBIAN_CODENAMES.get(name, 0)
+
+
+def _deduce_ubu_version(name, ubuntu_ver=re.compile(r"\d\d\.\d\d\Z")):
+    name = name.lower().strip()
+    if ubuntu_ver.match(name):
+        return int(name[:2] + name[3:])
```
gtema, line 423: What about just removing the dot from the string instead of cutting and rejoining it? I makes it pretty hard to read

Tests/iaas/entropy/entropy-check.py
```diff
@@ -402,38 +402,60 @@ def handle(self, record):
...
+        d_idx, deducer = DISTROS.get(img.os_distro.lower(), DISTROS[None])
+        return d_idx, deducer(img.os_version)
+    # os_distro not set; this could mean that the image in question is not a plain OS image.
+    # Try to parse the name, assuming that distro comes first so we don't need to scan twice
+    logger.debug(f"img {img.name!r} missing os_distro or os_version")
+    canonicalized = [part.strip() for part in img.name.lower().split()]
+    d_idx, deducer = DISTROS[None]
+    for part in canonicalized:
+        if d_idx == 0:
+            d_idx, deducer = DISTROS.get(part, DISTROS[None])
```
gtema, line 447: uhm, I think now _you_ are "requiring doctor title" from the reader ;-)

mbuechse: Hm.. on second thought

```python
             d_idx, deducer = DISTROS.get(part) or DISTROS[None]
```

would also work. Better?

Tests/iaas/entropy/entropy-check.py
mbuechse: f6ebaebcad4a0ee88636c79c09ed8ce8fb85c731

Tests/iaas/entropy/entropy-check.py
mbuechse: f6ebaebcad4a0ee88636c79c09ed8ce8fb85c731

mbuechse:
@gtema with my latest round of simplifications, it should be merge-able, no?

Tests/iaas/entropy/entropy-check.py
```diff
@@ -402,38 +402,52 @@ def handle(self, record):
...
+
+# only list stable versions here
+DEBIAN_CODENAMES = {
+    "buster": 10,
+    "bullseye": 11,
+    "bookworm": 12,
+}
+
+
+def _deduce_sort_debian(name, debian_ver=re.compile(r"\d+\Z")):
```
martinmo, line 418: I was slightly confused that the argument is called `name` instead of `version`, because in line 445 you pass `img.os_version` to it the deducer functions and the matching also seems to be based on names.

mbuechse: c2007be330686339b58e1a5be165ec4956418a0b

martinmo (approved):

PR #699, opened 2024-08-15
use the same signing key for all clouds on Zuul, use but one Zuul job

### LINKED ISSUE #664: Compliance Monitor: use the same ssh key for all test subjects that run on our Zuul instance
At present, we have one key per subject, but the Zuul pipeline is signing the result on behalf of the subject's owner, and so it wouldn't be wrong to use just one, Zuul-specific key whenever we sign a result on Zuul. It's already possible to put the same key for each subject, but I could also imagine slightly changing the data structure so that the key can be shared more explicitly.

### LINKED ISSUE #665: Compliance Monitor: test all subjects with but one Zuul job

mbuechse:
resolves #664
resolves #665

### DIFF
--- a/playbooks/compliance_check.yaml
+++ b/playbooks/compliance_check.yaml
@@ -4,35 +4,32 @@
   tasks:
     - name: Run compliance script
       ansible.builtin.shell: |
-        python3 ~/Tests/scs-compliance-check.py ~/Tests/scs-compatible-iaas.yaml -s {{ cloud }} -a os_cloud={{ cloud }} -o report.yaml -C
-      register: result
+        python3 ~/Tests/scs-compliance-check.py ~/Tests/scs-compatible-iaas.yaml -s {{ item }} -a os_cloud={{ item }} -o report-{{ item }}.yaml -C
+        ssh-keygen -Y sign -f ~/id_subject -n report report-{{ item }}.yaml
+        curl --data-binary @report-{{ item }}.yaml.sig --data-binary @report-{{ item }}.yaml -H "Content-Type: application/x-signed-yaml" -H "Authorization: Basic $auth" https://compliance.sovereignit.cloud/reports
       changed_when: true
       failed_when: false
       # ^^^ this task does not fail; the only failure would be if `report.yaml` didn't get produced,
       # but then the subsequent tasks would fail anyway
-
-    - ansible.builtin.debug:
-        msg: "{{ result.stdout }} {{ result.stderr }}"
-
-    - name: sign result YAML and upload to compliance monitor
-      no_log: true  # do not leak the secret in the shell part below
-      ansible.builtin.shell: |
-        ssh-keygen -Y sign -f ~/id_subject -n report report.yaml
-        curl --data-binary @report.yaml.sig --data-binary @report.yaml -H "Content-Type: application/x-signed-yaml" -H "Authorization: Basic {{ clouds_conf.zuul_ci_basic_auth }}" https://compliance.sovereignit.cloud/reports
+      environment:
+        auth: "{{ clouds_conf.zuul_ci_basic_auth }}"
+      loop: {{ clouds }}
 
     - name: Copy result YAML
       ansible.builtin.synchronize:
-        dest: "{{ zuul.executor.log_root }}/{{ cloud }}-iaas.yaml"
+        dest: "{{ zuul.executor.log_root }}/{{ item }}-iaas.yaml"
         mode: pull
-        src: "report.yaml"
+        src: "report-{{ item }}.yaml"
         verify_host: true
         owner: no
         group: no
+      loop: {{ clouds }}
 
     - name: Return artifact URL
       zuul_return:
         data:
           zuul:
             artifacts:
-              - name: "{{ cloud }}-iaas.yaml"
-                url: "{{ cloud }}-iaas.yaml"
+              - name: "{{ item }}-iaas.yaml"
+                url: "{{ item }}-iaas.yaml"
+      loop: {{ clouds }}
--- a/playbooks/compliance_check.yaml
+++ b/playbooks/compliance_check.yaml
@@ -6,13 +6,13 @@
       ansible.builtin.shell: |
         python3 ~/Tests/scs-compliance-check.py ~/Tests/scs-compatible-iaas.yaml -s {{ item }} -a os_cloud={{ item }} -o report-{{ item }}.yaml -C
         ssh-keygen -Y sign -f ~/id_subject -n report report-{{ item }}.yaml
-        curl --data-binary @report-{{ item }}.yaml.sig --data-binary @report-{{ item }}.yaml -H "Content-Type: application/x-signed-yaml" -H "Authorization: Basic $auth" https://compliance.sovereignit.cloud/reports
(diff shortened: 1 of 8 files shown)


playbooks/compliance_check.yaml
```diff
@@ -4,35 +4,33 @@
   tasks:
     - name: Run compliance script
       ansible.builtin.shell: |
-        python3 ~/Tests/scs-compliance-check.py ~/Tests/scs-compatible-iaas.yaml -s {{ cloud }} -a os_cloud={{ cloud }} -o report.yaml -C
-      register: result
+        python3 ~/Tests/cleanup.py -c {{ item }} --prefix _scs-
+        python3 ~/Tests/scs-compliance-check.py ~/Tests/scs-compatible-iaas.yaml -s {{ item }} -a os_cloud={{ item }} -o report-{{ item }}.yaml -C
+        ssh-keygen -Y sign -f ~/id_subject -n report report-{{ item }}.yaml
+        curl --data-binary @report-{{ item }}.yaml.sig --data-binary @report-{{ item }}.yaml -H "Content-Type: application/x-signed-yaml" -H "Authorization: Basic $auth" https://compliance.sovereignit.cloud/reports
```
martinmo, line 10: Is there a specific technical reason to put all these steps into on task? (Maybe to avoid repeated `loop` setups?) Because before, this was probably a bit more user friendly to have separate tasks for the cleanup, the actual check and the submission to the compliance monitor.

If the loop setup is the reason, maybe you can put the tasks into a separate file, e.g., `compliance_check_steps.yaml` and include that via `include_tasks` (there is also the `block` construct in ansible, but I had to look up if you can loop over blocks and apparently it is not possible, e.g., see [this blog post](https://ericsysmin.com/2019/06/20/how-to-loop-blocks-of-code-in-ansible/)).

playbooks/compliance_check.yaml
```diff
@@ -4,35 +4,33 @@
...
-    - ansible.builtin.debug:
-        msg: "{{ result.stdout }} {{ result.stderr }}"
-
-    - name: sign result YAML and upload to compliance monitor
-      no_log: true  # do not leak the secret in the shell part below
-      ansible.builtin.shell: |
-        ssh-keygen -Y sign -f ~/id_subject -n report report.yaml
-        curl --data-binary @report.yaml.sig --data-binary @report.yaml -H "Content-Type: application/x-signed-yaml" -H "Authorization: Basic {{ clouds_conf.zuul_ci_basic_auth }}" https://compliance.sovereignit.cloud/reports
+      environment:
+        auth: "{{ clouds_conf.zuul_ci_basic_auth }}"
```
martinmo, line 16:
```suggestion
        AUTH_TOKEN: "{{ clouds_conf.zuul_ci_basic_auth }}"
```

Just easier to see if this is also used like this in the `curl` call, plus convention for envvars is upper case AFAIK.

martinmo (commented):
The changes look consistent as far as I can see. Just left a few minor improvement suggestions.

playbooks/compliance_check.yaml
mbuechse: Sure! Somehow the examples in the docs didn't use uppercase, bu t I would prefer that as well. I have to change the curl as well, of course.

mbuechse: 54e320ecb2edb0f709ef26ad059038f6b00419aa

playbooks/compliance_check.yaml
mbuechse: Well, there are reasons, I don't know if they are technical.

1. I dislike Ansible. It's a bunch of peculiar, badly documented yaml that reinvents a lot of wheels just to claim that it's declarative, which doesn't make a lot of sense in this case.
2. I would like to make this whole loop run concurrently, and I think this might be possible with `async` (but then I would have to specify how long I expect this to take??) or I just use the tried and true shell (with `&` and `wait`).
3. Maybe there are more...

mbuechse: BTW: funny, because I had found the same blog post when I asked myself the same question :)

playbooks/compliance_check.yaml
```diff
@@ -4,35 +4,33 @@
   tasks:
     - name: Run compliance script
       ansible.builtin.shell: |
```
gtema, line 6: you lost `no_log: true` to prevent exposure of AUTH_TOKEN

playbooks/compliance_check.yaml
```diff
@@ -4,35 +4,33 @@
...
-        python3 ~/Tests/scs-compliance-check.py ~/Tests/scs-compatible-iaas.yaml -s {{ cloud }} -a os_cloud={{ cloud }} -o report.yaml -C
-      register: result
+        python3 ~/Tests/cleanup.py -c {{ item }} --prefix _scs-
+        python3 ~/Tests/scs-compliance-check.py ~/Tests/scs-compatible-iaas.yaml -s {{ item }} -a os_cloud={{ item }} -o report-{{ item }}.yaml -C
+        ssh-keygen -Y sign -f ~/id_subject -n report report-{{ item }}.yaml
+        curl --data-binary @report-{{ item }}.yaml.sig --data-binary @report-{{ item }}.yaml -H "Content-Type: application/x-signed-yaml" -H "Authorization: Basic $AUTH_TOKEN" https://compliance.sovereignit.cloud/reports
       changed_when: true
       failed_when: false
       # ^^^ this task does not fail; the only failure would be if `report.yaml` didn't get produced,
```
gtema, line 13: the comment is correct, but is it a desired behavior? Failure (i.e. outage) of single cloud could lead to absence of the report for all. Previously different targets were checked by individual jobs. With this change it is not the case anymore and that would lead to the problem. There are few ways how to deal with it in an "Ansible" way:

- stat the expected reports and only sync/return existing ones (something like https://opendev.org/openstack/codegenerator/src/branch/master/playbooks/rust/post.yaml#L46-L72)
- put the "compliance" as such into the ansible role and invoke it in the loop. That helps you to focus on single entity in the tasks instead of messing with loop in every task

Generally I think a combination of both should be done since loops in a main playbook tend to turn into disaster pretty quickly

gtema (changes requested):
Can you, please, also improve PR description. Use of single key is self-explaining, but use of single job is not, especially that it leads immediately to potential issues in the error handling and overall scheduling/scaling

mbuechse:
I will continue here once #715 and #716 are done.

playbooks/pre_cloud.yaml
```diff
@@ -19,13 +19,20 @@
         mode: "0600"
       no_log: true
 
-    - name: Create ssh key file
+    - name: Create signing key file
```
gtema, line 22: you deal with the secret here, no_log is a must since you get secret from variable which might be exposed

mbuechse: This has nothing to do with this PR, but I can do no_log

mbuechse: Screenshot from 2024-09-06 11-29-30
Zuul seems to be smart enough to hide the secret. It says "content: NULL" in the log.
Still, no_log shouldn't hurt. I can do that

gtema: thanks, you should not ever trust any "secret detection" when you know you deal with sensitive data

gtema (approved):

PR #703, opened 2024-08-16
simplify logic using extant (partly improved) library

mbuechse:


Tests/iaas/flavor-naming/flavor-add-extra-specs.py
```diff
@@ -298,64 +208,77 @@ def main(argv):
...
+            reference_key, reference = next(flavorname_items)
+            reference_core = _extract_core(reference)
+            # sanity check: claims must be true wrt actual flavor
+            errors += check_std_props(flavor, reference, " by name")
+            # sanity check: claims must coincide (check remaining flavornames)
+            for key, flavorname in flavorname_items:
+                errors += check_std_props(flavor, flavorname, " by name")
+                core = _extract_core(flavorname)
+                if core != reference_core:
+                    logger.error(f"Inconsistent {key} vs. {reference_key}: {core} vs. {reference_core}")
```
mbuechse, line 240: I think we can't output an error here because understatement is allowed. For instance, if we find a name that says the cpu is a dedicated core and one name that says the cpu is a crowded core, then we have to believe the former one and assume the latter one is an understatement. We could still output a warning though.

True inconsistencies would be something such as cpu architecture Intel vs. AMD.

mbuechse: We have the same problem when setting the values for `scs:cpu-type` – I think we have to consider all flavornames, not just the "reference", in order to find a "maximum".

mbuechse: Or maybe the maximum is not even desired? Maybe the user must disambiguate.

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
```diff
@@ -292,70 +186,86 @@ def main(argv):
...
+                logger.warning(f"Need to specify disk0-type for generating name for {flavor.name}, skipping")
+                continue
+            reference_key = None
+            reference = generate_flavorname(flavor, cpu_type, disk0_type)
+
+        expected = flavorname_to_dict(reference)
+        # set value for any unexpected, but present key to default value
+        # this will help us find necessary removals
+        for key in flavor.extra_specs:
+            expected.setdefault(key, DEFAULTS.get(key))
```
mbuechse, line 243: Only `scs:`

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
mbuechse: I changed it to a warning.

Tests/iaas/flavor-naming/flavor-add-extra-specs.py
```diff
@@ -298,64 +208,77 @@ def main(argv):
...
+
+        # generate default name-vN (ONLY if none present)
```
garloff: My suggestion would be to generate the extra names for a full set of 2 to 4 flavor scs:name-vN names.

mbuechse: Isn't this what's being done?

mbuechse: BTW, you seem to have been reviewing and outdated commit there.

PR #712, opened 2024-08-24
Be selective on FIP and port deletion.

garloff:
* All ports we create are in 10.1.0/24, so pnly delete those ports.
* Floating IPs that are attached won't be deleted.


garloff:
Fixes https://github.com/osism/issues/issues/960

garloff:
Sidenote: We have hardcoded the knowledge that all ports we created are in 10.1.0/24.
This should be a parameter that we pass in, I guess. For now the comment in the code will have to do.

garloff:
OK, testing has succeeded and yielded two cosmetic improvements:
- The help function in cleanup had one parameter name confused (`-p` vs `-i`)
- The entropy test could have also assigned a name to the subnet (not required for cleanup, but still preferrable)

garloff:
With these changes, the gxscs halth mon will work reliably again, so reenable link to it.

garloff:
> Sidenote: We have hardcoded the knowledge that all ports we created are in 10.1.0/24.
> This should be a parameter that we pass in, I guess. For now the comment in the code will have to do.

This has been fixed with the last commit.
We have reasonable logic now to select which ports to clean up that is controlled by parameters.

Tests/cleanup.py
```diff
@@ -192,6 +222,11 @@ def main(argv):
             prefix = opt[1]
         if opt[0] == "-c" or opt[0] == "--os-cloud":
             cloud = opt[1]
+        if opt[0] == "-i" or opt[0] == "--ipaddr":
+            if "," in opt[1]:
+                ipaddrs = opt[1].split(",")
+            else:
+                ipaddrs = opt[1]
```
mbuechse, line 229: I don't think this will work. Shouldn't it be `[opt[1]]`? Otherwise, if you iterate, you will get the substrings of length 1...

mbuechse: The best solution in my eyes is to always do the split, regardless of whether a comma is contained. The only annoying case is the empty string, which gets us `[""]`. (Matching any IP address, so probably even what is wanted.) However, this case doesn't seem too relevant here.

Tests/cleanup.py
```diff
@@ -26,15 +26,18 @@ def print_usage(file=sys.stderr):
...
+ [-c/--os-cloud OS_CLOUD]  sets cloud environment (default from OS_CLOUD env)
+ [-p/--prefix PREFIX]      sets prefix to identify resources (default from PREFIX env)
+ [-i/--ipaddr addr[,addr]] list of IP addresses to identify ports to delete (def: delete all)
+                           the specified strings will be matched against the start of the addrs
 """, end='', file=file)
 
 
 class Janitor:
-    def __init__(self, conn, prefix=""):
+    def __init__(self, conn, prefix="", ipfilter=[]):
```
mbuechse, line 37: Default values shouldn't be mutable. Better use a tuple here. Any list is also a tuple, so type-wise this is fine.

```suggestion
    def __init__(self, conn, prefix="", ipfilter=()):
```

mbuechse (commented):
Great work! Please have a look at my minor remarks.

Tests/cleanup.py
garloff: Indeed, adding insult to injury. Fixed with commit https://github.com/SovereignCloudStack/standards/pull/712/commits/28219cf384181f02aa239da8c2737257da013e7b.
Thanks!

playbooks/pre_cloud.yaml
```diff
@@ -27,5 +27,5 @@
         mode: "0600"
 
     - name: Clean up any lingering resources from previous run
-      ansible.builtin.shell: python3 ~/Tests/cleanup.py -c {{ cloud }} --prefix _scs-
+      ansible.builtin.shell: python3 ~/Tests/cleanup.py -c {{ cloud }} --prefix _scs- --ipaddr 10.1.0.
```
mbuechse, line 30:
```suggestion
      ansible.builtin.shell: python3 ~/Tests/cleanup.py -c {{ cloud }} --prefix _scs- --ipaddr 10.1.0.1
```

should be possible, since you changed the allocation pool to end with .199

mbuechse (approved):
LGTM, with one optional improvement (if I'm not mistaken)

playbooks/pre_cloud.yaml
garloff: This was the reason to change the allocation pool, as you correctly guessed, yes.
And then I refrained from using `10.1.0.1` over `10.1.0.`, as I was not 100% sure that we would not occasionally want the cleanup script to hit a DHCP port [.2] or gateway [.1] in this subnet in case the tear down has gone terribly wrong, so I left it with hitting the complete subnet. (OK, gateway would actually be hit ... as it matches the `startswith()` comparison as well.)

Issue #716, opened 2024-08-28
Compliance test: create simple end-user cli


mbuechse:
Create a CLI that can run tests for multiple subjects (in parallel if so desired) and upload the results to the compliance monitor. The CLI should be able to use a config file so that options that stay the same for the user need not be repeated for each invocation.

mbuechse:
WIP: simple command-line to test all clouds (at most 4 at the same time, notice that gx-scs will only be checked once wavestack is finished), put everything into one yaml and uplaod that to the compliance monitor

```shell
$ ./scs-test-runner.py run --preset all
INFO: module opc-v2022.11 missing checks or test cases
DEBUG: running './iaas/flavor-naming/flavor-names-openstack.py -c poc-wgcloud --mand=./iaas/scs-0100-v3-flavors.yaml'...
INFO: module opc-v2022.11 missing checks or test cases
DEBUG: running './iaas/flavor-naming/flavor-names-openstack.py -c wavestack --mand=./iaas/scs-0100-v3-flavors.yaml'...
INFO: module opc-v2022.11 missing checks or test cases
DEBUG: running './iaas/flavor-naming/flavor-names-openstack.py -c syseleven-ham1 --mand=./iaas/scs-0100-v3-flavors.yaml'...
INFO: module opc-v2022.11 missing checks or test cases
DEBUG: running './iaas/flavor-naming/flavor-names-openstack.py -c artcodix --mand=./iaas/scs-0100-v3-flavors.yaml'...
DEBUG: .. rc 0, 0 critical, 0 error
DEBUG: running './iaas/entropy/entropy-check.py -c wavestack -d'...
DEBUG: .. rc 0, 0 critical, 0 error
DEBUG: running './iaas/entropy/entropy-check.py -c syseleven-ham1 -d'...
DEBUG: .. rc 0, 0 critical, 0 error
DEBUG: running './iaas/entropy/entropy-check.py -c poc-wgcloud -d'...
DEBUG: .. rc 0, 0 critical, 0 error
DEBUG: running './iaas/entropy/entropy-check.py -c artcodix -d'...
DEBUG: .. rc 0, 0 critical, 0 error
DEBUG: running './iaas/image-metadata/image-md-check.py -c wavestack -v -s'...
DEBUG: .. rc 0, 0 critical, 0 error
DEBUG: running './iaas/standard-flavors/flavors-openstack.py -c wavestack -d ./iaas/scs-0103-v1-flavors.yaml'...
[...]

PR #717, opened 2024-08-28
accept multiple reports in one yaml

### LINKED ISSUE #715: Compliance monitor: accept multiple reports in one yaml

mbuechse:
resolves #715

### DIFF
--- a/compliance-monitor/monitor.py
+++ b/compliance-monitor/monitor.py
@@ -426,14 +426,14 @@ async def post_report(
     body_text = body.decode("utf-8")
     sep = body_text.find(SEP)
     if sep < 0:
-        raise HTTPException(status_code=401)
+        raise HTTPException(status_code=401, detail="missing signature")
     sep += len(SEP)
     signature = body_text[:sep - 1]  # do away with the ampersand!
     body_text = body_text[sep:]
     try:
         ssh_validate(keys, signature, body_text)
-    except ValueError:
-        raise HTTPException(status_code=401)
+    except Exception:
+        raise HTTPException(status_code=401, detail="verification failed")
 
     json_texts = []
     if content_type.endswith('-yaml'):
@@ -444,7 +444,8 @@ async def post_report(
         documents = [json.loads(body_text)]
         json_texts = [body_text]
     else:
-        raise HTTPException(status_code=500)
+        # unreachable due to the content-type check at the top
+        raise AssertionError("branch should never be reached")
 
     allowed_subjects.add(account[0])
     for document in documents:
--- a/compliance-monitor/monitor.py
+++ b/compliance-monitor/monitor.py
@@ -460,7 +460,7 @@ async def post_report(
             uuid, subject, checked_at = rundata['uuid'], document['subject'], document['checked_at']
             scopeuuid = document['spec']['uuid']
             try:
-                reportid = db_insert_report(cur, uuid, checked_at, subject, json_text, content_type, body)
+                reportid = db_insert_report(cur, uuid, checked_at, subject, json_text)
             except UniqueViolation:
                 raise HTTPException(status_code=409, detail="Conflict: report already present")
             for version, vdata in document['versions'].items():
--- a/compliance-monitor/monitor.py
+++ b/compliance-monitor/monitor.py
@@ -418,9 +418,10 @@ async def post_report(
         # see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/415
         raise HTTPException(status_code=415, detail="Unsupported Media Type")
 
+    auth_subject, _ = account
     with conn.cursor() as cur:
-        keys = db_get_keys(cur, account[0])
-        allowed_subjects = set(db_find_subjects(cur, account[0]))
+        keys = db_get_keys(cur, auth_subject)
+        delegation_subjects = db_find_subjects(cur, auth_subject)
 
     body = await request.body()
     body_text = body.decode("utf-8")
@@ -447,7 +448,7 @@ async def post_report(
         # unreachable due to the content-type check at the top
         raise AssertionError("branch should never be reached")
 
-    allowed_subjects.add(account[0])
+    allowed_subjects = {auth_subject} | set(delegation_subjects)
     for document in documents:
         check_role(account, document['subject'], ROLES['append_any'])
         if document['subject'] not in allowed_subjects:
--- a/compliance-monitor/monitor.py
+++ b/compliance-monitor/monitor.py
@@ -29,8 +29,9 @@ import uvicorn
 from sql import (
     db_find_account, db_update_account, db_update_publickey, db_filter_publickeys, db_get_reports,
     db_get_keys, db_insert_report, db_get_recent_results2, db_patch_approval2, db_get_report,
-    db_ensure_schema, db_get_apikeys, db_update_apikey, db_filter_apikeys,
-    db_patch_subject, db_get_subjects, db_insert_result2, db_get_relevant_results2,
+    db_ensure_schema, db_get_apikeys, db_update_apikey, db_filter_apikeys, db_clear_delegates,
+    db_patch_subject, db_get_subjects, db_insert_result2, db_get_relevant_results2, db_add_delegate,
+    db_find_subjects,
 )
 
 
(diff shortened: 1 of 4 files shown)


mbuechse:
It is now indeed possible to run the tests in parallel for multiple subjects, then put everything into one yaml, sign that one yaml, and upload it:

```bash
#! /bin/bash
declare -A subjects=( [wavestack]=wave1-compliance [artcodix]=artcodix [poc-wgcloud]=poc-wgcloud [gx-scs]=gxscs-compliance )

for subject in ${!subjects[@]}
do
    ../Tests/scs-compliance-check.py ../Tests/scs-compatible-iaas.yaml -C -o /tmp/tmpreport-${subject}.yaml -a "os_cloud=${subjects[$subject]}" -s "$subject" --debug &
done
wait
rm -f /tmp/tmpreport.yaml.sig
echo "" > /tmp/tmpreport.yaml
for subject in ${!subjects[@]}
do
    cat /tmp/tmpreport-${subject}.yaml >> /tmp/tmpreport.yaml
done
ssh-keygen -Y sign -f REDACTED -n report /tmp/tmpreport.yaml
time curl --data-binary @/tmp/tmpreport.yaml.sig --data-binary @/tmp/tmpreport.yaml -H "Content-Type: application/x-signed-yaml" -H "Authorization: Basic REDACTED" http://localhost/reports
```

mbuechse:
The next step then is to turn this bash snippet into a cool Python script that offers choosing the number of worker processes etc. -- see #716.

compliance-monitor/monitor.py
```diff
@@ -421,36 +430,42 @@ async def post_report(
...
-        json_text = json.dumps(document, cls=TimestampEncoder)
+        documents = yaml.load_all(body_text)
+        json_texts = [json.dumps(document, cls=TimestampEncoder) for document in documents]
     elif content_type.endswith("-json"):
-        document = json.loads(body_text)
-        json_text = body_text
+        documents = [json.loads(body_text)]
+        json_texts = [body_text]
     else:
         raise HTTPException(status_code=500)
```
martinmo, line 447: maybe use `raise AssertionError("branch should never be reached")` instead, has same result but is clearer

compliance-monitor/monitor.py
```diff
@@ -413,6 +417,11 @@ async def post_report(
     if content_type not in ('application/x-signed-yaml', 'application/x-signed-json'):
         # see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/415
         raise HTTPException(status_code=415, detail="Unsupported Media Type")
+
+    with conn.cursor() as cur:
+        keys = db_get_keys(cur, account[0])
```
martinmo, line 422: `account[0]` is used in several places and seems to refer to account name/subject name (?), might be good for readers to have it saved in a dedicated variable `subject_name`.

The type in the function signature also doesn't give more information about the meaning, it just says `tuple[str,str]`

compliance-monitor/monitor.py
```diff
@@ -421,36 +430,42 @@ async def post_report(
     sep += len(SEP)
     signature = body_text[:sep - 1]  # do away with the ampersand!
     body_text = body_text[sep:]
-    json_text = None
+    try:
+        ssh_validate(keys, signature, body_text)
+    except ValueError:
+        raise HTTPException(status_code=401)
```
martinmo, line 436: Potential problem: same status code used as above and no specific error message in the response, which can make debugging difficult for clients using the endpoint. (Of course, we also should not reveal too much information.)

compliance-monitor/monitor.py
```diff
@@ -421,36 +430,42 @@ async def post_report(
...
-    check_role(account, subject, ROLES['append_any'])
-    with conn.cursor() as cur:
-        keys = db_get_keys(cur, subject)
-    try:
-        ssh_validate(keys, signature, body_text)
-    except ValueError:
-        raise HTTPException(status_code=401)
-    scopeuuid = document['spec']['uuid']
+
+    allowed_subjects.add(account[0])
```
martinmo, line 449: This should be moved to where `allowed_subjects` is initially created, or `db_find_subjects(cursor, X)` always should return a list containing at least `X`.

compliance-monitor/monitor.py
```diff
@@ -421,36 +430,42 @@ async def post_report(
...
-                approval = 1 == result  # pre-approve good result
-                db_insert_result2(cur, checked_at, subject, scopeuuid, version, check, result, approval, reportid)
+        for document, json_text in zip(documents, json_texts):
+            rundata = document['run']
+            uuid, subject, checked_at = rundata['uuid'], document['subject'], document['checked_at']
+            scopeuuid = document['spec']['uuid']
+            try:
+                reportid = db_insert_report(cur, uuid, checked_at, subject, json_text, content_type, body)
+            except UniqueViolation:
+                raise HTTPException(status_code=409, detail="Conflict: report already present")
```
martinmo, line 463: What is the intended behavior with the other reports, which did not fail until the violation? Just asking if we would need some kind of all-or-nothing transactional style here instead.

compliance-monitor/monitor.py
```diff
@@ -421,36 +430,42 @@ async def post_report(
...
-            for check, rdata in vdata.items():
-                result = rdata['result']
-                approval = 1 == result  # pre-approve good result
-                db_insert_result2(cur, checked_at, subject, scopeuuid, version, check, result, approval, reportid)
+        for document, json_text in zip(documents, json_texts):
+            rundata = document['run']
+            uuid, subject, checked_at = rundata['uuid'], document['subject'], document['checked_at']
+            scopeuuid = document['spec']['uuid']
+            try:
+                reportid = db_insert_report(cur, uuid, checked_at, subject, json_text, content_type, body)
```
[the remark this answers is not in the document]
martinmo: > one solution would be to drop the whole column and no longer fill it

Yes, this sounds like a good idea

martinmo (changes requested):
Nice feature! First review round done.

compliance-monitor/monitor.py
mbuechse: Yes, I didn't want to reveal too much, but I could add a misty nebulous message

compliance-monitor/monitor.py
mbuechse: I will do something about that. I think the call to db_find_subjects should just store the result in a list, and it should be clear that this list is NOT the list of allowed subjects, but the list of delegations or something. And only here would I construct the set using both this list and the login info.

compliance-monitor/monitor.py
mbuechse: This whole thing is inside one database transaction, so we do have that already, or did I miss something?

compliance-monitor/monitor.py
mbuechse: 772b60390bfa214742e1b120fc5bf1a266bc2c41

compliance-monitor/monitor.py
mbuechse: 772b60390bfa214742e1b120fc5bf1a266bc2c41

compliance-monitor/monitor.py
mbuechse: 9f10620785229d67c4f6cc39b303fd829445d4d3

compliance-monitor/monitor.py
mbuechse: 9f10620785229d67c4f6cc39b303fd829445d4d3

compliance-monitor/monitor.py
martinmo: Ok, I think I missed it. I was confused because there is only an explicit commit at the end in the happy case.

So we have an implicit start of the transaction because of `Depends(get_conn)`. And with the `finally` in `get_conn`, the connection is closed after `post_report` returns and thus any non-committed transaction should be rolled back. Right?

mbuechse: Yes, the get_conn calls conn.close at the end, and without an explicit commit, this will roll back everything.

mbuechse: The implicit start (begin transaction) is being managed by psycopg, I think, with the first statement that is sent to the database: https://www.psycopg.org/psycopg3/docs/basic/transactions.html

mbuechse:
Removed column: 58c8ee02142ad9134ad3b4b47379658704bc19d0

martinmo (changes requested):

compliance-monitor/sql.py
```diff
@@ -282,12 +328,12 @@ def db_get_reports(cur: cursor, subject, limit, skip):
     return [row[0] for row in cur.fetchall()]
 
 
-def db_insert_report(cur: cursor, uuid, checked_at, subject, json_text, content_type, body):
+def db_insert_report(cur: cursor, uuid, checked_at, subject, json_text):
     # this is an exception in that we don't use a record parameter (it's just not as practical here)
     cur.execute('''
     INSERT INTO report (reportuuid, checked_at, subject, data, rawformat, raw)
     VALUES (%s, %s, %s, %s, %s, %s)
```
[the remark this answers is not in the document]
mbuechse: this defect: 21819a4cd7a0b14d2d2e78ba70afa30b7b84d0a8
and the other one that prevented my tests from revealing the first defect: bdceccfdd3aeab948cceae4dbcb36edb471ee7c4

martinmo (approved):

PR #720, opened 2024-08-29
Fix: scs-0100-v3.2 doesn't exist; it's actually v3.0 and v3.1

mbuechse:


garloff (approved):
Thanks for the correction, it was incorrect before, indeed

PR #721, opened 2024-08-29
Remove outdated compliance-check workflows

mbuechse:


garloff:
No problem with removing the v2 jobs.

Not so sure about v3: We have some clouds that do not pass v4 yet (and yes, working on fixing this!), so we might want to keep the ability to test v3.

Another reason to keep v3: We want to write standards such that cloud that pass v4 also pass v3 ... (or more general pass v<N-1> when vN is current). This provides backwards compatibility for users (Developers and Operators of apps written against an old SCS spec). Except for the v1 vs v2 change, we are in generally achieving this most of the time. I think we'd want to test it to be sure.
(Note: We have dropped some flavors from mandatory to recommended, which allows a v3 pass to be a v2 fail, but this is likely not hitting many users and our real-life clouds have not dropped the flavors anyway.)

In general it would seem wise to me to
* always test v<CURRENT> and display the result prominently. This is what counts.
* continue testing v<PREVIOUS> and have this available somewhere. Some clouds will be old. Some apps will have been developed against it.
* start testing v<NEXT> as soon as we're close to stabilizing it.

Let me know what you think.

mbuechse:
@garloff This sounds like a veritable discussion topic for our beloved SIG.

I'm a bit surprised that we want to care about old versions. But you have a point, and we might say that deprecated versions are "merely deprecated", as opposed to "obsolete¨. If we want to check the latest deprecated version, should we maybe call the others "obsolete", just to make this more explicit?

garloff:
Matthias,
I agree that this calls for a discussion.
Our standards should mostly move into one direction, i.e. becoming more encompassing and covering more pieces that application DevOps teams can rely upon. So a cert scope vCURRENT will typically be a superset of vPREVIOUS or close to it.
That said, we will occasionally stop requiring things or move things from MANDATORY to RECOMMENDED.
While we want our operators to follow the new standards quickly, we have less of a leverage on application developers. They will have developed against SCS-XaaS-vN and expect the application to work on SCS-XaaS-vN+. That should be typically the case, but to be sure, we can simply still test vPREVIOUS and see if a cloud really passes vCURRENT but not vPREVIOUS. It should be a rare exception ... making it visible seems useful to me.

mbuechse:
@garloff Your logic is convincing, but you haven't answered my question. Shall we be more explicit by distinguishing between deprecated and obsolete? This whole business with vCURRENT, vPREV seems a bit complicated to me.

garloff:
No clear opinion on deprecated vs obsolete.
Disadvantage: It adds one more state and one more state transition that we need to trigger.
Advantage: It's a bit more transparent and flexible than a statement that we keep testing the previous version to have visibility on (not) breaking it.

mbuechse:
@garloff We already have the timeline in our yaml file:

```yaml
timeline:
  - date: 2024-08-23
    versions:
      v5: draft
      v4: effective
  - date: 2024-07-31
    versions:
      v4: effective
  - date: 2024-04-30
    versions:
      v4: effective
      v3: warn
# and so on
```

We say that every unlisted version is considered `deprecated` (and vice versa: `deprecated` versions shouldn't be listed explicitly).

It would be very simple to say: yes, we do indeed list `deprecated` versions explicity, and any unlisted version is indeed `obsolete`.

Or we could just say: it is now permissible to list a `deprecated` version, and the only difference between not listing it and listing it explicitly as `deprecated` is that we want all versions that are listed here to be tested. So we avoid another formal state.

garloff:
The latter looks a tad easier to me.

garloff (approved):
/lgtm

PR #722, opened 2024-08-29
Create simple end-user cli

### LINKED ISSUE #716: Compliance test: create simple end-user cli

mbuechse:
eventually resolves #716

### DIFF
--- a/Tests/scs-test-runner.py
+++ b/Tests/scs-test-runner.py
@@ -51,6 +51,12 @@ class Config:
             self._auth_token = fileobj.read().strip()
         return self._auth_token
 
+    def get_subject_mapping(self, subject):
+        default_mapping = self.subjects.get('_', {}).get('mapping', {})
+        mapping = {key: value.format(subject=subject) for key, value in default_mapping.items()}
+        mapping.update(self.subjects.get(subject, {}).get('mapping', {}))
+        return mapping
+
     def abspath(self, path):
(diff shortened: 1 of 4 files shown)


mbuechse:
It's very rudimentary but it does work so far.

Tests/scs-test-runner.py
```diff
@@ -0,0 +1,180 @@
...
+        mapping.update(self.subjects.get(subject, {}).get('mapping', {}))
+        return mapping
+
+    def abspath(self, path):
+        return os.path.join(self.cwd, path)
+
+    def build_check_command(self, scope, subject, output):
+        cmd = [
+            sys.executable, self.scs_compliance_check, self.abspath(self.scopes[scope]['spec']),
+            '--debug', '-C', '-o', output, '-s', subject,
```
martinmo, line 66: Is it intentional to always pass `--debug` to the compliance test script even if it is disabled via `scs-test-runner.py --debug=false`?

martinmo (approved):

Tests/scs-test-runner.py
mbuechse: I would like to keep these two separated. But the question is pertinent...

Issue #650, opened 2024-06-26
Revise entropy check to work without floating IP


mbuechse:
Floating IPs are not viable for every provider. It might be possible to do without them, as follows:

- include necessary commands in cloud-init, see https://github.com/osism/openstack-simple-stress/blob/main/openstack_simple_stress/main.py#L275
- parse console output as provided by Nova, see https://github.com/osism/openstack-simple-stress/blob/main/openstack_simple_stress/main.py#L208

PR #724, opened 2024-08-30
Only consider image properties of public/community images.

garloff:
We may not have the luxury to test against a pristine project, so there may be private images or shared images from others. Only consider provider images with "public" or "community" visibility.
<https://wiki.openstack.org/wiki/Glance-v2-community-image-visibility-design>

Note: While testing PR #713, this did not fail the overall test, I saw a failed message on check-image-properties which was due to private images ... which could have been avoided.

### DIFF
--- a/Tests/iaas/entropy/entropy-check.py
+++ b/Tests/iaas/entropy/entropy-check.py
@@ -85,6 +85,7 @@ Options:
  [-c/--os-cloud OS_CLOUD] sets cloud environment (default from OS_CLOUD env)
  [-d/--debug] enables DEBUG logging channel
  [-i/--images IMAGE_LIST] sets images to be tested, separated by comma.
+ [-V/--image-visibility VIS_LIST] filters images by visibility (default: public,community)
 """, end='', file=file)
 
 
@@ -401,7 +402,7 @@ def main(argv):
     logger.addHandler(counting_handler)
 
     try:
-        opts, args = getopt.gnu_getopt(argv, "c:i:hd", ["os-cloud=", "images=", "help", "debug"])
+        opts, args = getopt.gnu_getopt(argv, "c:i:hdV:", ["os-cloud=", "images=", "help", "debug", "image-visibility="])
     except getopt.GetoptError as exc:
         logger.critical(f"{exc}")
         print_usage()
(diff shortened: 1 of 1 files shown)


berendt:
This should be configurable if a project is used in which there are no public or community images but only private images. Otherwise, such a project could not be used for testing. In the future, we will at least have the case that some of our projects do not contain images that are public/community. The images in the projects are still managed by us and contain all the necessary meta information.

Tests/iaas/entropy/entropy-check.py
```diff
@@ -428,7 +428,7 @@ def main(argv):
     try:
         logger.debug(f"Connecting to cloud '{cloud}'")
         with openstack.connect(cloud=cloud, timeout=32) as conn:
-            all_images = conn.list_images()
+            all_images = list(filter(lambda x: x.visibility == 'public' or x.visibility == 'community', conn.list_images()))
```
mbuechse, line 431: AFAICT, the image-metadata check uses server-side filtering, and it only checks public images. Shall we change that as well?

garloff: Three points:
1. Having consistency between the two checks would always be preferable.
2. I did not see an obvious way to do server side filtering, but I may not have looked deeply enough. It's always preferable.
3. I chose `public` and `community` because these are controlled by the cloud operator.

garloff:
Wouldn't you fail the image-md check if you don't at least have a public `Ubuntu 22.04`?
(Also I wonder why you provide your standard images as private images... does not seem efficient to me at all.) 

Adding an option `--all-visibility` is not a big deal. Helps in your case if you call the check manually with the option, but not if run from the conformance test suite...

Tests/iaas/entropy/entropy-check.py
mbuechse: So, let's change both tests, then?

berendt:
In at least one environment we have the requirement to work with private images only. It's not about efficiency there. To be as compliant as possible we make an image like the mentioned Ubuntu 22.04 available as a private image in every project of this environment so that it can be used in general. But it is neither public nor community.

garloff:
You can pass `--image-visibility=""` (or `--image-visibility="public,community,shared,private"` if you prefer) now to the entropy-check. The SCS conformance test however won't call the entropy checker with these flags ...

garloff:
@mbuechse -- it seems to me that this is good to merge now.
However, I created a PR against your #650 branch (PR #721), which may not be your preferred way to get this in.
Let me know if you prefer me rebasing against main and doing a PR against it.

PR #729, opened 2024-09-03
If something does not work, log the line number in the cleanup script

berendt:


berendt:
Reference: https://stackoverflow.com/questions/1278705/when-i-catch-an-exception-how-do-i-get-the-type-file-and-line-number

mbuechse:
I would prefer #730

PR #731, opened 2024-09-03
Bugfix: reflect (formerly, temporarily) mistaken docs in iaas scope

mbuechse:
I think the docs were wrong from ca. Nov 30, 2023 through Aug 14, 2024, between the introduction of (then-draft) v4 and the changes introducing fine-grained certificate scopes.

The scs-0100-v3 always stated that v3.0 still contained the list of flavors, but we didn't explicitly state that scope v3 referred to v3.0 of that standard instead of v3.1.

### DIFF
--- a/Tests/scs-compatible-iaas.yaml
+++ b/Tests/scs-compatible-iaas.yaml
@@ -160,6 +160,7 @@ timeline:
       v5: draft
       v4: effective
       v3: deprecated
+      v3-orig: deprecated
   - date: 2024-07-31
     versions:
       v4: effective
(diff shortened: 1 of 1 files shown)


mbuechse:
To be sure:

```text
syseleven-ham1 SCS-compatible IaaS v3 (deprecated):
- main: PASS (2 passed)
syseleven-ham1 SCS-compatible IaaS v3-orig (deprecated):
- main: FAIL (1 passed, 1 failed)
  - FAILED:
    - flavor-name-check:
      > Must fulfill all requirements of <https://docs.scs.community/standards/scs-0100-v3-flavor-naming -- plus the list of mandatory> and recommended flavors found in <https://docs.scs.community/standards/scs-0100-v2-flavor-naming#standard-scs-flavors>
```

garloff (approved):
Thanks!
This will make one of our partners happy, as it confirms that they had done what's needed by the docs at some point in our history.
Obviously we all need to focus on v4 (and at some point v5) now.

PR #732, opened 2024-09-03
if sonic adr is global, the number should reflect this

fkr:
Signed-off-by: Felix Kronlage-Dammers <fkr@hazardous.org

### DIFF
--- a/Standards/scs-0005-v1-sonic.md
+++ b/Standards/scs-0005-v1-sonic.md
@@ -9,7 +9,7 @@ description: |
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


mbuechse (changes requested):
Sign-off doesn't work due to missing closing angular bracket (I suppose).
Rename probably okay.
The other changes introduce more typos than they rectify. Better to do the rename in isolation.

fkr:
@mbuechse actually sorry for this weird PR. The file should’ve been just renamed, no idea where the content changes come from. Sorry to have wasted your time.

PR #741, opened 2024-09-06
Bugfix: add missing directory

mbuechse:
do it this way instead of complicating all manner of playbooks etc.


gtema (commented):
to be improved in the next iteration touching playbook, but is ok for now

gtema (approved):

PR #742, opened 2024-09-06
Allow reconfiguration without restart

### LINKED ISSUE #666: Compliance Monitor: allow reconfiguration without restart

mbuechse:
Closes #666.

### DIFF
--- a/compliance-monitor/monitor.py
+++ b/compliance-monitor/monitor.py
@@ -1,4 +1,17 @@
 #!/usr/bin/env python3
+# AN IMPORTANT NOTE ON CONCURRENCY:
+# This server is based on uvicorn and, as such, is not multi-threaded.
+# (It could use multiple processes, but we don't do that yet.)
+# Consequently, we don't need to use any measures for thread-safety.
+# However, if we do at some point enable the use of multiple processes,
+# we should make sure that all processes are "on the same page" with regard
+# to basic data such as certificate scopes, templates, and accounts.
+# One way to achieve this synchronicity could be to use the Postgres server
+# more, however, I hope that more efficient ways are possible.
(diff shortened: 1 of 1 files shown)


mbuechse:
This is still in draft state. What it does so far: upon receiving SIGUSR2, it will increase the global counter, thus causing a reload of all certificate scopes next time they are needed.

I confirmed that it works by changing the file and invoking

```shell
docker compose kill -s SIGUSR2 web
```

However, if the file cannot be loaded (because of syntax errors, for instance), the error will only appear later on (as I said, the next time the certificate scopes are needed). I want to see the error at once. For this and other reasons, I still want to implement the following changes

- load the scope files at once
- either share the precomputed values across threads (because they are read only) or compute them from pre-loaded spec files
- also reload bootstrap.yaml
- (maybe?) no longer keep account data in the database, but use the same technique as for the scopes?

martinmo (commented):
Is it intentional that functionality which seems unrelated to online reload is removed in this PR? (e.g., `get_subjects` and so on.)

The signal part seems to be alright as far as I can see, *if* really only a single process is ever used to serve requests. That might change when workers are scaled, because this starts separate Python processes. Also, in the future, using USR2 *could* be problematic if the server needs to be changed (e.g., using gunicorn to run the uvicorn workers), because libraries like gunicorn could be using signals for similar purposes of "online reconfiguration".

mbuechse:
@martinmo It's now based on #751 to remove the subject parts, so maybe merge that one first.

Your remarks regarding multiprocessing are correct, and I added a note. We would have to change a few things anyway, and frankly, this is a problem for the far future.

martinmo (approved):

PR #745, opened 2024-09-09
Issue/662 default storage class

fraugabel:
fixed two bugs: 
1. make sure sonobuoy result file can be written
2. cleanup test-pod at failure aswell, so the next run won't crash


Tests/kaas/k8s-default-storage-class/helper.py
```diff
@@ -56,19 +56,17 @@ def setup_k8s_client(kubeconfigfile=None):
...
-    result_file["name"] = test_name
-    result_file["status"] = test_status
-    result_file["details"]["messages"] = error_msg
+        directory_path = os.path.dirname(f"./{test_name}.result.yaml")
+        os.makedirs(directory_path, exist_ok=True)
 
-    with open(f"./{test_name}.result.yaml", "w") as file:
-        yaml.dump(result_file, file)
+        with open(f"./{test_name}.result.yaml", "w") as file:
+            yaml.dump(result_file, file)
```
martinmo, line 72: This still creates a file hierarchy replicating the absolute path to the script file, e.g. for me it creates:

`/home/martinmorgenstern/workspace/scs/standards/Tests/home/martinmorgenstern/workspace/scs/standards/Tests/kaas/k8s-default-storage-class`

The reason is that `__file__` (which is the absolute path to the script) is used as the `test_file_name` argument passed to `gen_sonobuoy_result_file` and used without further processing.

In `main()`, you should use `os.path.basename(__file__)` to get only the script filename without the directories.

Tests/kaas/k8s-default-storage-class/helper.py
```diff
@@ -56,19 +56,17 @@ def setup_k8s_client(kubeconfigfile=None):
 
 
 def gen_sonobuoy_result_file(error_n: int, error_msg: str, test_file_name: str):
-
     test_name = test_file_name.replace(".py", "")
```
martinmo, line 59:
```suggestion
    test_name = test_file_name.removesuffix(".py")
```

martinmo (commented):
First round of review done, see inline comments.

Some general observations:

If the connection to the cluster fails, the script tries to continue anyway with the all the steps, which it shouldn't do.

The script still doesn't check if a provisioner such as cinder-csi is used.

Tests/kaas/k8s-default-storage-class/helper.py
fraugabel: yes you're right. ithought i already changed that

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -135,10 +143,9 @@ def check_default_persistentvolumeclaim_readwriteonce(k8s_api_instance, storage_
     pod_info = json.loads(api_response.read().decode("utf-8"))
     pod_status = pod_info["status"]["phase"]
 
-    # Check if pod is up and running:
+    # Check if pod is up and running: Default Retries 30
```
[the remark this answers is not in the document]
fraugabel: good idea

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -152,6 +159,7 @@ def check_default_persistentvolumeclaim_readwriteonce(k8s_api_instance, storage_
     if pod_status != "Running":
         raise SCSTestException(
             "pod is not Running not able to setup test Enviornment",
+            cleanup(),
```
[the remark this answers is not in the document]
fraugabel: you're right, though SCSTestException() still worked as expected

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -172,28 +180,24 @@ def check_default_persistentvolumeclaim_readwriteonce(k8s_api_instance, storage_
...
                 )
 
     # 4. Delete resources used for testing
-    logger.debug(f"delete pod:{pod_name}")
-    api_response = k8s_api_instance.delete_namespaced_pod(pod_name, namespace)
-    logger.debug(f"delete pvc:{pvc_name}")
-    api_response = k8s_api_instance.delete_namespaced_persistent_volume_claim(
-        pvc_name, namespace
-    )
+    cleanup()
```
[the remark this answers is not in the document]
fraugabel: thanks, i'll checkout the context manager, that seems to be pretty neat

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -73,23 +81,15 @@ def check_default_storageclass(k8s_client_storage):
...
     1. Create PersistantVolumeClaim
     2. Create pod which uses the PersitantVolumeClaim
-    3. Check if PV got succesfully created using ReadWriteOnce
-    4. Delete resources used for testing
     """
-
-    namespace = "default"
-    pvc_name = "test-pvc"
-    pv_name = "test-pv"
-    pod_name = "test-pod"
```
mbuechse, line 87: please turn these into arguments with default values, i.e.,

`namespace=NAMESPACE, pvc_name=PVC_NAME, ...`

this makes the function more versatile, and we don't litter the code with these allcaps identifiers

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -25,8 +25,11 @@
 import time
 import json
 import logging
+import os
+import inspect
```
mbuechse, line 29: it hasn't been sorted before, but please make it sorted

- builtin Python modules, sorted by alphabet
- blank line
- modules from dependencies, sorted by alphabet
- blank line
- modules from the same source tree, sorted by alphabet

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -180,26 +182,112 @@ def check_default_persistentvolumeclaim_readwriteonce(k8s_api_instance, storage_
...
+    def __init__(self, kubeconfig):
+        self.namespace = NAMESPACE
+        self.pod_name = POD_NAME
+        self.pvc_name = PVC_NAME
+        self.k8s_core_api = None
+        self.return_code = 0
+        self.return_message = "return_message: FAILED"
+        self.kubeconfig = kubeconfig
+
+    def prepare(self):
```
mbuechse, line 199: This function has quite a bit of logic. The return value should be documented!

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -180,26 +182,112 @@ def check_default_persistentvolumeclaim_readwriteonce(k8s_api_instance, storage_
...
+        self.clean()
+        if self.return_code == 0:
+            self.return_message = "all tests passed"
+        if isinstance(exc_value, SCSTestException):
+            self.return_message = exc_value.args[0]
+            self.return_code = exc_value.return_code
+            print(f"SCSTestException occurred with return_code: {self.return_code}")
+        else:
+            # No specific exception, handle normally
+            print(f"Exiting the context with return_code: {self.return_code}")
```
mbuechse, line 274: This seems really convoluted. The exceptions are being caught in the main function anyway. And why should the test environment carry the burden of storing the return code? That violates the single-responsibility principle (aka separation of concerns).

mbuechse (commented):
I'm not sure I can follow, but it does seem to be work in progress anyway.

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
fraugabel: done (isort)

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
fraugabel: because in this case I assumed that these exceptions are only intercepted afterwards. i'll check this again.

fraugabel: the return code is kept there to feed it into the sonobuoy result, which is created in the exit, to make sure it gets created, after any error whatsoever. This is actually the reason why i created this context management to avoid having the same exception (SCSTestException) raised in 5 different places.

mbuechse:
@berendt Hi Christian, can you provide us with a k8s cluster for testing this?

Tests/iaas/security-groups/default-security-group-rules.py
mbuechse, line 1: looks like a mistake, this file does not belong to this PR

Tests/iaas/key-manager/check-for-key-manager.py
mbuechse, line 1: this file seems to have been added in error as well

Tests/iaas/security-groups/default-security-group-rules.py
fraugabel: true

Tests/iaas/key-manager/check-for-key-manager.py
fraugabel: strange.. at checkout they should be reseted

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -180,28 +184,139 @@ def check_default_persistentvolumeclaim_readwriteonce(k8s_api_instance, storage_
                     "access mode 'ReadWriteOnce' is not supported",
                     return_code=42,
                 )
+    return 0
```
mbuechse, line 187: What if the persistant volume is not found at all? Shouldn't this trigger an error??

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -180,28 +184,139 @@ def check_default_persistentvolumeclaim_readwriteonce(k8s_api_instance, storage_
...
 
-    # 4. Delete resources used for testing
-    logger.debug(f"delete pod:{pod_name}")
-    api_response = k8s_api_instance.delete_namespaced_pod(pod_name, namespace)
-    logger.debug(f"delete pvc:{pvc_name}")
-    api_response = k8s_api_instance.delete_namespaced_persistent_volume_claim(
-        pvc_name, namespace
-    )
 
+def check_csi_provider(k8s_core_api, allowed_csi_prov=ALLOWED_CSI_PROV):
```
mbuechse, line 190: What is this good for, and how was the list `ALLOWED_CSI_PROV` determined? I can't see any mention of this in the standard nor the testing notes. A comment in the code is required, but also some explanation in the testing notes.

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -180,28 +184,139 @@ def check_default_persistentvolumeclaim_readwriteonce(k8s_api_instance, storage_
...
+            if any(provider in pod.metadata.name for provider in allowed_csi_prov):
+                csi_list.append(pod.metadata.name)
+                logger.info(f"CSI-Provider: {pod.metadata.name}")
+            else:
+                raise SCSTestException(
+                    f"CSI-Provider: {pod.metadata.name} not recommended",
+                    return_code=33,
+                )
+    if not csi_list:
+      logger.info("CSI-Provider: No CSI Provider found.")
```
mbuechse, line 204: Why is this not an error? (I really don't know. Maybe a comment would help)
Also please note that the indentation is wrong here.

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -180,28 +184,139 @@ def check_default_persistentvolumeclaim_readwriteonce(k8s_api_instance, storage_
...
+                    logger.debug(
+                        f"PVC '{self.pvc_name}' exists in namespace '{self.namespace}'"
+                    )
+                    return True
+            return False
+        except ApiException as e:
+            logger.error(f"Error preparing Environment: {e}")
+            return False
+
+    def clean(self):
```
mbuechse, line 256: I think we can get rid of (big parts of) `prepare` as well as the variable `cleanup` by making this method ignore any error that says "resource not found". To me, this would be very much preferable. (It would in principle also reduce the risk of race conditions.)

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -213,55 +328,49 @@ def main(argv):
...
+        return 2
+
+    with TestEnvironment(kubeconfig) as env:
+        # Check if default storage class is defined (MANDATORY)
+        k8s_core_api = env.k8s_core_api
+        logger.debug("check_default_storageclass()")
+        try:
+            default_class_name = check_default_storageclass(env.k8s_storage_api)
+        except SCSTestException:
+            raise
```
mbuechse, line 344:
```suggestion
```

this achieves literally nothing

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -213,55 +328,49 @@ def main(argv):
...
+        except Exception:
+            env.return_code = 1
+            logger.debug("check_default_storageclass() failed")
+            return env.return_code
```
mbuechse, line 348:
```suggestion
            logger.debug("check_default_storageclass() failed")
            return 1
```

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
```diff
@@ -213,55 +328,49 @@ def main(argv):
...
+            env.return_code = create_pvc_pod(k8s_core_api, default_class_name)
+        except ApiException as api_exception:
+            if api_exception.status == 409:
+                logger.info(
+                    "conflicting resources, "
+                    "try to clean up left overs, then start again"
+                )
+                env.return_code = 3
+                env.return_message = "(409) conflicting resources"
+                return
```
mbuechse, line 360: I think the `return` is in error here.

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
fraugabel: its not an error, in case a kind cluster is used

fraugabel: it should simply filter out "wrong" csi providers

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
fraugabel: it should pass on the SCSTestException, i am open for better suggestions

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
fraugabel: this is a good idea

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
mbuechse: Okay, but this doesn't address any of my points.

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
mbuechse: My suggestion is to delete these two lines, because this has the same effect.

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
fraugabel: well, i can catch a "404 resource not found"  now and then skip the clean() execution, but if only one of both (pvc,pod) is missing the other one does not get deleted. so i am thinking, that i should ratehr catch that in the clean() function itself. what do you think?

mbuechse: My proposal was to call `delete_namespace_pod` and `delete_namespaced_persistent_volume_claim` unconditionally, and to swallow any exception that indicates that deletion was unsuccessful because the resource didn't exist in the first place.

Tests/iaas/security-groups/default-security-group-rules.py
mbuechse, line 1: This file seems to be here in error.

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
fraugabel: unfortunately env.return_code has to be set otherwise it's not detected as an error in the exit part.. remember: i need the env.return_code variable everywhere in the context to print it in the report afterwards

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
fraugabel: no it needs the return to get to the __exit__ where the error is handled and printed out (however, this whole exception might be unnedeeded as the conflicted resources is already dealed with in the prepare part )

Tests/kaas/k8s-default-storage-class/k8s-default-storage-class-check.py
fraugabel: i have done so now.. this leads to another problem: when the 404 is triggered, it doesn' t distinguish between the resources that are not found, so if e.g. the pvc is already built and only the pod is not found it will not delete the pvc.. so i will need to establish a flag

PR #747, opened 2024-09-12
Fix golang version in example Dockerfile

michal-gubricky:
While following the [SCS Kaas Sonobuoy test suite and the K8s e2e test framework](https://github.com/SovereignCloudStack/standards/tree/main/Tests/kaas/kaas-sonobuoy-go-example-e2e-framework) guide step by step, I encountered this error::
```
 > [build 7/7] RUN --mount=type=cache,target=/root/.cache/go-build     go test -c -o custom.test ./...:                                                                                  
0.447 # golang.org/x/net/http2                                                                                                                                                           
0.447 /go/pkg/mod/golang.org/x/net@v0.23.0/http2/transport.go:1096:13: tc.NetConn undefined (type *tls.Conn has no field or method NetConn)
0.447 note: module requires Go 1.18
```


anjastrunk:
I'm not competent enough to review this PR. I added @cah-patrickthiem and @tonifinger instead.

Tests/kaas/kaas-sonobuoy-go-example-e2e-framework/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.17-buster as build
+FROM golang:1.18-buster as build
```
tonifinger, line 1:
```suggestion
FROM golang:1.23
```

Since this PR is aimed at updating the Golang version, I would suggest upgrading it to the latest Golang version `1.23`.
Additionally, since we are touching this line, the “as build” could be removed as we do not currently use multi-stage builds.

tonifinger (changes requested):

Tests/kaas/kaas-sonobuoy-go-example-e2e-framework/Dockerfile
michal-gubricky: Sure, I agree. I just upgraded the Go version to the minimum that worked because I thought there were some restrictions on the versions or something like that since `1.17` is a couple of years old and was used in this Dockerfile only a few months ago.

tonifinger (approved):
LGTM

PR #748, opened 2024-09-16
Issue/718 default sg rule

fraugabel:
added alternative test, because the current test only works for the latest versions of the network and compute services, though older versions are not depricated yet


Tests/iaas/security-groups/default-security-group-rules.py
```diff
@@ -47,16 +48,18 @@ def test_rules(cloud_name: str):
...
     else:
         for rule in rules:
-            direction = rule.direction
-            ethertype = rule.ethertype
+            direction = rule["direction"]
+            ethertype = rule["ethertype"]
+            r_custom_sg = rule["used_in_non_default_sg"]
+            r_default_sg = rule["used_in_default_sg"]
             r_custom_sg = rule.used_in_non_default_sg
             r_default_sg = rule.used_in_default_sg
```
mbuechse, line 56: This looks redundant now.

Tests/iaas/security-groups/default-security-group-rules.py
```diff
@@ -82,34 +85,125 @@ def test_rules(cloud_name: str):
...
+    sg = conn.network.create_security_group(name=sg_name, description=description)
+    return sg.id
+
+
+def delete_security_group(conn, sg_id):
+    conn.network.delete_security_group(sg_id)
+    try:
+        conn.network.find_security_group(name_or_id=sg_id)
+    except Exception:
+        print(f"Security group {sg_id} was deleted successfully.")
```
mbuechse, line 130: This cannot be right.

Tests/iaas/security-groups/default-security-group-rules.py
```diff
@@ -82,34 +85,125 @@ def test_rules(cloud_name: str):
...
-        "Ingress Rules": ingress_rules,
-        "Egress Rules": egress_rules
-    }
+    result_dict = {"Ingress Rules": ingress_rules, "Egress Rules": egress_rules}
+    return result_dict
+
+
+def create_security_group(
+    conn, sg_name: str = "default-test-sg", description: str = "default-test-sg"
+):
```
mbuechse, line 111: The only added value of this function is the default values for the arguments?

mbuechse (changes requested):
Error handling should be improved (don't just swallow exceptions by printing something to stdout), assert statements should not be used in production, and I suppose the code can be made less redundant and more concise.

mbuechse:
BTW, I noticed that the test script is incredibly noisy when --debug is used. Okay, probably fine. However, logging is not initialized, so the channel is not included in the log line. This should be changed -- please include the following line:

```python
    logging.basicConfig(format='%(levelname)s: %(message)s', level=logging.INFO)
```

And if `--debug` is supplied, the following must be done:

```python
    logging.getLogger().setLevel(logging.DEBUG)
```

Last not least, the output of the script is not usable in any way. What we need is the following:

- for every violation of a requirement: output a message on the ERROR channel
- for every violation of a recommendation: output a message on the WARNING channel
- the return code should be non-zero precisely when an ERROR had been emitted

Every other output should be on the debug channel.

Tests/iaas/security-groups/default-security-group-rules.py
fraugabel: i turned the default values into globals, now or would you recommend to get rid of this function completely?

Tests/iaas/security-groups/default-security-group-rules.py
fraugabel: of course, you're right.

Tests/iaas/security-groups/default-security-group-rules.py
```diff
@@ -128,7 +126,7 @@ def delete_security_group(conn, sg_id):
     try:
         conn.network.find_security_group(name_or_id=sg_id)
     except Exception:
-        print(f"Security group {sg_id} was deleted successfully.")
+        print(f"Security group {sg_id} was not deleted successfully.")
```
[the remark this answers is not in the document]
fraugabel: a yes thanks.. &yes, when i coded it it made total sense to have the exception thrown in the success-case, but looking at it a month later.. was confusing (maybe i should add a comment here) and you are tottaly right, the exception must be very specific in that case

Tests/iaas/security-groups/default-security-group-rules.py
fraugabel: i think i will port the whole part to a function

mbuechse (approved):
I massaged it a bit. LGTM -- please check whether it looks good to you, too.

PR #750, opened 2024-09-16
Drop redundant copying of checkout, run in pod

mbuechse:


gtema (approved):
Looks good, at least check-adr job is working fine as seen in results

gtema (approved):

PR #751, opened 2024-09-16
remove unused table subject

mbuechse:


compliance-monitor/monitor.py
```diff
@@ -331,8 +328,8 @@ def get_scopes():
             _scopes[ident] = current = {'_counter': -1}
     if current['_counter'] != counter:
         current.clear()
-        import_cert_yaml_dir(yaml_path, current)
         current['_counter'] = counter
+        import_cert_yaml_dir(yaml_path, current)
```
martinmo, line 332: @mbuechse An (unrelated?) fix sneaked in here, please double check if that's wanted. But it looks like a valid bugfix to me.

martinmo (approved):

compliance-monitor/monitor.py
mbuechse: Yes, that bugfix is intended. It sneaked in here like you say, and I didn't want to spend the time to untangle it.

PR #752, opened 2024-09-17
Stabilize Volume Backup Functionality standard

markus-hentsch:


josephineSei (approved):
If everyone is fine, we may stabilize it.
But maybe we should also ask, if CSPs need more time to adopt this?

mbuechse:
At any rate, we should handle https://github.com/SovereignCloudStack/standards/issues/795 first.

PR #756, opened 2024-09-19
add endpoints to render certificate scopes

mbuechse:


mbuechse:
Screenshot from 2024-09-19 20-58-51

almost the same as https://docs.scs.community/standards/scs-compatible-iaas

only that the former shows more versions, and this is better fixed by changing https://github.com/SovereignCloudStack/standards/blob/main/Tests/scs-compatible-iaas.yaml (remove `v3-orig` from latest timeline entry)

mbuechse:
This can be reviewed, I mainly need to update README.md

compliance-monitor/README.md
```diff
@@ -209,13 +209,24 @@ Needs to be authenticated (via basic auth).
 
 Supports content type `text/plain; version=0.0.4; charset=utf-8` only.
 
-### GET /pages
+### GET /{view_type}/table\[_full\]
 
-Returns the compliance table for all active subjects (type `text/html`).
+Returns the compliance table for all active subjects, where `view_type` can be
```
cah-patrickthiem, line 214: Is anything missing at the end of the line?

compliance-monitor/monitor.py
```diff
@@ -569,7 +575,7 @@ async def get_detail(
         rows2, get_scopes(), grace_period_days=GRACE_PERIOD_DAYS,
         subjects=(subject, ), scopes=(scopeuuid, ),
     )
-    return render_view(VIEW_DETAIL, view_type, results2, base_url=settings.base_url, title=f'{subject} compliance')
+    return render_view(VIEW_DETAIL, view_type, results=results2, base_url=settings.base_url, title=f'{subject} compliance')
```
cah-patrickthiem, line 580: wouldn't it be better to directly name this variable results instead of overwriting it?

cah-patrickthiem (commented):
Looks good overall, I still have some minor complaints or questions.

compliance-monitor/monitor.py
mbuechse: The variable can be renamed; the name was necessary in the past, but it isn't anymore.
However, the `results=results2` (or then `results=results`) will be necessary anyway, because it must be a keyword argument, so that `render_view` can propagate the argument to the call of `render` (line 556).

cah-patrickthiem (approved):
Thanks, looks good now.

PR #759, opened 2024-09-25
Add cluster hardening tests

michal-gubricky:
This PR adds tests for K8s cluster hardening
Fixes https://github.com/SovereignCloudStack/issues/issues/670


cah-patrickthiem:
I started reviewing this PR some days ago and some tests are failing every time I try unfortunately. Even when I try different kind configurations.
failing tests:
- Authentication_Methods
- K8s_Endpoints_HTTPS
- Control_Plane_Ports_Security

I am not sure why, maybe a special cluster config is needed, at least they should work for you, so can you share your cluster configuration please?

cah-patrickthiem:
> I started reviewing this PR some days ago and some tests are failing every time I try unfortunately. Even when I try different kind configurations. failing tests:
> 
> * Authentication_Methods
> * K8s_Endpoints_HTTPS
> * Control_Plane_Ports_Security
> 
> I am not sure why, maybe a special cluster config is needed, at least they should work for you, so can you share your cluster configuration please?

Hm, actually the kind config file is already there. So I am more confused why this is not working for me. 
Can you please check that said tests are really working on your side?

michal-gubricky:
> > I started reviewing this PR some days ago and some tests are failing every time I try unfortunately. Even when I try different kind configurations. failing tests:
> > 
> > * Authentication_Methods
> > * K8s_Endpoints_HTTPS
> > * Control_Plane_Ports_Security
> > 
> > I am not sure why, maybe a special cluster config is needed, at least they should work for you, so can you share your cluster configuration please?
> 
> Hm, actually the kind config file is already there. So I am more confused why this is not working for me. Can you please check that said tests are really working on your side?

Sorry for the confusion. I would say tests are working but these 3 are not passing due to the current KinD setup. If it is mandatory that all tests need to be passed against KinD then we need to adjust the config of the kind cluster.

- Authentication_Methods
  - problem here is that based on standard:
   ```
   cluster MUST implement at least two methods for authentication. One of these MUST be Service Account Tokens,
   in order to provide full functionality to Pods. A second authentication mechanisms can be chosen depending on the 
   requirements of the provider and/or customer. 
   ```
   As the second it can be OIDC or token webhook, so If it is possible I can try to create some fake second authentication method if it is all right.
- K8s_Endpoints_HTTPS
  - here is the problem that I list all the k8s endpoints and then checks if they are secured via HTTPS like standard say. So should we just test k8s api endpoint or? 
- Control_Plane_Ports_Security
  - based on standard, It MUST NOT be possible to access Kubernetes ports from outside the internal network hosting the Kubernetes cluster. Do you have some idea how to handle it?

cah-patrickthiem:
> > > I started reviewing this PR some days ago and some tests are failing every time I try unfortunately. Even when I try different kind configurations. failing tests:
> > > 
> > > * Authentication_Methods
> > > * K8s_Endpoints_HTTPS
> > > * Control_Plane_Ports_Security
> > > 
> > > I am not sure why, maybe a special cluster config is needed, at least they should work for you, so can you share your cluster configuration please?
> > 
> > 
> > Hm, actually the kind config file is already there. So I am more confused why this is not working for me. Can you please check that said tests are really working on your side?
> 
> Sorry for the confusion. I would say tests are working but these 3 are not passing due to the current KinD setup. If it is mandatory that all tests need to be passed against KinD then we need to adjust the config of the kind cluster.
> 
> * Authentication_Methods
>   
>   * problem here is that based on standard:
>   
>   ```
>   cluster MUST implement at least two methods for authentication. One of these MUST be Service Account Tokens,
>   in order to provide full functionality to Pods. A second authentication mechanisms can be chosen depending on the 
>   requirements of the provider and/or customer. 
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   As the second it can be OIDC or token webhook, so If it is possible I can try to create some fake second authentication method if it is all right.
> * K8s_Endpoints_HTTPS
>   
>   * here is the problem that I list all the k8s endpoints and then checks if they are secured via HTTPS like standard say. So should we just test k8s api endpoint or?
> * Control_Plane_Ports_Security
>   
>   * based on standard, It MUST NOT be possible to access Kubernetes ports from outside the internal network hosting the Kubernetes cluster. Do you have some idea how to handle it?

Alright, I see, I will try to test again against a real K8s cluster.

Tests/kaas/kaas-sonobuoy-tests/scs-conformance-sonobuoy-plugin.yaml
```diff
@@ -7,7 +7,7 @@ spec:
   command:
   - bash
   args: ["-c","go tool test2json ./custom.test -test.v | tee ${SONOBUOY_RESULTS_DIR}/out.json ; echo ${SONOBUOY_RESULTS_DIR}/out.json > ${SONOBUOY_RESULTS_DIR}/done"]
-  image: ghcr.io/sovereigncloudstack/standards/scs-kaas-conformance:dev
+  image: registry.dnation.cloud/test-mg/scsconformance:dev #ghcr.io/sovereigncloudstack/standards/scs-kaas-conformance:dev
```
tonifinger, line 10: It seems to me that this is a development artifact. Could you revert these changes?

michal-gubricky: That is true, I will change it.

cah-patrickthiem:
Some implementation notes for things I needed to configure:

For Test "Test_scs_0217_sonobuoy_PodSecurity_Standards_And_Admission_Controller_Enabled":

Two things needed to be changed in the cluster:


Enable PodSecurity admission controller:
```
yamlCopy# Edit /etc/kubernetes/manifests/kube-apiserver.yaml
spec:
  containers:
  - command:
    - kube-apiserver
    - --enable-admission-plugins=NodeRestriction,PodSecurity  # Add PodSecurity
```

Add enforce policy annotations to namespaces:
```
bashCopy# For each namespace (replace <namespace>):
kubectl annotate namespace <namespace> \
    pod-security.kubernetes.io/enforce=baseline \
    pod-security.kubernetes.io/enforce-version=latest
```
For system namespaces, restrict the following:
```
kubectl annotate namespace kube-system \
    pod-security.kubernetes.io/enforce=restricted \
    pod-security.kubernetes.io/enforce-version=latest
```

Enable Pod Security:
```vi enable-pod-security.sh
chmod +x enable-pod-security.sh
```
(See file content below)
```
sudo ./enable-pod-security.sh
```
```
#enable-pod-security.sh
#!/bin/bash

# Use correct kubeconfig
export KUBECONFIG=/etc/kubernetes/admin.conf

# Enable PodSecurity admission plugin
APISERVER_MANIFEST="/etc/kubernetes/manifests/kube-apiserver.yaml"
sed -i '/--enable-admission-plugins/s/NodeRestriction/NodeRestriction,PodSecurity/' $APISERVER_MANIFEST

# Wait for API server to be ready
echo "Waiting for API server to be ready..."
until kubectl get nodes &>/dev/null; do
    sleep 5
done
echo "API server is ready"

# Add enforce policy to namespaces
for ns in $(kubectl get namespaces -o jsonpath='{.items[*].metadata.name}'); do
    echo "Setting pod security policy for namespace: $ns"
    if [[ "$ns" == "kube-system" ]]; then
        kubectl annotate namespace $ns \
            pod-security.kubernetes.io/enforce=restricted \
            pod-security.kubernetes.io/enforce-version=latest --overwrite
    else
        kubectl annotate namespace $ns \
            pod-security.kubernetes.io/enforce=baseline \
            pod-security.kubernetes.io/enforce-version=latest --overwrite
    fi
done
```

For Test "Test_scs_0217_sonobuoy_Authentication_Methods", I needed to use another auth method.
Enable OIDC authentication in kube-apiserver config:
```
# /etc/kubernetes/manifests/kube-apiserver.yaml
- --oidc-issuer-url=https://localhost:8443
- --oidc-client-id=kubernetes
- --oidc-username-claim=sub
```

Tests/kaas/kaas-sonobuoy-tests/scs_k8s_conformance_tests/scs_0217_cluster_hardening_test.go
```diff
@@ -0,0 +1,825 @@
...
+							}
+						}
+					}
+				}
+			}
+			return ctx
+		})
+
+	testenv.Test(t, f.Feature())
+}
```
cah-patrickthiem, line 179: In order to make this test work on my cluster,  needed to change a few things:
Added required HTTPS ports map:
```
goCopyrequiredHTTPSPorts := map[string]bool{
    apiServerPort:         true,
    controllerManagerPort: true, 
    schedulerPort:         true,
    kubeletApiPort:        true,
}
```
This ensures only the ports are checked that must use HTTPS per the standard.

Added etcd ports dynamically:
```
goCopyfor port := etcdPortRangeStart; port <= etcdPortRangeEnd; port++ {
    requiredHTTPSPorts[strconv.Itoa(port)] = true
}
```
Handles the etcd port range (2379-2380).

Modified endpoint check loop to only test required ports:
```
goCopyportStr := strconv.Itoa(int(port.Port))
if requiredHTTPSPorts[portStr] {
    // HTTPS check
}
```
This excludes ports like DNS (53) and metrics (9153) that don't require HTTPS.

Fixed context usage (not needed for passing the test, but TODO should now be there):
```
goCopycontext.Background() instead of context.TODO()
```
My suggestion (complete code):
```
// Test_K8s_Endpoints_HTTPS checks if all Kubernetes endpoints are secured via HTTPS.
func Test_scs_0217_sonobuoy_K8s_Endpoints_HTTPS(t *testing.T) {
   f := features.New("Kubernetes endpoint security").Assess(
       "All Kubernetes endpoints should be secured via HTTPS",
       func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
           requiredHTTPSPorts := map[string]bool{
               apiServerPort:         true,
               controllerManagerPort: true, 
               schedulerPort:         true,
               kubeletApiPort:        true,
           }

           // Add etcd ports
           for port := etcdPortRangeStart; port <= etcdPortRangeEnd; port++ {
               requiredHTTPSPorts[strconv.Itoa(port)] = true
           }

           endpoints := &corev1.EndpointsList{}
           if err := cfg.Client().Resources().List(context.Background(), endpoints); err != nil {
               t.Fatal("failed to list endpoints:", err)
           }

           for _, ep := range endpoints.Items {
               for _, subset := range ep.Subsets {
                   for _, addr := range subset.Addresses {
                       for _, port := range subset.Ports {
                           portStr := strconv.Itoa(int(port.Port))
                           if requiredHTTPSPorts[portStr] {
                               if isPortSecuredWithHTTPS(addr.IP, port.Port, connectionTimeout) {
                                   t.Logf("Endpoint %s:%d (%s) is secured via HTTPS", addr.IP, port.Port, port.Name)
                               } else {
                                   t.Errorf("Error: Endpoint %s:%d (%s) is not secured via HTTPS", addr.IP, port.Port, port.Name)
                               }
                           }
                       }
                   }
               }
           }
           return ctx
       })
   testenv.Test(t, f.Feature())
}
```

Tests/kaas/kaas-sonobuoy-tests/scs_k8s_conformance_tests/scs_0217_cluster_hardening_test.go
```diff
@@ -0,0 +1,825 @@
...
+			if err != nil {
+				t.Fatal("failed to create rest config:", err)
+			}
+
+			kubeClient, err := kubernetes.NewForConfig(restConf)
+			if err != nil {
+				t.Fatal("failed to create Kubernetes client:", err)
+			}
+
+			nodeList, err := kubeClient.CoreV1().Nodes().List(context.TODO(), v1.ListOptions{})
```
cah-patrickthiem, line 197: there are several places where "context.TODO()" is used. Please change that to "context.Background()".
That should work everywhere and is not as "ugly" as TODO

Tests/kaas/kaas-sonobuoy-tests/scs_k8s_conformance_tests/scs_0217_cluster_hardening_test.go
```diff
@@ -0,0 +1,825 @@
...
+	address := net.JoinHostPort(ip, port)
+	conn, err := net.DialTimeout("tcp", address, timeout)
+	if err != nil {
+		// Connection failed, the port is likely closed
+		return false
+	}
+	// Connection succeeded, the port is open
+	conn.Close()
+	return true
+}
```
cah-patrickthiem, line 518: There are some helper functions, please group them and put them in the beginning (please group the tests accordingly). That way, the tests are a bit more clearly structured. For reference see: https://github.com/SovereignCloudStack/standards/pull/714/files

cah-patrickthiem (commented):
Overall great work! Please address my comments and we can merge it soon!

michal-gubricky:
> Overall great work! Please address my comments and we can merge it soon!

Thanks for the review, I will try to address the comments as soon as possible.

Tests/kaas/kaas-sonobuoy-tests/scs_k8s_conformance_tests/scs_0217_cluster_hardening_test.go
michal-gubricky: see 5ed2bfa9b43e1ae4d598d8576c7f5d79096c5fb2

Tests/kaas/kaas-sonobuoy-tests/kind_config.yaml
```diff
@@ -3,6 +3,18 @@ apiVersion: kind.x-k8s.io/v1alpha4
...
+    apiVersion: kubeadm.k8s.io/v1beta3
+    kind: ClusterConfiguration
+    metadata:
+      name: config
+    apiServer:
+      extraArgs:
+        enable-admission-plugins: "NodeRestriction,PodSecurity"
+        oidc-issuer-url: https://localhost:8443
+        oidc-client-id: kubernetes
+        oidc-username-claim: sub
```
cah-patrickthiem, line 17: unfortunately since kind and kubeadm are different things, we cannot leave the configuration here.
Imo this needs to be addressed in the implementation node section in another issue/PR. Please exclude it here.

cah-patrickthiem (changes requested):
Please address the last minor thing, after that LGTM.

Tests/kaas/kaas-sonobuoy-tests/kind_config.yaml
michal-gubricky: I included these settings because they are necessary to ensure that certain tests pass against kind. However, if you'd prefer to address this configuration in the implementation node section in another PR/issue, I will remove it from this PR as you suggested.

cah-patrickthiem (approved):

PR #762, opened 2024-09-25
Only debug output the public/community images.

garloff:
This avoids confusion by seeing images that are private. Like in the mandatory image test, this can be overriden with -p/--private.


garloff:
I had images like `EVIL` registered to test vulnerabilities in my environment.
When I saw the compliance test suite reporting their existence I also expected that I would fail, as these do not have the necessary metadata ...
It did not ... in the end the tool listing all images was the one that looked for mandatory standard images.
The tool looking for metadata does disregard private images by default.
So let's align the behavior: Let's look only at "community" and "public" images in both tools by default and let's support `-p` and `--private` in both tools to override this restriction.

mbuechse (commented):
I propose to do https://github.com/SovereignCloudStack/standards/pull/766 instead.

PR #764, opened 2024-09-26
Add current generation CPUs and GPUs.

garloff:
So we add AMD Zen-5/5c, intel Sierra Forest (Crestmont) and Granite Rapids (Redwood Cove), ARM AmpereOne and Cortex-72x/NeoverseN3/V3.

On the GPU side, add nVidia Hopper (h), C/RDNA 3.5 (3.5) and C/RDNA4 (4), intel Xe12.2 (2) and Arc/12.7/DG2 (3).

Added it to the spec and the code.
Spec does not get a new revision number, as this is a fully backwards- compatible enahncement (and logical continuation). We could of course update the minor version (and call this v3.2).


mbuechse (approved):
To be frank, this seems a bit messy to me, which is a testament to the fact that flavor naming is a crutch, but I won't object because that seems to be the state of the art for better or worse...

Issue #749, opened 2024-09-16
Define security of iaas service software


josephineSei:
In https://github.com/SovereignCloudStack/issues/issues/528 we define sets of mandatory and supported APIs.
This is a lengthy discussion.

But we may not only define, what APIs we expect to see in a cloud but also how old they are allowed to be.
On the KaaS Layer there is already a standard: https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0210-v1-k8s-new-version-policy.md , which advises to closely follow the K8s releases.

We need to also define a period of Liveness for the IaaS Layer.

## Definition of Done:

Please refer to [scs-0001-v1](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0001-v1-sovereign-cloud-standards.md) for details.

- [ ] Proposal has been written with name of the form `scs-xxxx-v1-slug.md` (only substitute `slug`)
- [ ] Proposal has the fields `status`, `type`, `track` set
- [ ] Proposal has been voted upon in the corresponding team
- [ ] Status has been changed into `Draft`, file renamed: `xxxx` replaced by document number
- [ ] If applicable: test script has been written (this item may be moved into a separate issue so long as the state is `Draft`)

josephineSei:
The question here is: How can we determine, which version of the IaaS-API is used? And how many version we want to support on IaaS-Level.

1. One Option would be to allow all IaaS APIs, that are still maintained (e.g. OpenStack services, that are still maintained and not in EOL or Unmaintained: https://releases.openstack.org/).
2. Another Option would be to also allow Unmaintained release, when CSPs will A) fix all security issues themself and B) maybe have an upgrade plan to finally rely on maintained IaaS services.

Testing this will be quite tricky.

We could use the `openstack version show`:
```
stack@devstack:~/devstack$ openstack versions show
+-------------+----------------+---------+-----------+---------------------------------------------+------------------+------------------+
| Region Name | Service Type   | Version | Status    | Endpoint                                    | Min Microversion | Max Microversion |
+-------------+----------------+---------+-----------+---------------------------------------------+------------------+------------------+
| RegionOne   | placement      | 1.0     | CURRENT   | http://192.168.23.161/placement/            | 1.0              | 1.39             |
| RegionOne   | compute        | 2.0     | SUPPORTED | http://192.168.23.161/compute/v2/           | None             | None             |
[...]

PR #765, opened 2024-09-30
Create a standard for the security of iaas service software

josephineSei:
closes https://github.com/SovereignCloudStack/standards/issues/749


josephineSei:
We discussed today in the IaaS meeting, how to proceed.
The Conclusions were:
- expecting security patches to be deployed is reasonable for *SCS-Compatible*, but may be not easily possible to test
    -> this would need self-reports from CSPs in the reasonable timeframe, that they have updated their deployments
- SBOMs are part of *SCS-Open* and will give hints on both: overall software version and security patches
- behavior changes and/or must-have new API-endpoints should be done through standards with reasonable timeframe to become "mandatory"
    -> this means that a CSP will have to look through the standards and check, whether their used software version fulfill all standards
    -> the standards will have to be adjusted ob a regular basis
        -> we should think about also enabling another test for future requirements of standards (this means nearly all SHOULDS !)

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,109 @@
...
+Instead of requiring a defined minimum software version, it could be derived from the standards.
+Because: Whenever there is a new wanted behavior a standard should be created and a resonable timeframe given to CSPs to adopt a software version that can fulfill the new standard.
+Through the combination of all standards that are in place, the minimum version for the IaaS service software is implicitly given.
+
+This would avoid to have conflicting versions of software in terms of feature parity, while also allowing older software.
+Using this approach requires an additional advise to CSPs to update or implement patches for security issues.
+
+#### Advise CSPs to integrate software updates
+
+As long as maintained versions of software is used, updates with security patches are available and only need to be integrated.
```
markus-hentsch, line 79:
```suggestion
As long as maintained versions of software are used, updates with security patches are available and only need to be integrated.
```

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,109 @@
...
+This can and should be done in a reasonable short timeframe.
+
+But CSPs may even use releases of IaaS software, that are either not maintained anymore by an open source community or may be even closed source implementations of the mandatory IaaS APIs.
+Allowing older versions or closed source software would only be acceptable, when CSPs assure (e.g. in documentation), that they themself will patch the software within their deployments.
+Security bug fixes must be implemented and proof of the fix then provided.
+Only under that circumstances deployments with older or alternative IaaS Layer software, may be handled as compliant.
```
markus-hentsch, line 85:
```suggestion
Only under these circumstances deployments with older or alternative IaaS Layer software may be handled as compliant.
```

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,109 @@
...
+
+This option could be taken for granted, but to actually advise using it may encourage CSPs to take a closer look on their life-cycle management and security risk handling.
+And CSPs using OpenStack could even be encouraged to upgrade their deplyoments.
```
markus-hentsch, line 88:
```suggestion
And CSPs using OpenStack could even be encouraged to upgrade their deployments.
```

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Standard for a minimum IaaS Layer Software version
+
+If a maintained[^1] version of OpenStack is used as implementation for IaaS Layer software, security patches noted in OSSNs and OSSAs MUST be integrated within a reasonable timeframe.
```
markus-hentsch, line 92: OSSN and OSSA are missing from the glossary.

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,109 @@
...
+Otherwise the CSP MUST implement security bug fixes themself within a reasonable timeframe.
```
markus-hentsch, line 93: This standard does not seem to define "reasonable timeframe" and as such the "MUST" condition is hard to validate either on CSP and SCS side. I think the timeframe matter could be part of the design considerations and discussed in more detail there.

Standards/scs-XXXX-w1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,40 @@
...
+It is not always possible to automatically test, whether the software has the newest security updates.
+This is because software versions may differ or some CSPs might have added downstream code parts or using other software than the reference.
+Also vulnerabilites and their fixes are quite different in testing, some might not be testable while others are.
+Additionally testing might be perceived as an attack on the infrastructure.
+So this standard will rely on the work and information CSPs must provide.
+There are different cases and procedures which are addressed in the following parts, that lead to compliance for this standard.
+
+### Procedure to become compliant to the security of IaaS service software Standard
+
+This is the procedure when a new deployment wants to achieve scs-conformancy.
```
markus-hentsch, line 21:
```suggestion
This is the procedure when a new deployment wants to achieve SCS-conformancy.
```

Standards/scs-XXXX-w1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,40 @@
...
+There are two states such a deployment can be in:
+
+1. When a deployment is newly build or installed it usually uses software which includes all the latest security and bug fixes.
+Such deployments should be considered compliant to the standard.
+
+2. When a CSP wants to make an older deployment compliant to the scs standards and thus also to this standard, it should be checked, whether the running software is up to date.
```
markus-hentsch, line 27:
```suggestion
2. When a CSP wants to make an older deployment compliant to the SCS standards and thus also to this standard, it should be checked, whether the running software is up to date.
```

Standards/scs-XXXX-w1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,40 @@
...
+Any updates or upgrades to even newer versions should be done before the scs-compliance is checked.
```
markus-hentsch, line 28:
```suggestion
Any updates or upgrades to even newer versions should be done before the SCS compliance is checked.
```

Standards/scs-XXXX-w1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,40 @@
...
+Information about such updates or upgrade should be provided by the CSP.
+
+### Procedure when new Vulnerabilites are discovered
```
markus-hentsch, line 31:
```suggestion
### Procedure when new vulnerabilites are discovered
```

Standards/scs-XXXX-w1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,40 @@
...
+
+Whenever there are new vulnerabilities discovered in IaaS service software like OpenStack there is either an internal discussion ongoing or it is just a smaller issue.
+In the first case CSPs should have someone following such discussions and may even help preparing and testing patches.
+From the moment on the vulnerability is announced it will be used by attackers.
```
markus-hentsch, line 35:
```suggestion
From the moment on the vulnerability is disclosed publicly, the risk of it being actively exploited increases greatly.
```

Standards/scs-XXXX-w1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,40 @@
...
+So CSPs MUST watch out for announcements like in the OSSAs and OSSNs and when they are affected, update their deployment as soon as possible.
```
markus-hentsch, line 36: Should we maybe consider cases here where vulnerabilities are not applicable to certain deployments? E.g. vulnerability concerns storage backend driver A whereas CSP is exclusively using storage backend driver B.
In such case, the update would not be a MUST for the CSP, in my opinion.

We should maybe put a preceeding step here to instruct the CSP to do an analysis first ASAP whether they are affected and if ...

- ... they are affected, they MUST apply the update and report having done so to SCS
- ... they are not affected, they MAY apply the update but MUST report a reasoning why they are not affected to SCS

markus-hentsch (changes requested):

markus-hentsch:
Good first draft! I added some minor spelling/phrasing correction suggestions and some questions.
I think the topic is not easy and we will need to be careful about two matters in particular:

1. the timeframe/deadline imposed on CSPs to apply updates
2. the way CSPs report back to SCS/OSBA about their compliance concerning software versions

No. 1 needs to be a sweet spot between giving CSPs enough room to implement/test and being strict enough to ensure security in a meaningful way.

Regarding no. 2 we maybe could be a bit more precise or provide a template. Unless I have missed something this could currently range from hand-written letters to SBOM XML files that are gigabytes in size which no human can look through in a timely manner.

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
josephineSei: I added them.

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,108 @@
...
+| CSP                 | Cloud Service Provider, provider managing the OpenStack infrastructure.                                                                  |
+| SLURP               | Skip Level Upgrade Release Process - A Process that allows upgrades between two releases, while skipping the one in between them.        |
+| OSSN                | [OpenStack Security Note](https://wiki.openstack.org/wiki/Security_Notes) - security issues from 3rd parties or due to misconfigurations. |
+| OSSA                | [OpenStack Security Advisories](https://security.openstack.org/ossalist.html) - security issues and advices for OpenStack.               |
+
+## Motivation
+
+In software projects like e.g. OpenStack the software will be modified and receive bug fixes continuously and will receive releases of new versions on a regular basis.
+Older releases will at some point not receive updates anymore, because maintaining more and more releases simultaneously requires too much manpower.
+Thus older software will also eventually not receive security updates anymore.
```
anjastrunk, line 28:
```suggestion
Thus older versions will also eventually not receive security updates anymore.
```

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,108 @@
...
+Using software which does not receive updates anymore threatens the baseline security of deployments and should be avoided under all circumstances.
```
anjastrunk, line 29:
```suggestion
Using versions which does not receive updates anymore threatens the baseline security of deployments and should be avoided under all circumstances.
```

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,108 @@
...
+Because: Whenever there is a new wanted behavior a standard should be created and a resonable timeframe given to CSPs to adopt a software version that can fulfill the new standard.
+Through the combination of all standards that are in place, the minimum version for the IaaS service software is implicitly given.
+
+This would avoid to have conflicting versions of software in terms of feature parity, while also allowing older software.
+Using this approach requires an additional advise to CSPs to update or implement patches for security issues.
+
+#### Advise CSPs to integrate software updates
+
+As long as maintained versions of software are used, updates with security patches are available and only need to be integrated.
+This can and should be done in a reasonable short timeframe.
```
anjastrunk, line 79: What is a reasonable time? IMO, standard should be explicit here in order to guide CSPs.

How does customers know, whether or not a specific bug is fixed? Do we "force" CSPs to document software version/patch version somewhere?

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,108 @@
...
+Allowing older versions or closed source software would only be acceptable, when CSPs assure (e.g. in documentation), that they themself will patch the software within their deployments.
+Security bug fixes must be implemented and proof of the fix then provided.
+Only under these circumstances deployments with older or alternative IaaS Layer software may be handled as compliant.
+
+This option could be taken for granted, but to actually advise using it may encourage CSPs to take a closer look on their life-cycle management and security risk handling.
+And CSPs using OpenStack could even be encouraged to upgrade their deployments.
+
+## Standard for a minimum IaaS Layer Software version
+
+If a deployment is affected by a security issue and a maintained[^1] version of OpenStack is used as implementation for IaaS Layer software, security patches noted in OSSNs and OSSAs MUST be integrated within a reasonable timeframe.
```
anjastrunk, line 91: What is a reasonable time? IMO, standard should be explicit here in order to guide CSPs.

How does customers know, whether or not a specific bug is fixed? Do we "force" CSPs to document software version/patch version somewhere?

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,108 @@
...
+Otherwise the CSP MUST implement security bug fixes themself within a reasonable timeframe, when the deplyoment is affected by a security issue.
+
+In both cases proof of the update MUST be send to the OSBA, so that the compliance will not be revoked.
```
anjastrunk, line 94: I'm afraid, this will result in a lot of paper work. Isn't there any possibility to get software version, automatically?

anjastrunk:
I had a look to this standard, too. IMO the standard should dictate patch time explicitly, such as two weeks. Describing time passed between publishing a patch and applying it, should not be described softly, such as "very soon", "as soon as possible". The standards does not bring any value-added to customers, as everybody interprets these words differently. 

Secondly, I'm not convinced of the process a CSP uses to proof a certain vulnerability is fixed. It looks like manual steps are required, which are error-prone and will delay process additionally. I prefer to develop an automatic process, e.g. using digital signatures.

anjastrunk (changes requested):
I had a look to this standard, too. IMO the standard should dictate patch time explicitly, such as two weeks. Describing time passed between publishing a patch and applying it, should not be described softly, such as "very soon", "as soon as possible". The standards does not bring any value-added to customers, as everybody interprets these words differently.

Secondly, I'm not convinced of the process a CSP uses to proof a certain vulnerability is fixed. It looks like manual steps are required, which are error-prone and will delay process additionally. I prefer to develop an automatic process, e.g. using digital signatures.

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
josephineSei: That really is a problem:
We want to create standards that should be followed. Therefore we need tests.
But future CVEs or OSSNs cannot be known now and may be not easily testable. Should we consider writing a test for each of these vulnerabilities? This might be several tests a year to write, if a test from the outside is possible at all. And those tests might be considered as attacks (I mean they somehow are ;) ) and might trigger responses on the side of CSPs.

So a CSP giving notice about the fixing of a vulnerability, which already implicates, that we need to trust the CSP, could be an option. Maybe - if we already need to trust the CSPs - we may reduce the paper work:
MArkus suggested that there should be some kind of form file, that could be provided. We could do something like:

```
How did you deal with the vulnerability on your deployment XYZ:
[ ] Deployment was NOT affected, because ....... (e.g. we use another backend)
[ ] Functionality was disabled in the deployment.
[ ] Deployment was updated with a fix for the vulnerability, that .... (e.g. was provided be upstream openstack)
```

And only require short descriptions like above.
This may even be easier to evaluate in the compliance checker?

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
josephineSei: I can only propose a timeframe, but this need to be discussed with CSPs. I think i will attend the lean operator coffee next monday to ask for their opinion.

josephineSei:
> For now I just want to comment on something I don't see explained or discussed neither in the proposal nor in any of the current commentary:
> 
> This standard - in it's current form - talks broadly about "(the) software" comprising the IaaS layer, and I assume this mostly is meant to mean Openstack software.
> 
Indeed this standard should be applied on the IaaS Level, meaning all services with user-accesible APIs on that layer (this includes OpenStack and S3 if used).

Nevertheless we definitely should talk about security on all layers from Metal to KaaS. We may discuss this in the IAM call next week. But I think putting all of it into a single standard may be too complex?

> However I would find it important that we clarify this, because a typical IaaS Deployment is comprised of the following - very much different - software layers, especially when it comes to security updates.
> 
>     * Firmware needed for bare metal components, e.g. NIC and SSD firmware, CPU microcode
> 
>     * basic operating system software needed to boot and manage hardware, e.g. grub bootloader, linux kernel, initrd etc.
> 
>     * basic operating system - usually linux - packages to support a base layer of software and to start the next layer: systemd, glibc, a package manager (apt/dpkg), network configuration software (Netplan), a container runtime (docker, podman, k8s), others
> 
>     * some kind of container, which often contains it's own layer of basic operating system packages again (see above point), minus the container runtime stuff itself
> 
>     * the actual openstack services, usually python files (these are _mostly_ covered by the openinfra security advisories/ OSSAs)
> 
>     * direct openstack service dependencies also developed under the openinfra umbrella - also covered by the openinfra security process, e.g. oslo libraries
> 
>     * (in)direct openstack service dependencies developed outside openinfra security process, e.g. docker-py, various low level python librariers, like python-ldap (for keystone)
> 
>     * adjacent services deployed as containers alongside openstack infrastructure, e.g. logging and monitoring (fluentd, prometheus, ELK stack, keycloak) which are also not part of the openinfra security process
> 
> 
> So I think it is important to at least distinguish which software from the above list the CSP should patch. If the answer to this is: "all of it" then we must be clear that only a tiny fraction of it is covered by openinfra security processes!
> 
This is a good point that needs to be integrated into the standard regardless of how detailed the standard itself is.
> e.g. even direct dependencies of core openstack services are actually not monitored by openinfra for security issues if these are not developed under the openinfra umbrella!
> 
> of course, if issues are reported these are fixed, but e.g. the requirements files listing tested versions which are known to work with various openstack services are no indicator that these releases are secure!
> 
> So I think this is important to highlight to the user. This is the reason upstream provided kolla docker images are only provided for testing purposes and not for production, because these are not regularly updated for security fixes.
> 
> So imho we should keep in mind that CSPs need to monitor way more than just OSSA announcements if they want to be secure.
> 
> Thanks

Thank you for your feedback. I will integrate the focus on dependencies that are not monitored by OpenStack. For the rest I think we should discuss, whether this standard should have a very broad scope or just focus on the IaaS Layer. When we decide for the latter, we will most likely need standards for the other layers too.

josephineSei:
@anjastrunk and I discussed @artificial-intelligence comment.
We both think, that we should have a standard for security and updates on each Layer from Hardware to OS to IaaS and up to KaaS.
I wrote 3 other issues, to reflect this:
Hardware Layer: https://github.com/SovereignCloudStack/standards/issues/790
OS Layer: https://github.com/SovereignCloudStack/standards/issues/791
KaaS Layer: https://github.com/SovereignCloudStack/standards/issues/792

@artificial-intelligence I hope this reflects your concerns.

This would also narrow done the scope of this PR to exactly the  IaaS Layer and dependencies of it.
If you miss some software or want to explicitly call out some in the standard, please let me know.

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
josephineSei: This was defined in the most recent version.

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
josephineSei: In the newest version we have timeframes.
We cannot force CSPs to document software versions, but for higher versions than SCS-compliant, like scs-open there is a requirement for SBOMs.
In the end, we must trust the CSPs to update their deployments fast.

fkr:
@Adri2000 @neuroserve @matfechner please review this - I think it is important that the view from CSPs is incorporated.

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,138 @@
...
+
+Considering that OpenStack as one provider of IaaS Layer Software has two releases per year, with one SLURP release per year, this option would require CSPs to update their deployment once or twice a year.
+Updating a whole deployment is a lot of work and requires also good life-cycle management.
+Following only the SLURP releases would reduce this work to once per year.
+
+While following new releases closely already provides a deployment with recent bug fixes and new features, it also makes developing standards easier.
+Differences between releases will accumulate eventually and may render older releases non-compliant to the SCS standards at some point.
+
+On the other hand on the IaaS Level there aren't many breaking changes introduced by releases and also most standards will also work with older releases.
+Security Updates and Bug fixes are also provided by OpenStack for a few older releases with the state `maintained`.
```
markus-hentsch, line 50:
```suggestion
Security updates and bug fixes are also provided by OpenStack for a few older releases with the state `maintained` according to the OpenStack releases overview[^2].
```

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,138 @@
...
+Additionally the [SCS reference implementation](https://github.com/SovereignCloudStack/release-notes/blob/main/Release7.md) is integrating OpenStack releases after half a year - so about the time when a new release is cut by OpenStack.
```
markus-hentsch, line 51:
```suggestion
Additionally the [SCS reference implementation](https://github.com/SovereignCloudStack/release-notes/blob/main/Release7.md) is integrating OpenStack releases after half a year - so about the time when a new release is published by OpenStack.
```

Did you mean something like "published"? "cut" sounds more like discontinued I think.

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,138 @@
...
+Considering a CSP that wants to use only SLURP releases and waiting for the reference implementation will already be over a year or 2 releases behind, which cannot be considered as using the current version of IaaS Layer Software.
```
markus-hentsch, line 52:
```suggestion
Considering a CSP that wants to use only SLURP releases and waits for the reference implementation to adopt them, will already lag over a year (i.e. 2 OpenStack releases) behind the latest release, this cannot be considered as using the current version of IaaS Layer Software.
```

I found the sentence hard to understand on the first read so I tried to rephrase it slightly for better understanding.

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,138 @@
...
+Thus this option can be discarded.
+
+#### Allow only maintained versions of Software
+
+While following closely to the newest releases could be advised, there are several downsides to requiring this workflow, even if it would be only for SLURP releases.
+Following the SCS reference implementation for example would also lead into being a little bit behind the newest release.
```
markus-hentsch, line 58:
```suggestion
Following the SCS reference implementation for example would also lead into being a little bit behind the newest OpenStack release.
```

Just adding this to make it clear that not the SCS reference implementation release is meant in the second part here.

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,138 @@
...
+But this is not as bad as it may seem to be, because security related fixes and bug fixes are backported to older but still `maintained` releases.
+All releases that are still maintained can be looked up at the releases page from OpenStack[^1].
+
+Allowing maintained versions would give CSPs a little bit more time to update and test their environments, while still receiving relevant security updates and bug fixes.
+Also CSPs that want to become SCS-compliant may not have the burden to upgrade their deployments, but can test before an upgrade, where they need to put in additional work to become SCS-compliant.
```
markus-hentsch, line 63:
```suggestion
Also CSPs that want to become SCS-compliant will not have to take on the burden to upgrade their deployments to very recent releases immediately, but can instead test with an existing release before an upgrade and identify where they need to put in additional work to become SCS-compliant.
```

I read the sentence several times but could not fully understand the intended meaning/point. Is this the meaning you intended?

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,138 @@
...
+
+One problem is, that there might be new features implemented in the newest versions of the software, which are desired by other standards to be SCS-compliant.
```
markus-hentsch, line 65:
```suggestion
One problem is, that there might be new features implemented in the newest versions of the software, which are desired by other SCS standards to be SCS-compliant.
```

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,138 @@
...
+In that case allowing all maintained versions would lead to a two-year timespan customers would need to wait for before such a feature becomes available in all SCS-compliant deployments.
+In case of security relevant features this is not advisable.
+
+#### Standards implicitly define the minimum versions of Software
+
+Instead of requiring a defined minimum software version, it could be derived from the standards.
```
markus-hentsch, line 71:
```suggestion
Instead of requiring a defined minimum software version centrally, it could be derived from the individual standards.
```

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,138 @@
...
+Security bug fixes must be implemented and proof of the fix then provided.
+Only under these circumstances deployments with older or alternative IaaS Layer software may be handled as compliant.
+
+This option could be taken for granted, but to actually advise using it may encourage CSPs to take a closer look on their life-cycle management and security risk handling.
+And CSPs using OpenStack could even be encouraged to upgrade their deployments.
+
+#### Dependencies of the IaaS Layer Software
+
+While the IaaS service software like OpenStack itself is monitored and security issues announced in OSSNs and OSSAs, these services have lots of dependecies, that are not monitored by the same entity.
+When dependencies have security issues, there might be no OSSN or OSSA, so CSPs also need to watch CVEs concerning these dependencies themself.
```
markus-hentsch, line 94:
```suggestion
When dependencies have security issues, there might be no OSSN or OSSA, so CSPs also need to watch CVEs concerning these dependencies themselves.
```

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,138 @@
...
+Those dependencies must also be updated in a reasonable timeframe, when a security issue is disclosed.
+
+#### What timeframe is needed to fix the issue?
+
+CSPs should be encouraged to fix security issues as fast as possible.
+Some security issues are very easy to exploit so as soon as the vulnerability is disclosed attacks on deployments will start.
+Other vulnerabilities may need much knowladge and more time to be exploited.
```
markus-hentsch, line 101:
```suggestion
Other vulnerabilities may need much knowledge and more time to be exploited.
```

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,138 @@
...
+Also the impact of different vulnerabilites will differ.
```
markus-hentsch, line 102:
```suggestion
Also the impact of different vulnerabilities will differ.
```

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,138 @@
...
+
+So it can be concluded that some security issues need to be fixed immediately while for others it is okay to take some time.
+The BSI already has some guidance[^1] on how fast CSPs should response.
```
markus-hentsch, line 105:
```suggestion
The BSI already has some guidance[^1] on how fast CSPs should respond.
```

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,138 @@
...
+This standard will follow this guidance and refer to these timeframes as "reasonable timeframes".
+
+## Standard for a minimum IaaS Layer Software version
+
+If a deployment is affected by a security issue and a maintained[^2] version of OpenStack is used as implementation for IaaS Layer software, security patches noted in OSSNs and OSSAs MUST be integrated within a reasonable timeframe according to the severity of the security issue[^1].
+Otherwise the CSP MUST implement security bug fixes themself within a reasonable timeframe, when the deplyoment is affected by a security issue according to the severity of the security issue[^1].
+
+In both cases a notice of the update MUST be send to the OSBA, so that the compliance will not be revoked.
+
+If a deployment uses a dependency of the IaaS service software which is affected ba a security issue, this software also MUST be updated with security patches within a reasonable timeframe[^1].
```
markus-hentsch, line 124:
```suggestion
If a deployment uses a dependency of the IaaS service software which is affected by a security issue, this software also MUST be updated with security patches within a reasonable timeframe[^1].
```

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,138 @@
...
+An open SBOM list MAY be used to propagate the current version of the software and may be used as proof of updates.
+
+[^2]: [OpenStack versions and their current status](https://releases.openstack.org)
+
+## Conformance Tests
+
+In case of provided SBOMs the version numbers of the software could be checked.
+But this is not a requirement, so there cannot be such a test.
+Tests on the integration of security patches itself are difficult.
+And even if tests for certain security issues are possible, then those might be received as an attack.
```
markus-hentsch, line 135:
```suggestion
And even if tests for certain security issues are possible, then those might be interpreted as an attack.
```

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,138 @@
...
+This is the reason there will be no conformance test.
+
+Rather the standard requiring that CSPs provide notice of the fixed vulnerabilites themself.
```
markus-hentsch, line 138:
```suggestion
Rather the standard requires that CSPs provide notice of the fixed vulnerabilites themselves.
```

markus-hentsch (commented):
I added some minor suggestions to the standard text.

anjastrunk (approved):

Standards/scs-XXXX-v1-security-of-iaas-service-software.md
```diff
@@ -0,0 +1,138 @@
...
+So it can be concluded that some security issues need to be fixed immediately while for others it is okay to take some time.
+The BSI already has some guidance[^1] on how fast CSPs should respond.
+From the moment a vulnerability is disclosed these are the advised reaction times ranked by the severity of the vulnerability:
+
+1. Critical (CVSS = 9.0 – 10.0): 3 hours
+2. High (CVSS = 7.0 – 8.9): 3 days
+3. Mid (CVSS = 4.0 – 6.9): 1 month
+4. Low (CVSS = 0.1 – 3.9): 3 months
+
+[^1]: [C5 criteria catalog with timeframes for responses on page 75.](https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Publikationen/Broschueren/C5_2020.pdf?__blob=publicationFile&v=3)
```
markus-hentsch, line 113:
```suggestion
[^1]: [C5 criteria catalog with timeframes for responses on page 70.](https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/CloudComputing/ComplianceControlsCatalogue/2020/C5_2020.pdf?__blob=publicationFile&v=3)
```

There seems to be an english version of that document that also contains the referenced paragraph.

markus-hentsch (approved):
I think this is in a very good shape for a draft now. Any open points such as the exact manner of providing proof of applied security fixes or applicable SBOM formats can be discussed with the community once this draft lands in main, in my opinion.

gtema (approved):

PR #767, opened 2024-10-01
Fix duplicate file names

tonifinger:
Fixes duplicate file names in the standards directory.


josephineSei (approved):
This just changes the standard number, because the 0118 was already taken. So I approve.

b1-lender (approved):

PR #770, opened 2024-10-07
Fix: give CSPs 2 weeks for providing k8s patch releases

mbuechse:
as discussed in SIG Standardization/Certification [meeting of 2024-08-22](https://github.com/SovereignCloudStack/minutes/blob/main/sig-standardization/20240822.md#stabilize-scs-compatible-kaas-v1).


Wieneo:
2 weeks could still be a bit short, but we want to set a good example.
LGTM!

garloff (approved):
Took me a moment to understand that the used 1.28.5 version is only used for testing ...
Change looks good to me!

garloff:
Approved in Team Container, 2024-10-10.

Issue #735, opened 2024-09-04
Make sonobuoy golang framework useable for production


tonifinger:
As described in DR, [scs-0200-v1-using-sonobuoy-for-kaas-conformance-tests.md](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0200-v1-using-sonobuoy-for-kaas-conformance-tests.md), it was decided to use sonobuoy as framework to deploy tests to Kubernetes clusters.

The current Sonobuoy setup within the repository was merely a PoC. We want to make it ready for production use.

The following must therefore be adjusted/implemented:

####  PREREQUISITES:

- [x] Implementation of the Golang test framework in accordance with the PoC (see: [kaas-sonobuoy-go-example-e2e-framework](https://github.com/SovereignCloudStack/standards/tree/main/Tests/kaas/kaas-sonobuoy-go-example-e2e-framework) )
- [ ] ZuuL job to create the Sonobuoy image and upload it to the Github container registry
  - apply multi platform builds using the multi-arch setup
  - rework plugin config to dynamically chose image tag
  - check go sum file

#### TASKS

- [x]  multi-arch support for `arm64` and `amd64` #758 
- [x]  Upgrade golang version: #747
- [ ] Write documentation on how you can contribute new tests. As part of [docs/issues/201](https://github.com/SovereignCloudStack/docs/issues/201)
- [ ] Update the DR to describe the current setup
- [ ] Zuul job to run the conformance test: #710 

#### OPTIONAL

- [ ] Write documentation on how a User(not a developer) can run the sonobuoy tests on its own.
  - currently blocked by the image not beeing build so far.
- [ ] apply golang linter https://golangci-lint.run/

anjastrunk:
@tonifinger How is the status here. Can this issue be closed or are there still open tasks?

PR #771, opened 2024-10-08
Add golang linter

michal-gubricky:
This PR adds the make command for golang linter and also adds linting for go language into the github actions

Related to https://github.com/SovereignCloudStack/standards/issues/735

### DIFF
--- a/Tests/kaas/kaas-sonobuoy-tests/Makefile
+++ b/Tests/kaas/kaas-sonobuoy-tests/Makefile
@@ -54,6 +54,7 @@ dev-prerequests:
 	@jq --version
 	@go version
 	@docker buildx version
+	# @golangci-lint version --format short	
 
 
 dev-setup: kind-init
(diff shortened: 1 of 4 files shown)


Tests/kaas/kaas-sonobuoy-tests/Makefile
```diff
@@ -54,6 +54,7 @@ dev-prerequests:
 	@jq --version
 	@go version
 	@docker buildx version
+	# @golangci-lint version --format short
```
tonifinger, line 57: I would say, we can uncommented this and apply it here as a prerequisite.

michal-gubricky: Or should we remove it? It's unnecessary because I added logic that automatically checks if `golangci-lint` is installed every time you run the `make lint` command. If it's not installed, it will automatically install version `1.61.0` for you.

michal-gubricky: What do you think @tonifinger, should I remove it or just uncommented it?

tonifinger: I think your approach of installing the linter automatically is good.
I agree it can be removed.

michal-gubricky: Thanks, I removed that comment in https://github.com/SovereignCloudStack/standards/pull/771/commits/8e69fa487a25311da1d839d2b5d59c7dbda7ad4a

tonifinger (approved):

PR #777, opened 2024-10-10
compliance monitor root url should be absolute

### LINKED ISSUE #775: [BUG] compliance monitor root url should be absolute
The markdown fragment ([example](https://compliance.sovereignit.cloud/markdown/table)) contains a lot of relative URLs, which is no good if the fragment is to be embedded somewhere else. The root url must be configured for the instance and be absolute.

mbuechse:
resolves #775


mbuechse:
I have deployed this branch to the live service: https://compliance.sovereignit.cloud/page/table
it works as expected

gtema (approved):

PR #778, opened 2024-10-10
Remove redundant clouds table

mbuechse:
Part of #672.


garloff (approved):
Thanks for leaving a link here to the other (authoritative) location.
One piece of feedback from @manuela-urban : She is missing the colorful green passed or red failed signage.
On the other hand we have more information now, as we can see all the scopes that pass, which I consider a plus.
Maybe a combination would be three columns:
Previous scpoes  | Current scopes | Future scopes
v3* PASS(green) | v4 PASS (green) | v5 FAIL(red)

Or is that too much information?

Anyhow, it's a possible future improvment, certainly not required to merge this one.

garloff:
Agreed:
- Yes, this is a separate PR
- Just a green color in front of passed versions that are currently effective would give the wanted visual clues.

PR #786, opened 2024-10-18
Shorten GPU names by dropping h. Fix intel Gen 12.7.

garloff:
Plus some typos.

### DIFF
--- a/Tests/iaas/flavor-naming/flavor_names.py
+++ b/Tests/iaas/flavor-naming/flavor_names.py
@@ -250,14 +250,19 @@ class Flavorname:
 
     def shorten(self):
         """return canonically shortened name as recommended in the standard"""
-        if self.hype is None and self.hwvirt is None and self.cpubrand is None:
+        if self.hype is None and self.hwvirt is None and self.cpubrand is None and self.gpu is None:
             return self
+        # GPU: Remove h modifiers - cleaner would be to build a c'tor like with CPUBrand
+        gpu = self.gpu
+        if self.gpu:
+            gpu.perf = ""
+            gpu.vramperf = ""
         # For non-x86-64, don't strip out CPU brand for short name, as it contains the architecture
         if self.cpubrand and self.cpubrand.cpuvendor not in ('i', 'z'):
             return Flavorname(cpuram=self.cpuram, disk=self.disk,
                               cpubrand=CPUBrand(self.cpubrand.cpuvendor),
-                              gpu=self.gpu, ib=self.ib)
-        return Flavorname(cpuram=self.cpuram, disk=self.disk, gpu=self.gpu, ib=self.ib)
+                              gpu=gpu, ib=self.ib)
+        return Flavorname(cpuram=self.cpuram, disk=self.disk, gpu=gpu, ib=self.ib)
 
 
 class Outputter:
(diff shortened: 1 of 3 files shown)


garloff:
This is for the flavor name generator, where we suggest a short name in addition to a full name.
Strip out the h qualifiers there.

mbuechse:
Shall we only drop the `h` qualifiers or other stuff as well? Dropping `h` alone doesn't do much to shorten the name...

garloff:
Well, I think GPU capacity ("cores") and memory are key properties a user would look for. Also vendor.
We could debate generation, but they are more relevant for GPUs than for CPUs, so I'd rather not drop them...

mbuechse (commented):
Oh, I forgot to submit this review.

Tests/iaas/flavor-naming/flavor_names.py
```diff
@@ -250,14 +250,19 @@ def __init__(
 
     def shorten(self):
         """return canonically shortened name as recommended in the standard"""
-        if self.hype is None and self.hwvirt is None and self.cpubrand is None:
+        if self.hype is None and self.hwvirt is None and self.cpubrand is None and self.gpu is None:
             return self
+        # GPU: Remove h modifiers - cleaner would be to build a c'tor like with CPUBrand
```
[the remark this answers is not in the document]
mbuechse: see e8975f4ab837a72a4487903b1067d8b00ba3762a

mbuechse (approved):
LGTM (please consider my suggestion in the first review)

Tests/iaas/flavor-naming/flavor_names.py
```diff
@@ -226,13 +244,29 @@ class GPU:
     vram = OptIntAttr("#.V:GiB VRAM")
     vramperf = TblAttr("Bandwidth", {"": "Std BW {<~1GiB/s)", "h": "High BW", "hh": "Very High BW"})
 
+    def __init__(self, gputype="g", brand="N", gen='', cu=None, perf='', vram=None, vramperf=''):
```
garloff, line 247: I'd need to double check that we are not defaulting to having a virtualized NVidia GPU this way...

Tests/iaas/flavor-naming/flavor_names.py
```diff
@@ -171,20 +174,29 @@ class Disk:
     disksize = OptIntAttr("#.GB Disk")
     disktype = TblAttr("Disk type", {'': '(unspecified)', "n": "Networked", "h": "Local HDD", "s": "SSD", "p": "HiPerf NVMe"})
 
+    def shorten(self):
+        return self
```
garloff, line 178: We could drop disktypes that are networked here, but that would be new functionality... Good to have the option now, but it's a separate discussion, I guess.

Tests/iaas/flavor-naming/flavor_names.py
mbuechse: The constructor is only used in one place, namely shorten, and there gputype and brand are propagated.

Tests/iaas/flavor-naming/flavor_names.py
mbuechse: Disk type could even be replaced by `''` in any case (even if it's `n`)

garloff:
Thanks, @mbuechse!

PR #788, opened 2024-10-24
Add new standards to scs-compatible-iaas-v5 and make scripts executable

mbuechse:
NOTE: This is a draft. There are still a few glitches in the new scripts (I suppose some of the open PRs have to be merged first), and the scripts don't yet output the necessary line (`testcase-id: PASS/FAIL`). I will address the latter point soon.


anjastrunk (changes requested):
I couldn't execute yaml file itself. Hence,  my review is limited on list of requested standards and URLs and I missed standard for [domain manager role](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0302-v1-domain-manager-role.md). It is not part of yaml file. Was this intended?

Furthermore, all standards are in draft state, currently. Should we stabilize them prior to publishing an new certificate version??

mbuechse:
> I couldn't execute yaml file itself. Hence, my review is limited on list of requested standards and URLs and I missed standard for [domain manager role](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0302-v1-domain-manager-role.md). It is not part of yaml file. Was this intended?

No, that was a mistake on my part.

> Furthermore, all standards are in draft state, currently. Should we stabilize them prior to publishing an new certificate version??

Yes, we should!

anjastrunk (approved):
LGTM

PR #793, opened 2024-10-28
Apply CI job to build the sonobuoy image for kaas conformance testing

tonifinger:


.github/workflows/build-docker.yml
```diff
@@ -44,3 +45,34 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+
+  build-and-push-scs-compliance-check-image:
```
mbuechse, line 49: Just FYI: this key seems to be a duplicate of the one in line 18

tonifinger: Good catch! I changed the job name.

PR #797, opened 2024-11-01
Add ScaleUp Open Cloud Region OCC2

fzakfeld:


mbuechse:
@fzakfeld That's looking quite well. We still need to amend `.zuul.d/secure.yaml`, I'm afraid. Do you think you can do that, or will you send me the necessary credentials? You can find me on Matrix, for instance, and send me an encrypted message there. My handle is @halfbigdata:tchncs.de

mbuechse (approved):
LGTM. Secret still needs to be added to Github, but let's give it a try.

PR #798, opened 2024-11-01
UX improvement: table_full leads to detail_full

mbuechse:
In other words, when you are in a privileged table view, you will also be directed to a privileged details view (which not only shows most recent failures, but also includes links to the corresponding reports)


garloff (approved):
Thanks, appreciated!

PR #753, opened 2024-09-18
 Enable compliance tests to use plugins for cluster provisioning

tonifinger:
This PR introduces an interface that enables the provision of kubeconfig files for the sonobuoy test framework.
Each Kubernetes provider must derive its own specific plugin from this interface class in order to provide a cluster for testing

PR #800, opened 2024-11-02
Feat/add cluster stacks plugin

### LINKED ISSUE #774: [Feature Request]  Implement plugin for cluster-stacks
Based on  #710, a plugin must be added that creates k8s clusters with the usage of [cluster-stacks](https://github.com/SovereignCloudStack/cluster-stacks)

To do this, we need to derive a plugin from [Tests/kaas/plugin/interface.py](https://github.com/SovereignCloudStack/standards/blob/710-feature-request-enable-compliance-tests-to-use-plugins-for-cluster-provisioning/Tests/kaas/plugin/interface.py)   that implements the following two methods:

* [_create_cluster](https://github.com/SovereignCloudStack/standards/blob/f994cde8da7cb4142d5557c80e87a35a81ab6dc7/Tests/kaas/plugin/interface.py#L33)
*  [_delete_cluster](https://github.com/SovereignCloudStack/standards/blob/f994cde8da7cb4142d5557c80e87a35a81ab6dc7/Tests/kaas/plugin/interface.py#L42)

The current plugin implementation from kind can be used as an example: [Tests/kaas/plugin/plugin_kind.py](https://github.com/SovereignCloudStack/standards/blob/710-feature-request-enable-compliance-tests-to-use-plugins-for-cluster-provisioning/Tests/kaas/plugin/plugin_kind.py)

michal-gubricky:
This PR implement plugin for cluster-stacks

Fixes #774


Tests/kaas/plugin/cluster.yaml
mbuechse, line 1: This file and the next one should be commented as well.

Tests/kaas/plugin/plugin_kind.py
```diff
@@ -21,7 +21,7 @@ def _create_cluster(self):
         if self.config is None:
             self.cluster.create()
         else:
-            self.cluster.create(self.config)
+            self.cluster.create()
```
mbuechse, line 24: Why this change? It has nothing to do with cluster stacks.

Tests/kaas/plugin/run_plugin.py
mbuechse, line 1: this needs to be rebased on the latest version of the target branch

Tests/kaas/plugin/.pytest-kind/my-cluster/kind-v0.17.0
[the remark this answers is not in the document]
michal-gubricky: Yes, it was. I will remove it.

Tests/kaas/plugin/README.md
```diff
@@ -23,17 +23,91 @@
...
+
+To set the `GIT_ACCESS_TOKEN`, run the following command in your terminal:
+
+```bash
+export GIT_ACCESS_TOKEN=<your-github-token>
+```
+
+### Configuring clusterspec.yaml file
+
+The `clusterspec.yaml` file is used to set parameters for creating a Kubernetes cluster with the `cluster-stacks` plugin. This file allows you to specify details related to the cluster-stack, Git integration, and cluster configurations.
```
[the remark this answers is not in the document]
michal-gubricky: Ah, okay. Then we need to think about some new config file where will be defined parameters/variables specific to the plugins (kind, cluster-stacks, etc.) or we can say that the user/tester needs to define some env variables before executing the plugin.

Tests/kaas/plugin/README.md
```diff
@@ -23,17 +23,91 @@
...
+
+### Mandatory Parameter
+
+The only mandatory parameter in `clusterspec.yaml` is `clouds_yaml_path` field, which points to the `clouds.yaml` file for OpenStack. If other parameters are omitted, the default values will be used.
```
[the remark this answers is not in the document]
michal-gubricky: It’s mandatory because the plugin needs access to the OpenStack instance. However, we could set the default value of this variable to `~/.config/openstack/clouds.yaml`, since that's what you’re using most of the time.

Tests/kaas/plugin/kubeconfig-cs-cluster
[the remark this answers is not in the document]
michal-gubricky: My mistake, this was accidentally pushed to the branch and should be removed.

Tests/kaas/plugin/cluster.yaml
michal-gubricky: Sure, I can do that.

Tests/kaas/plugin/README.md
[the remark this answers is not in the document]
mbuechse: Yes, the file is mandatory, but the path shouldn't be, because we have the default value for that.

Tests/kaas/plugin/README.md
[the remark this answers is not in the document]
mbuechse: The plugin's constructor will take the path of a config file, and that config file should contain everything that is specific to the plugin and the test subject (it could in turn refer to other files, if necessary). The path of the config file can be specified per subject in config.toml.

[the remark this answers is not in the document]
michal-gubricky: I thought `clusterspec.yaml` was the config file, as it was the only argument (called `CLUSTERSPEC_PATH`) you could pass to the plugin. However, it seems that it was changed in yesterday's commits. I will take a closer look into that.

Tests/kaas/plugin/README.md
```diff
@@ -23,17 +23,91 @@
...
+
+You can include additional parameters in `clusterspec.yaml` to customize the cluster setup. These optional parameters are grouped below by their category.
+
+#### Cluster-Stack Related Parameters
+
+These parameters configure specific settings for the cluster-stack:
+
+```yaml
+cs_name: <cs_name>                  # Default: "scs"
+cs_k8s_version: <cs_k8s_version>    # Default: "1.29"
```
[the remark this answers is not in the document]
michal-gubricky: Yes, I agree that the `cs_k8s_version` should be taken from the branch parameter in the `clusterspec.yaml` file.

Tests/kaas/plugin/README.md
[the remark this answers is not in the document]
michal-gubricky: It now take an additional configuration file specific to the plugin. I have also added an example file, see `plugin-cluster-stacks-config.yaml`. 

And I have one question regarding this. How should the plugin configuration file look when you want to provision multiple clusters? Do you need a separate configuration file for each cluster, or is a single file sufficient, with the `clusterspec_cluster` variable determining the cluster?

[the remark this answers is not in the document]
mbuechse: In my mind, the plugin will merge the plugin configuration with the clusterspec information.

Tests/kaas/plugin/run_plugin.py
michal-gubricky: Rebased to the latest version

Tests/kaas/plugin/plugin_kind.py
michal-gubricky: This change is currently outdated, as the code in #753 has also been modified.

Tests/kaas/plugin/README.md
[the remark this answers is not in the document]
michal-gubricky: So it should be merged based on the `clusterspec_cluster`. In my mind, the config file for the plugin would then look something like this:

```yaml
current-k8s-release:
  clouds_yaml_path: "~/.config/openstack/clouds.yaml"   # Path to the OpenStack clouds.yaml file
  cs_name: "scs"                                        # Cluster Stack Name
  cs_version: "v1"                                      # Cluster Stack Version
  # Additional configurations
  ...
  ...
  ...

current-k8s-release-1:
  ...
  ...
  ...
  ```

**Comment by mbuechse** [comment]
If this has been rebased to the latest version, why then does it have merge conflicts?

**Comment by michal-gubricky** [comment]
> If this has been rebased to the latest version, why then does it have merge conflicts?

As you can see, I added that comment on 11.11.2024 at 10:37 a.m., and then, after rebasing, a new commit was added to the target branch later that afternoon.

**Comment by mbuechse** [comment]
> As you can see, I added that comment on 11.11.2024 at 10:37 a.m., and then, after rebasing, a new commit was added to the target branch later that afternoon.

Understood. Thanks for doing all the great work. Please work with Toni to get this merged asap.

File: Tests/kaas/plugin/plugin_cluster_stacks.py [hunk 5]
```diff
@@ -0,0 +1,366 @@
...
+            logger.info("All CAPI system pods are ready.")
+            return True
+
+        logger.info("Waiting for all CAPI pods to become ready...")
+        time.sleep(interval)
+
+    raise TimeoutError(f"Timed out after {timeout} seconds waiting for CAPI and CAPO system pods to become ready.")
+
+
+def wait_for_cso_pods_ready(timeout=240, interval=15):
```
mbuechse, line 69: This seems like a variation of the function above (if not literally copy and paste), only with different `namespaces`. The function above should take `namespaces` as parameter, renamed `wait_for_pods`, and then `wait_for_cso_pods_ready` and `wait_for_capi_pods_ready` be implemented as a simple call to `wait_for_pods`.

Tests/kaas/plugin/plugin_cluster_stacks.py
```diff
@@ -0,0 +1,366 @@
...
+            logger.info("All CSO pods in 'cso-system' namespace are ready.")
+            return True
+
+        logger.info("Waiting for CSO pods in 'cso-system' namespace to become ready...")
+        time.sleep(interval)
+
+    raise TimeoutError(f"Timed out after {timeout} seconds waiting for CSO pods in 'cso-system' namespace to become ready.")
+
+
+def wait_for_workload_pods_ready(namespace="kube-system", timeout=600, kubeconfig_path=None, max_retries=3, delay=30):
```
mbuechse, line 117: Interesting that this function works quite differently from the ones above.

Tests/kaas/plugin/plugin_cluster_stacks.py
```diff
@@ -0,0 +1,366 @@
...
+
+    raise RuntimeError(f"Error waiting for pods in namespace '{namespace}' to become ready.")
+
+
+def load_config(config_path):
+    """
+    Loads the configuration from a YAML file.
+    """
+    if not os.path.exists(config_path):
+        raise FileNotFoundError(f"Configuration file {config_path} not found.")
```
mbuechse, line 150: I'm pretty sure this is totally unnecessary, because the subsequent `open` will raise the same exception. This kind of checking also introduces the possibility of race conditions, because the `os.path.exists` can return `True`, but then the file can get deleted before the script arrives at `open`.

```suggestion
```

Tests/kaas/plugin/plugin_cluster_stacks.py
```diff
@@ -0,0 +1,366 @@
...
+    with open(config_path, 'r') as file:
+        config = yaml.safe_load(file) or {}
+        return config
+
+
+def setup_environment_variables(self):
+    # Cluster Stack Parameters
+    self.clouds_yaml_path = self.config.get('clouds_yaml_path', '~/.config/openstack/clouds.yaml')
+    self.cs_k8s_version = self.cluster_version
+    self.cs_name = self.config.get('cs_name', 'scs')
```
mbuechse, line 161: If we really need to hard-code all these defaults (which I doubt because we deliver an example file, and this thing won't work without a config file), then the defaults should be put into a dict, for instance

```python
DEFAULTS = {
    'cs_name': 'scs',
    'cs_version': 'v1',
    ...
}
```

Then I would do

```python
for key, value in DEFAULTS.items():
    self.config.setdefault(key, value)   # will only set if not already present
```

Also, for the environment variables, I think the detour with the instance variables is unnecessary.

You can do just

```python
required_env = {key.upper(): value for key, value in self.config}
```

Of course, this is not quite correct yet, because some values have to be computed differently, and not all keys from the config would be used. For the first point, we can always amend `required_env`, and for the second, we can add a guard to the dict comprehension like so:

```python
required_env = {key.upper(): value for key, value in self.config if key in ENV_KEYS}
```

where `ENV_KEYS` would be a global variable of type `set`, such as

```python
# don't use uppercase here because this refers to config.yaml keys
ENV_KEYS = {'cs_name', 'cs_version', ...}
```

Tests/kaas/plugin/plugin_cluster_stacks.py
```diff
@@ -0,0 +1,366 @@
...
+        'GIT_PROVIDER_B64': base64.b64encode(git_provider.encode()).decode('utf-8'),
+        'GIT_ORG_NAME_B64': base64.b64encode(git_org_name.encode()).decode('utf-8'),
+        'GIT_REPOSITORY_NAME_B64': base64.b64encode(git_repo_name.encode()).decode('utf-8')
+    })
+
+    git_access_token = os.getenv('GIT_ACCESS_TOKEN')
+    if git_access_token:
+        os.environ['GIT_ACCESS_TOKEN_B64'] = base64.b64encode(git_access_token.encode()).decode('utf-8')
+    else:
+        raise ValueError("GIT_ACCESS_TOKEN environment variable not set.")
```
mbuechse, line 217:
```suggestion
    if not git_access_token:
        raise ValueError("GIT_ACCESS_TOKEN environment variable not set.")
    os.environ['GIT_ACCESS_TOKEN_B64'] = base64.b64encode(git_access_token.encode()).decode('utf-8')
```

Tests/kaas/plugin/plugin_cluster_stacks.py
```diff
@@ -0,0 +1,366 @@
...
+        self.working_directory = os.getcwd()
+        logger.debug(f"Working from {self.working_directory}")
+
+    def create_cluster(self, cluster_name="scs-cluster", version=None, kubeconfig_filepath=None):
+        self.cluster_name = cluster_name
+        self.cluster_version = version
+
+        # Setup variables
+        setup_environment_variables(self)
+        setup_git_env(self)
```
mbuechse, line 233: I don't think we should use `os.environ.update` to update the environment of this process. We should construct the dict with the updates that we need and then pass this dict to `subprocess.run` (well possible with `self._run_subprocess`).

mbuechse (commented):
A few observations

Tests/kaas/plugin/plugin_cluster_stacks.py
```diff
@@ -0,0 +1,366 @@
...
+                url = "https://github.com/SovereignCloudStack/cluster-stack-provider-openstack/releases/latest/download/cspo-infrastructure-components.yaml"
+                command = f"curl -sSL {url} | /tmp/envsubst | kubectl apply -f -"
+            else:
+                raise ValueError(f"Unknown file or URL: {yaml_file}")
+
+            self._run_subprocess(command, error_msg, shell=True)
+        except subprocess.CalledProcessError as error:
+            raise RuntimeError(f"{error_msg}: {error}")
+
+    def _get_kubeadm_control_plane_name(self):
```
mbuechse, line 324: this looks like a static method (no reference of self), but that might change once we stop polluting `os.environ`

Tests/kaas/plugin/plugin_cluster_stacks.py
```diff
@@ -0,0 +1,366 @@
...
+                    print(f"KubeadmControlPlane name: {kcp_name_stdout}")
+                    return kcp_name_stdout
+            except subprocess.CalledProcessError as error:
+                print(f"Error getting kubeadmcontrolplane name: {error}")
+            # Wait before retrying
+            time.sleep(delay_between_retries)
+        else:
+            raise RuntimeError("Failed to get kubeadmcontrolplane name")
+
+    def _wait_kcp_ready(self, kcp_name):
```
mbuechse, line 344: same here

mbuechse:
After looking at this PR in more detail, I think it does need some more work. Nothing seems to have happened today after 9 am CET, which baffles me a bit, since you had put a thumbs up under my remark that went

> Please work with Toni to get this merged asap.

mbuechse:
@michal-gubricky @tonifinger Please communicate proactively (using Matrix or whatever suits you) to get this done ASAP, ideally until tomorrow (November 14) at noon.

mbuechse:
Sorry, one more thing, but now back to the subject matter at hand. If I understand correctly, deploys two clusters

- a management cluster on Kind
- the workload cluster on OpenStack

Well, at least roughly. To me, this seems like a lot of work, and it doesn't feel like Kubernetes as a service at all. Not in the least. Where is the service if you have to download and invoke Helm charts? Shouldn't a service come with something like an API? Are we doing something wrong here???

mbuechse:
My impression (after talking to Toni) is that we shouldn't set up the management cluster ourselves. The management cluster should be ready to use and the corresponding Kubeconfig file provided to us by our partner. I have written to the Team Container Matrix channel to get confirmation on that.

michal-gubricky:
> After looking at this PR in more detail, I think it does need some more work. Nothing seems to have happened today after 9 am CET, which baffles me a bit, since you had put a thumbs up under my remark that went

> I see now that you requested a review from Toni at 12:25 or so. If I say "work with XY asap", I think requesting a review via a button in a UI does not quite fit the bill...

First of all, this is not true. I just did not assign Toni. We have already had a few conversations about it. I asked him if he could take a look to see if anything was missing from his perspective. He also mentioned that you were planning to store the config file under playbooks/k8s_configs/." That's why I this [commit](https://github.com/SovereignCloudStack/standards/pull/800/commits/f880dfec0e00e5da0eba86b4530cc17bd0b3e765).

michal-gubricky:
Thanks for 2nd round of review, I will take a look at it.

mbuechse:
> First of all, this is not true. I just did not assign Toni. We have already had a few conversations about it. I asked him if he could take a look to see if anything was missing from his perspective. He also mentioned that you were planning to store the config file under playbooks/k8s_configs/." That's why I this [commit](https://github.com/SovereignCloudStack/standards/pull/800/commits/f880dfec0e00e5da0eba86b4530cc17bd0b3e765).

You're right, and I apologize. First, you did also ping him, and second, you committed something after 9 am.

But my overall point still stands, because I don't see any change after that commit.

BTW, I don't know went wrong, but you shouldn't have moved the file. Yes, I do plan to have the config files in the playbooks directory, but only real config files from our partners, but not templates or illustrative examples.

michal-gubricky:
> Sorry, one more thing, but now back to the subject matter at hand. If I understand correctly, the plugin deploys two clusters
> 
> * a management cluster on Kind
> * the workload cluster on OpenStack
> 
> Well, at least roughly. To me, this seems like a lot of work, and it doesn't feel like Kubernetes as a service at all. Not in the least. Where is the service if you have to download and invoke Helm charts? Shouldn't a service come with something like an API? Are we doing something wrong here???

> My impression (after talking to Toni) is that we shouldn't set up the management cluster ourselves. The management cluster should be ready to use and the corresponding Kubeconfig file provided to us by our partner. I have written to the Team Container Matrix channel to get confirmation on that.

In short, cluster stacks work in such a way that you need a management cluster. For this purpose, you can use a simple kind cluster or some already existing k8s cluster. In that mgmt cluster, you need to deploy the necessary components. After that, you just apply those two templates `clusterstack.yaml` and `cluster.yaml`.

mbuechse:
I don't mean to blame you personally. Maybe Toni was unresponsive. He says he will try to set up a call so that you can indeed work together in a focused manner.

Tests/kaas/plugin/plugin_cluster_stacks.py
```diff
@@ -0,0 +1,366 @@
...
+
+        # Wait for workload system pods to be ready
+        print(self.kubeconfig_cs_cluster_filename)
+        wait_for_workload_pods_ready(kubeconfig_path=self.kubeconfig_cs_cluster_filename)
+
+    def delete_cluster(self, cluster_name=None, kubeconfig_filepath=None):
+        kubeconfig_cs_cluster_filename = f"kubeconfig-{cluster_name}.yaml"
+        try:
+            # Check if the cluster exists
+            check_cluster_command = f"kubectl get cluster {cluster_name} --kubeconfig {kubeconfig_filepath}"
```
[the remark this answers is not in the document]
michal-gubricky: In this command, I want to check if the cluster resource exists, which is located in the management cluster(in our case it is kind now)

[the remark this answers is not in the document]
mbuechse: Oh, sorry, you're right. This is a custom resource, right? Got it!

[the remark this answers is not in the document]
michal-gubricky: Yes, it is a custom resource.

Tests/kaas/plugin/plugin_cluster_stacks.py
michal-gubricky: I will try to merge these two functions into the one `wait_for_pods`

Tests/kaas/plugin/plugin_cluster_stacks.py
michal-gubricky: In this function, I am trying to get the name of the `kubeadmcontrolplane` resource because it is generated dynamically. In the next function, `_wait_kcp_ready`, the name of the `kcp` (kubeadmcontrolplane) resource is used to check whether it is ready. Once confirmed, you can proceed to retrieve the kubeconfig of the workload cluster.

Tests/kaas/plugin/plugin_cluster_stacks.py
```diff
@@ -0,0 +1,366 @@
...
+        setup_git_env(self)
+
+        # Create the Kind cluster
+        self.cluster = KindCluster(name=cluster_name)
+        self.cluster.create()
+        self.kubeconfig = str(self.cluster.kubeconfig_path.resolve())
+        if kubeconfig_filepath:
+            shutil.move(self.kubeconfig, kubeconfig_filepath)
+        else:
+            kubeconfig_filepath = str(self.kubeconfig)
```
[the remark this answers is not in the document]
michal-gubricky: I also discussed this with @tonifinger today, and I will make the change. We agreed that the name/path of the kubeconfig file will be sourced from the `clusterspec.yaml` file from field `kubeconfig`. And the kubeconfig file for the management (kind) cluster will be named something like `kubeconfig-mngmt`.

Tests/kaas/plugin/plugin_cluster_stacks.py
```diff
@@ -0,0 +1,327 @@
...
+        logger.debug(self.config)
+        self.working_directory = os.getcwd()
+        for key, value in DEFAULTS.items():
+            self.config.setdefault(key, value)
+        self.clouds_yaml_path = os.path.expanduser(self.config.get('clouds_yaml_path'))
+        self.cs_namespace = self.config.get('cs_namespace')
+        logger.debug(f"Working from {self.working_directory}")
+        self.kubeconfig_mgmnt_path = "kubeconfig-mgmnt.yaml"
+
+    def create_cluster(self, cluster_name="scs-cluster", version=None, kubeconfig_filepath=None):
```
tonifinger, line 137:
```suggestion
    def create_cluster(self, cluster_name=None, version=None, kubeconfig_filepath=None):
```
I have also adapted this in the “parent” issue.  There should be no default values, as only the test subject handled by `scs-test-runner-py` should specify a name for the cluster creation here.

Tests/kaas/plugin/plugin_cluster_stacks.py
```diff
@@ -0,0 +1,327 @@
...
+        self.cluster_name = cluster_name
+        self.cluster_version = version
+        self.kubeconfig_cs_cluster = kubeconfig_filepath
+
+        # Create the Kind cluster
+        self.cluster = KindCluster(name=cluster_name)
+        self.cluster.create()
+        self.kubeconfig_mgmnt = str(self.cluster.kubeconfig_path.resolve())
+        if self.kubeconfig_mgmnt:
+            shutil.move(self.kubeconfig_mgmnt, self.kubeconfig_mgmnt_path)
```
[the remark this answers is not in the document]
michal-gubricky: Now we do not need to store the `kubeconfig_mgmnt` file on the local filesystem, see 3b32e322496a675b8aef615ec32e2554e05b43d7.

Tests/kaas/plugin/plugin_cluster_stacks.py
```diff
@@ -0,0 +1,327 @@
...
+
+        # Remove kubeconfigs
+        if os.path.exists(kubeconfig_cs_cluster_filename):
+            os.remove(kubeconfig_cs_cluster_filename)
+        if os.path.exists(self.kubeconfig_mgmnt):
+            os.remove(self.kubeconfig_mgmnt)
+
+    def _apply_yaml_with_envsubst(self, yaml_file, error_msg, kubeconfig=None):
+        """
+        Applies a Kubernetes YAML configuration file to the cluster, substituting environment variables as needed.
```
tonifinger, line 217:
```suggestion
        Applies a Kubernetes YAML configuration file to the  mgmnt cluster, substituting environment variables as needed.
```
I would suggest referring to "mgmnt cluster" here to not confuse it with "workload clusters"

Tests/kaas/plugin/plugin_cluster_stacks.py
mbuechse: No default values means no default values. None as default would still be wrong

mbuechse:
superseded by #916

PR #803, opened 2024-11-04
enable better differentiation of badge color

### LINKED ISSUE #801: [Feature Request] Compliance monitor colors
We currently assign green if a cloud passes the currently effective version of a certification scope, red if it only passes deprecated versions and no color is nothing is passed.
Green is fine, but I would revert the colorization of the latter:
* Red if NOTHING passes
* No color (or amber) if only old versions pass.

mbuechse:
resolves #801


mbuechse:
Screenshot from 2024-11-04 20-32-37

garloff (approved):
Nice!

garloff:
One more thing, but maybe for another PR:
The symbols +<;dagger:) and * lack an explanation.

Issue #639, opened 2024-06-17
[Standard] Stabilize node distribution standard


cah-hbaum:
Follow-up for https://github.com/SovereignCloudStack/standards/pull/524
The goal is to set the `Node distribution` standard to `Stable` after all discussion topics are debated and decided and the necessary changes derived from these discussions are integrated into the Standard and its test.

The following topics need to be discussed:
- [ ] How is node distribution handled on installations with shared-control plane nodes (Kamaji, Gardener, etc) - see e.g. https://github.com/SovereignCloudStack/standards/pull/524#pullrequestreview-2122476212
- [ ] What should be done about control-planes with e.g. 3 nodes containing 3 etcd members, which are only distributed on 2 physical machines (and similar scenarios)what to do about control-planes with e.g. 3 control plane nodes and 2 etcd nodes - see e.g. https://github.com/SovereignCloudStack/standards/pull/524#discussion_r1642411303
- [ ] Where is the differentiation between `Node distribution` and things like `High Availability` or `Redundancy`? Should this standard only be a precursor for a `High Availability' standard? (more information under #579)
- [ ] Should information about external `etcd` (https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/ha-topology/#external-etcd-topology) be integrated here? (see https://github.com/SovereignCloudStack/standards/pull/524#discussion_r1642540079)

cah-hbaum:
Topic 1: How is node distribution handled on installations with shared-control planes (Kamaji, Gardener, etc)?

cah-hbaum:
### Topic 2: Differentiation between `Node distribution` and things like `High Availability` or `Redundancy`

martinmo:
I brought this issue up in today's Team Container Call and edited the above sections accordingly. As part of #649 we will also get access to Gardener and soon Kamaji clusters.

One thing I want to make you aware of @cah-hbaum: in the call, it was pointed out that term *shared control-plane* isn't correct. The control-plane isn't shared, instead, the control-plane *nodes* are shared and thus we should always say *shared control-plane node*.

(I edited above texts accordingly as well to refer to shared control-plane *nodes*.)

joshmue:
Another potential problem with the `topology.scs.community/host-id` label:

The concept of using the "host-id" may not play nice with VM live migrations.

[...]

PR #806, opened 2024-11-05
providing new standard for node distribution in draft version - to be…

piobig2871:
providing new standard for node distribution in draft version. 

Key points to discuss: 
* how to combine the information from previous standards to make it true in whole scope
* if all information are correct 
* possible improvement 

Related to https://github.com/SovereignCloudStack/standards/issues/639

### DIFF
--- a/Standards/scs-0214-v3-k8s-node-distribution.md
+++ b/Standards/scs-0214-v3-k8s-node-distribution.md
@@ -1,0 +1,130 @@
+---
+title: Kubernetes Node Distribution and Availability
+type: Standard
+status: Draft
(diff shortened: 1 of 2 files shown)


mbuechse:
Please note that v2 is still in draft state and can be changed. No urgent need to introduce a new major version.

see https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0214-v2-k8s-node-distribution.md

piobig2871:
> Please note that v2 is still in draft state and can be changed. No urgent need to introduce a new major version.
> 
> see https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0214-v2-k8s-node-distribution.md

@mbuechse: I will change that then.

PR #820, opened 2024-11-13
Hotfix: fix front matter in Supplement to scs-0302

mbuechse:
Otherwise the docs page cannot be built

### DIFF
--- a/Standards/scs-0302-w1-domain-manager-implementation-notes.md
+++ b/Standards/scs-0302-w1-domain-manager-implementation-notes.md
@@ -1,8 +1,8 @@
 ---
 title: Domain Manager implementation notes
 type: Supplement
-track: IaaS
-status: Proposal
+track: IAM
+status: Draft
 supplements:
   - scs-0302-v1-domain-manager-role.md
 ---


Standards/scs-0302-w1-domain-manager-implementation-notes.md
```diff
@@ -1,8 +1,8 @@
 ---
 title: Domain Manager implementation notes
 type: Supplement
-track: IaaS
-status: Proposal
+track: IAM
+status: Draft
```
markus-hentsch, line 5:
```suggestion
status: Stable
```

The related standard [is stable on main](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0302-v1-domain-manager-role.md?plain=1#L4) shouldn't the supplement be too then?

mbuechse: No, supplements are meant to be updated more often than standards. That is one reason why we have them. All of them stay in draft state.

markus-hentsch: Oh TIL they don't share the same state as the standard they are supplementing, didn't notice that. My apologies.

> All of them stay in draft state.

Do they? Any supplement I looked at in `main` seems to be in Proposal state though.

markus-hentsch (approved):
I'm approving this as a non-invasive hotfix to get the docs page built.

We will have to check the `status` of other supplements though as marking this as "Draft" is not in-line with the existing ones. This however seems to be a minor detail that is negligible for the time being.

Standards/scs-0302-w1-domain-manager-implementation-notes.md
mbuechse: You are right, and that is strage. I will investigate. My feeling is that a Proposal (as opposed to a Draft) should never enter main branch.

mbuechse:
I'm done investigating. Findings:

- [scs-0001](https://docs.scs.community/standards/scs-0001-v1-sovereign-cloud-standards#process) does not contain a formal Proposal state (not any more, since you @markus-hentsch have lobbied to remove it)
- the same document (at the end of the linked section) states "Supplements may be kept in Draft state, because they are not authoritative."

So what I suspected at first is correct: Proposal should not go into main, and Draft would be correct. Stable would be possible, but it would defeat the purpose of a Supplement. I will open a PR to correct the other supplements.

PR #824, opened 2024-11-13
Stabilize SCS-compatible IaaS v5

mbuechse:
make it effective from 2024-11-16 and turn v4 into warn from 2025-01-01


garloff (approved):
/lgtm
And we approved the pieces in the IaaS meeting and the final std in the SIG Std/Cert.

PR #826, opened 2024-11-14
Add compliance test for standard scs-0219-v1

kgube:


Tests/scs-compatible-kaas.yaml
```diff
@@ -40,7 +53,8 @@ versions:
   - version: v1
     include:
       - ref: cncf-k8s-conformance
-      - ref: scs-0210-v2
-      - ref: scs-0214-v2
+      #~ - ref: scs-XXXX-v1
+      #~ - ref: scs-0210-v2
+      #~ - ref: scs-0214-v2
```
garloff, line 58: This change looks unrelated to me.

mbuechse:
Will be covered by #832.

PR #834, opened 2024-11-20
Revert any non-editorial changes to scs-0214-v1 that happened after s…

mbuechse:
…tabilization


garloff (approved):
These changes are in v2, but apparently ended up in v1 by mistake as well.
Thanks for spotting and fixing!

Issue #837, opened 2024-11-21
[BUG] subject_root variable missing


sluetze:
`./scs-compliance-check.py -v -a kubeconfig=path/to/kubeconfig.yaml -s SUBJECT scs-compatible-kaas.yaml`

does not work 
```
./scs-compliance-check.py -v -a kubeconfig=${HOME}/.kube/config -s OCP4.16 scs-compatible-kaas.yaml
CRITICAL: Missing variable assignments (via -a) for: subject_root
Traceback (most recent call last):
  File "/home/sluetzen/scs/standards/Tests/./scs-compliance-check.py", line 382, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/home/sluetzen/scs/standards/Tests/./scs-compliance-check.py", line 339, in main
    raise RuntimeError(f"Missing variable assignments (via -a) for: {', '.join(missing_vars)}")
RuntimeError: Missing variable assignments (via -a) for: subject_root
```

The Readme says to use it this way

https://github.com/SovereignCloudStack/standards/blob/main/Tests/README.md?plain=1#L34-L44

Running `./scs-compliance-check.py -v -a subject_root=./kubeconfig -s OCP4.16 scs-compatible-kaas.yaml`
where ./kubeconfig is a folder having `kubeconfig.yaml` works

garloff:
@mbuechse : Doc bug or code bug ?

mbuechse:
@garloff doc bug

mbuechse:
@garloff On second thought... Technically, it's a doc bug, but it's also just a questionable UX. I think I will try to find a way to make the UX better as well.

sluetze:
another thing (maybe a new issue?) the cncf compatibility checks are not done by the script for now. I do not know it this is expected (as the result says failed). Also I am missing any docs on how to run the cncf compatibility checks with the scs tooling.

sluetze:
Topic better UX: would be great if the tool would just use the `KUBECONFIG` env or look in default paths 
> The KUBECONFIG environment variable holds a list of kubeconfig files. For Linux and Mac, the list is colon-delimited. For  Windows, the list is semicolon-delimited. The KUBECONFIG environment variable is not required. If the KUBECONFIG environment variable doesn't exist, kubectl uses the default kubeconfig file, $HOME/.kube/config.

just needed to login for an updated kubeconfig, then rename it to kubeconfig.yaml...


[...]

PR #838, opened 2024-11-21
fix docs regarding used variable for kubeconfig

sluetze:
fixing https://github.com/SovereignCloudStack/standards/issues/837


mbuechse (commented):
Thanks for spotting this! I'm afraid this state of affairs is far from optimal. We should try to revert back to the old variable (as it's documented).

mbuechse:
superseded by #884

PR #839, opened 2024-11-21
allow versions which provide build info to be versionchecked

sluetze:
Some Distributions provide buildinfo as extension to the semver

```
<valid semver> ::= <version core>
                 | <version core> "-" <pre-release>
                 | <version core> "+" <build>
                 | <version core> "-" <pre-release> "+" <build>
```

this might result in version strings like `Kubernetes Version: v1.29.8+f10c92d`

currently, the version_check fails with "Unrecognized version format:"

This small change allows Kubernetes Distributions which provide the buildinfo to be versionchecked.

### DIFF
--- a/Tests/kaas/k8s-version-policy/k8s_version_policy.py
+++ b/Tests/kaas/k8s-version-policy/k8s_version_policy.py
@@ -171,7 +171,7 @@ K8sVersion.MINIMUM = K8sVersion(0, 0)
 
 
 def parse_version(version_str: str) -> K8sVersion:
-    cleansed = version_str.strip().removeprefix("v")
+    cleansed = version_str.strip().removeprefix("v").split("+")[0]
     try:
         major, minor, patch = cleansed.split(".")
         return K8sVersion(int(major), int(minor), int(patch))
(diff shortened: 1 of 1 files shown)


Tests/kaas/k8s-version-policy/k8s_version_policy.py
```diff
@@ -171,7 +171,7 @@ def __str__(self):
 
 
 def parse_version(version_str: str) -> K8sVersion:
-    cleansed = version_str.strip().removeprefix("v")
+    cleansed = version_str.strip().removeprefix("v").split("+")[0]
```
mbuechse, line 174: Thanks for spotting this!

```suggestion
    cleansed = version_str.strip().removeprefix("v").split("+")[0]  # remove leading v as well as build info
```

mbuechse: It still won't recognize `-prerelease`, but maybe pre-releases are not relevant here.

sluetze: not really into the scs policies, but from what I read it mostly targets operative environments. I would not expect any prerelease versions in such environments.

sluetze: I would also expect (even if it is not explicitly defined) to not require any prerelease versions as valid versions. As far as I understand, the goal is to have timely updates and not lag behind but stay on "upstream-supported" releases. using pre-releases will reduce stability and supportability.

mbuechse: Yes, I agree!

mbuechse (approved):

PR #841, opened 2024-11-22
Draft: Feat/node distribution implementation notes

tsmado:
A lot of time was spent trying to spawn a working cluster in the gx-scs, with the correct configuration and labelling, which in the end wasn't properly achieved due to some issues with loadbalancers, which we weren't able to resolve. There have been many changes and discussions about the [standard](https://github.com/SovereignCloudStack/standards/issues/639) during the work on the task. Once the standard is stabilised, further discussion will probably be necessary and the implementation could also be properly finalized. For now, it should be easy to manually add the `hostID` label to the nodes based on their underlying bare metal or virtual machine.


garloff (commented):
Ideally, this gets automated in a few lines of python (using the openstack and k8s SDKs) and becomes a plugin into the cluster creation process.

mbuechse:
I'm afraid this is obsolete :(

PR #842, opened 2024-11-25
fix: Test id not machting sonobuoy 'check_name'

tonifinger:
In the `scs-compatible-kaas.yaml`, the ID of the test case must match the `check_name` that is passed to the `sonobuoy_handler.py` to execute the test.
Otherwise the test will be listed as `MISSING` in the final report.


tonifinger:
> It must match the testcase-id, and that is the case here.
Ah I see I was looking at the wrong ID -.-
Sorry for the inconvenience, thanks for the quick reply!

PR #843, opened 2024-11-26
Stabilize SCS-compatible KaaS v1

mbuechse:


garloff (approved):
We have discussed and approved this in the SIG std/cert and I see no editorial reasons not to approve. Thus approved!

PR #844, opened 2024-11-26
Hotfix: syntax error in yaml file

mbuechse:


garloff (approved):
Ouch, I overlooked this in my previous review.
Thanks for spotting and fixing!

PR #846, opened 2024-12-03
Add cc-rrze

cpt-kernel-afk:
Initial commit for CC@RRZE

### DIFF
--- a/playbooks/clouds.yaml.j2
+++ b/playbooks/clouds.yaml.j2
@@ -119,3 +119,12 @@ clouds:
       application_credential_id: "{{ clouds_conf.wavestack_ac_id }}"
       application_credential_secret: "{{ clouds_conf.wavestack_ac_secret }}"
     auth_type: "v3applicationcredential"
+  cc-rrze:
+    region_name: "DE-ERL"
+    interface: "public"
+    identity_api_version: 3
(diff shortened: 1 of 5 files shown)


playbooks/clouds.yaml.j2
```diff
@@ -119,3 +119,12 @@ clouds:
       application_credential_id: "{{ clouds_conf.wavestack_ac_id }}"
       application_credential_secret: "{{ clouds_conf.wavestack_ac_secret }}"
     auth_type: "v3applicationcredential"
+  cc-rrze:
+    region_name: "DE-ERL"
+    interface: "public"
+    identity_api_version: 3
+    auth_type: "v3applicationcredential"
+    auth:
+      auth_url: AUTH_URL
```
berendt, line 128: This looks wrong.

cpt-kernel-afk: fixed with 8c50ef4c968c55ab3c0564bd4482c7b992c8deb7

mbuechse (approved):

mbuechse:
@cpt-kernel-afk Please note that I will have to update the compliance monitor with the new bootstrap file before it can accept reports from cc-rze. I will do that tonight (at 8 or 9).

PR #852, opened 2024-12-17
nominate Michael Bayr for the Project Board election

fkr:
discussed with and ok @michaelbayr


michaelbayr (approved):
thanks for your nomination @fkr - really appreciate it.

PR #859, opened 2024-12-20
Fix: introduce SCS-compatible IaaS v5.1

mbuechse:
- this now manifests the original intention properly
- postpone moving v4 into warn mode by one month
- as discussed by SIG Std/Cert on 2024-12-19


garloff (approved):
Looks good to me, thanks for fixing this up our oversight!

PR #867, opened 2025-01-10
Adjust PCO SCS IaaS-compatible checks.

garloff:
- Using AppCreds for Prod1 and Prod2 (was using usernm, passwd before)
- Name Secrets OS_ACSECRET_PCOPROD[1-4] rather than OS_PASSWORD_...
- Run v4 checks 5mins past v3 checks.

### DIFF
--- a/.github/workflows/check-pco-prod1-v4.yml
+++ b/.github/workflows/check-pco-prod1-v4.yml
@@ -1,7 +1,7 @@
 name: "Compliance IaaS v4 of pco-prod1"
 
 on:
-  # Trigger compliance check every day at 4:30 UTC
+  # Trigger compliance check every day at 4:35 UTC
   schedule:
     - cron:  '30 4 * * *'
   # Trigger compliance check after Docker image has been built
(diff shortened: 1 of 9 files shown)


mbuechse (approved):

.github/workflows/check-pco-prod1-v4.yml
```diff
@@ -1,7 +1,7 @@
 name: "Compliance IaaS v4 of pco-prod1"
 
 on:
-  # Trigger compliance check every day at 4:30 UTC
+  # Trigger compliance check every day at 4:35 UTC
   schedule:
     - cron:  '30 4 * * *'
```
berendt, line 6: comment does not match the schedule. I think this should be 35 instead of 30?

berendt (changes requested):

.github/workflows/check-pco-prod1-v4.yml
mbuechse:
```suggestion
    - cron:  '35 4 * * *'
```

garloff: @mbuechse suggestion fixed it.

berendt (approved):

PR #873, opened 2025-01-13
rotate app cred for wavestack

mbuechse:
keep trailing equals sign intact ;)


garloff (approved):
I can not check whether these are correct, as my brain's decryption functions are a bit weak ...
I assume they are.

PR #878, opened 2025-02-10
Add draft: SCS GitHub Organization - Management of Inactive Users and…

matofeder:


garloff (commented):
I like the spirit and the content of the document, thanks @matofeder.
Two observations:
1. Notifying inactive users by email is not necessarily well-defined.
  * A few users have a public email address on their github account
  * Most users (but not all) have accounts on our UCS/Nextcloud instance which gives us an email address (it may or may not still work ...)
  * Maybe we open a PR to remove that user on github, flagging the github handle and have some text with "please get in touch with us within 30d ..." and at the same time write an email IF we have an email address.
2. Maintainers of stale repositories may not be well-defined, until we have CODEOWNERS deployed everywhere. So we would open an issue, flagging the suspected interested people (most active contributors?) again plus emails if we can?

garloff (approved):
/lgtm

PR #879, opened 2025-02-11
feat: Rework domain-manager standard

gtema:
Life shows that not all CSPs are capable to adopt the domain manager
requirement due to certain architecture decisions or security requirement.
Suggest a rework that splits requirements into mandatory and recommended.

Signed-off-by: Artem Goncharov <artem.goncharov@gmail.com>

### DIFF
--- a/Standards/scs-0302-v2-domain-manager-role.md
+++ b/Standards/scs-0302-v2-domain-manager-role.md
@@ -1,0 +1,143 @@
+---
+title: Domain Manager adoption notes
+type: Standard
+track: IAM
+status: Draft
+replaces: scs-0302-v1-domain-manager-role.md
+---
+
+## Introduction
+
+After the domain manager persona has been implemented certain issues in the
+standard adoption and verification started being raised by CSPs:
+
+- Not every CSP is using domains to separate customers.
+
+- CSP may rely on the Identity federation in which case it is impossible or is
+prohibited to manage identities on OpenStack side (OpenStack is a service
+provider and not an identity provider).
+
+- CSP may customize authorization policies in a different way so that domain
+manager can not be implemented by simply reusing the upstream implementation.
+
+As such simple enforcement of the Domain Manager persona can not be achieved.
+
+This standard clarifies base standard and splits requirements into recommended
+and mandatory to provide better granularity while still giving guidance with
+the goal to provide a smooth user experience for the end users.
+
+Requiring customer to use CSP specific APIs to manage identity data is
+contradicting with the idea of certification as such. It hinders customers from
+having a smooth user experience across different cloud providers forcing them
+to adapt their management strategies on such clouds. Moreover it represent a
+lock-in what is contradicting with the idea of SovereignCloudStack.
+
+### Mandatory capabilities
+
+#### Assign roles to users/group on projects/domain
+
+One of the main initial concerns of the Domain Manager was the ability of the
+customer to manage user permissions in a self-service manner. OpenStack
+Keystone provides an easy possibility to smoothly integrate role assignments
+with arbitrary external systems in a transparent way (a role assignment backend
+plugin can be provided to persist assignments in any external system). As such
+this capability MUST be supported by the CSP using OpenStack APIs or role
+assignments.
+
+#### Project creation
+
+Another important requirement was to provide self-service capability for
+customers to create projects as desired without requesting CSP support. This
+capability MUST be available using native cloud APIs.
+
+#### Project editing
+
+Customer must be able to activate or deactivate project access without
+requesting CSP support. This capability provides possibility to temporarily
+disable users to authorize into certain project by modifying`enabled` property
+of the project. Further control of the project name, description, options and
+tags MUST be provided to the customer using native Keystone API.
+
+:::info
+
+Project deletion using Keystone API is not mandatory since CSP may have certain
+expectations on the resources cleanup. This requirement is described in detail
+in a dedicated chapter.
+
+:::
+
+### Recommended capabilities
+
+Relying on the Identity federation conceptually changes ways of identity
+resource management. This makes it impossible to fulfill them as MUST
+requirements. This chapter describes remaining capabilities of the initial
+Domain Manager as SHOULD implement.
+
+#### Domains usage
+
+It is strongly suggested to rely on the Domain concept of Keystone to implement
+multi-tenancy.
+
+Usage of domains by itself allows to implement form of self-service management
+by the customer. Only identity resources are owned by domains (with projects
+being also identity resources). Other services use projects to for resources
+isolation. They do not need to be domain aware.
+
+Using domains allows implementing [domain
+limits](https://docs.openstack.org/keystone/latest/admin/unified-limits.html#domain-limits)
+which allow to set a global resources limit for the customer. Without domains
+specific limits artificial control of the overall customer consumption must be
+implemented.
+
+#### User management
+
+User management (creation, activation, deactivation, deletion) SHOULD be
+possible using OpenStack APIs.
+
+When an external IdP is being used (IdP federation) there is still an
+expectation that local users may be required by customers. As such creation of
+users (pre-creating federated users or regular local users) within customer
+domain SHOULD me possible. Keystone does not allow certain operations on
+federated users (i.e. password change, MFA, name) as such allowing customers to
+manage users using OpenStack APIs should not conflict with any additional
+requirements.
+
+Security requirements on the customer side or on the CSP side to only allow
+federated users to consume platform services was used as the limitating factor
+forcing degrading of the capability requirement.
+
+#### Group management
+
+Management of groups SHOULD be possible using OpenStack APIs.
+
+In case of federated Identities there it is possible that groups on the IdP
+side do not match groups on the cloud provider side. In addition to that there
+might be a need to combine federated and local users. This would only be
+possible when groups are managed by the OpenStack.
+
+It is advised to keep user groups as mapped entities between external systems
+of CSP and Keystone. Upon user login (or using SCIM) user group relation may be
+synchronized between both platforms.
+
+#### Project deletion
+
+As described above project deletion may be implemented differently by CSPs.
+There are few ways of achieving that:
+
+- Forbid project deletion when resources (i.e. VMs) are still provisioned
+inside of those projects. This scenario assumes that the customer is
+responsible for cleaning projects before their deletion.
+
+- Automatically purge all project resources by the CSP when project deletion
+request is received. In this scenario CSP is implementing custom functionality
+to delete all resources before deleting the project.
+
+- Leave orphant resources. In this scenario project is being deleted by the API
+with custom cleanup procedures being responsible for dropping orphant resources.
+
+Leaving orphant resources MAY NOT be allowed.
+
+Forbidding project deletion making customer responsible for the cleanup SHOULD
+be preferred since it allows preventing the accidental deletion of the
+resources. Supplementary methods for purging project resources MAY be offered by
+the CSP.
--- a/Standards/scs-0302-v2-domain-manager-role.md
+++ b/Standards/scs-0302-v2-domain-manager-role.md
@@ -110,7 +110,7 @@ forcing degrading of the capability requirement.
 
 Management of groups SHOULD be possible using OpenStack APIs.
 
-In case of federated Identities there it is possible that groups on the IdP
+In the case of federated Identities the possibility exists that groups on the IdP
 side do not match groups on the cloud provider side. In addition to that there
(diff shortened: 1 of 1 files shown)


Standards/scs-0302-v2-domain-manager-role.md
```diff
@@ -0,0 +1,143 @@
...
+manager can not be implemented by simply reusing the upstream implementation.
+
+As such simple enforcement of the Domain Manager persona can not be achieved.
+
+This standard clarifies base standard and splits requirements into recommended
+and mandatory to provide better granularity while still giving guidance with
+the goal to provide a smooth user experience for the end users.
+
+Requiring customer to use CSP specific APIs to manage identity data is
+contradicting with the idea of certification as such. It hinders customers from
```
fkr, line 30: Somehow in my brain this sentence works better with standardization (since that comes before certification, which is just the attestation of successful standardization):

```suggestion
Requiring customer to use CSP specific APIs to manage identity data is
contradicting the idea of standardization  as such. It hinders customers from
```

Standards/scs-0302-v2-domain-manager-role.md
```diff
@@ -0,0 +1,143 @@
...
+
+Another important requirement was to provide self-service capability for
+customers to create projects as desired without requesting CSP support. This
+capability MUST be available using native cloud APIs.
+
+#### Project editing
+
+Customer must be able to activate or deactivate project access without
+requesting CSP support. This capability provides possibility to temporarily
+disable users to authorize into certain project by modifying`enabled` property
```
fkr, line 57:
```suggestion
disable users to authorize into certain project by modifying `enabled` property
```

Standards/scs-0302-v2-domain-manager-role.md
```diff
@@ -0,0 +1,143 @@
...
+requirements. This chapter describes remaining capabilities of the initial
+Domain Manager as SHOULD implement.
+
+#### Domains usage
+
+It is strongly suggested to rely on the Domain concept of Keystone to implement
+multi-tenancy.
+
+Usage of domains by itself allows to implement form of self-service management
+by the customer. Only identity resources are owned by domains (with projects
```
fkr, line 82:
```suggestion
Usage of domains by itself allows to implement a form of self-service management
by the customer. Only identity resources are owned by domains (with projects
```

Standards/scs-0302-v2-domain-manager-role.md
```diff
@@ -0,0 +1,143 @@
...
+being also identity resources). Other services use projects to for resources
+isolation. They do not need to be domain aware.
```
fkr, line 84:
```suggestion
being also identity resources). Other services use projects for resources
isolation. They do not need to be domain aware.
```

Standards/scs-0302-v2-domain-manager-role.md
```diff
@@ -0,0 +1,143 @@
...
+
+#### User management
+
+User management (creation, activation, deactivation, deletion) SHOULD be
+possible using OpenStack APIs.
+
+When an external IdP is being used (IdP federation) there is still an
+expectation that local users may be required by customers. As such creation of
+users (pre-creating federated users or regular local users) within customer
+domain SHOULD me possible. Keystone does not allow certain operations on
```
fkr, line 100:
```suggestion
expectation that local users may be required by customers. As such, creation of
users (pre-creating federated users or regular local users) within customer
domain SHOULD be possible. Keystone does not allow certain operations on
```

Standards/scs-0302-v2-domain-manager-role.md
```diff
@@ -0,0 +1,143 @@
...
+Security requirements on the customer side or on the CSP side to only allow
+federated users to consume platform services was used as the limitating factor
+forcing degrading of the capability requirement.
+
+#### Group management
+
+Management of groups SHOULD be possible using OpenStack APIs.
+
+In case of federated Identities there it is possible that groups on the IdP
+side do not match groups on the cloud provider side. In addition to that there
```
fkr, line 114:
```suggestion
In the case of federated Identities the possibility exists that groups on the IdP
side do not match groups on the cloud provider side. In addition to that there
```

Standards/scs-0302-v2-domain-manager-role.md
```diff
@@ -0,0 +1,143 @@
...
+with arbitrary external systems in a transparent way (a role assignment backend
+plugin can be provided to persist assignments in any external system). As such
+this capability MUST be supported by the CSP using OpenStack APIs or role
+assignments.
+
+#### Project creation
+
+Another important requirement was to provide self-service capability for
+customers to create projects as desired without requesting CSP support. This
+capability MUST be available using native cloud APIs.
```
berendt, line 51: This will lead to problems and effort for providers who have external processes on projects.

Standards/scs-0302-v2-domain-manager-role.md
```diff
@@ -0,0 +1,143 @@
...
+
+#### Project editing
+
+Customer must be able to activate or deactivate project access without
+requesting CSP support. This capability provides possibility to temporarily
+disable users to authorize into certain project by modifying `enabled` property
+of the project. Further control of the project name, description, options and
+tags MUST be provided to the customer using native Keystone API.
```
berendt, line 59: Same.

fzakfeld: This will also cause a stale state between Keystone and the CSPs source of truth, unless we build even more CSP-specific glue for backsynching this.

I see why this is important, but realistically and according to various certification requirements, this needs to be implemented at a CSP level. Let's say Bob gets their account disabled `enabled=False` in Keystone. Bob can still open a support case with the CSP to destroy resources. Having disabled the account gives some false security to whoever did that.

fzakfeld (changes requested):

Standards/scs-0302-v2-domain-manager-role.md
gtema: I do not think that somebody is preventing user to create K8 namespace from kubernetes directly. Why should all of a sudden this is then causing problems on OpenStack? In my eyes this is just a wrong integration. As an OpenStack user (on a daily basis) I am incredibly frustrated by inability to create projects using native OpenStack API. I come to the CSP for OpenStack UX and I can't use it.
OpenStack emits audit messages for all major events (projects creation/deletion as well). Proper integration would be catching such events and allocating resources in the "external systems" as necessary without influencing UX

Standards/scs-0302-v2-domain-manager-role.md
gtema: I strongly disagree here. If the process is working like that - it is wrong. I call your support, request deletion of a server in your account and they just do it? CSP MUST implement validation of support requests. Imagine I am a customer bringing my own LDAP with 1000 users. Does that automatically mean I can not manage them on my side? Or does that mean that any of those 1000 users can request whatever they want?

Standards/scs-0302-v2-domain-manager-role.md
```diff
@@ -0,0 +1,143 @@
...
+
+#### Assign roles to users/group on projects/domain
+
+One of the main initial concerns of the Domain Manager was the ability of the
+customer to manage user permissions in a self-service manner. OpenStack
+Keystone provides an easy possibility to smoothly integrate role assignments
+with arbitrary external systems in a transparent way (a role assignment backend
+plugin can be provided to persist assignments in any external system). As such
+this capability MUST be supported by the CSP using OpenStack APIs or role
+assignments.
```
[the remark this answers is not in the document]
gtema: It's technically a 100-200 lines of python code depending on the desired integration. I repeat - I (user) come for OpenStack UX, why can't I have it? When CSP is not capable to integrate OpenStack into their business system without disrupting the designed flows they simply should not do this at all.

Standards/scs-0302-v2-domain-manager-role.md
gtema: As a user spending 50% of my day in OpenStack managing complex infrastructures always have the necessity to manage projects on my own without exiting boundaries of OpenStack. I (with my customer hat on) am always incredibly frustrated and angry when I can't do this.
Purpose of the standard is to serve customer and not to protect OEM, otherwise we would all still live now in the world with 40 different chargers for mobile phones.

Standards/scs-0302-v2-domain-manager-role.md
fzakfeld: As a CSP we handle authorisation of user accounts in systems other than keystone. I'd argue that this is common practise pretty much everywhere. When a support case comes in, the support system shows what the requestor is authorised do to.

We can and will not change this. It is part of multiple certifications. But discussing business processes should be beyond the scope of SCS anyways.

Standards/scs-0302-v2-domain-manager-role.md
fzakfeld: Perhaps it would be useful if other CSPs can give some insights here, for us we don't really have customers that feel the way about this as you do. Only from "don't need" to "nice to have"

fkr:
From reading the discussion (and debating this in today's PB meeting), I think the way forward to deal with this is to RECOMMEND it and not use MUST, since from what I gather from @gtema's comments is that we the desired features are anticipated and we want CSPs to move towards that.
We've said in the past that we use RECOMMENDATIONS for these cases, since that way CSPs can adopt certain best practices.

garloff:
A few further thoughts, building on top of what @fkr said:
- We have the goal of ensuring offering standardized self-service APIs for users; this reduces the pain for users when dealing with CSPs and in particular when dealing with multiple SCS-compatible CSPs.
- When a certain functionality that is relevant for many users is covered by an OpenStack API, then this would typically be the API that we standardize on; we did investigate whether we want to outsource a relevant chunk of IAM functions to keycloak APIs, but that did not turn out to be agreeable by our ecosystem.
- Getting there may be a multi-step process; some functions we may consider a baseline -- maybe read-only acccess to lists of projects and users and groups and maybe read-write for role assignments ... this gives something that users and workloads, processes, ... designed to work on all SCS compatible can rely on. I imagine e.g. a security assessment tool.  These would be the MANDATORY requirements now. I would hope that all of our partners can implement these without major pain. MANDATORY is the common baseline and it's black and white; either you 100% comply with MANDATORY requirements and you deserve the SCS-compatible label or you don't. This limits in how high we can put the bar, because a standard that only a small subset of our partners are able to meet would not provide value to users either.
- Some things we agree we want may be RECOMMENDED for now; we document what we consider best practice and support our not-yet CSP partners to move into this direction. Things that are RECOMMENDED now may become MANDATORY in the future. (In other words: all recommended requirements are candidates for future mandatory requirements; when such a promotion happens, we would still give our partners some time to adapt.)
- We also want transparency for the RECOMMENDED sections of our standards; so we want to included these in the tests and make their fulfillment transparent in our compliance monitor. Already discussed some ideas with @mbuechse how to do this. This would serve two purposes: Create an incentive for providers to fulfill RECOMMENDED requirements and also help users to understand their provider choices when they need one of the RECOMMENDED properties. Ideally, we'd have some systematic feedback channel; would need to think how to do this best.

That said, I could agree with demoting a few things that I really consider useful to RECOMMENDED for now.

Standards/scs-0302-v2-domain-manager-role.md
```diff
@@ -0,0 +1,143 @@
...
+- Forbid project deletion when resources (i.e. VMs) are still provisioned
+inside of those projects. This scenario assumes that the customer is
+responsible for cleaning projects before their deletion.
+
+- Automatically purge all project resources by the CSP when project deletion
+request is received. In this scenario CSP is implementing custom functionality
+to delete all resources before deleting the project.
+
+- Leave orphaned resources. In this scenario project is being deleted by the API
+with custom cleanup procedures being responsible for dropping orphant resources.
```
garloff, line 136:
```suggestion
with custom cleanup procedures being responsible for dropping orphaned resources.
```
Typo.

Standards/scs-0302-v2-domain-manager-role.md
```diff
@@ -0,0 +1,143 @@
...
+
+Leaving orphaned resources MAY NOT be allowed.
+
+Forbidding project deletion making customer responsible for the cleanup SHOULD
+be preferred since it allows preventing the accidental deletion of the
+resources. Supplementary methods for purging project resources MAY be offered by
+the CSP.
```
garloff, line 143: To avoid a race condition (be it with checking for existing resources to reject project deletion if there are still resources or be it with deleting the resources automatically), we would need a two phase commit:
(a) mark the project as read-only, so no new resources can be created
(b) do the checks/cleanups
(c) finally remove the project

mbuechse:
No consensus in sight

PR #881, opened 2025-02-13
Supplement for scs-0004: regulations for achieving certification

mbuechse:
Just to keep this alive


Standards/scs-0004-w1-achieving-certification-implementation.md
```diff
@@ -0,0 +1,133 @@
...
+4. The cloud can be listed on the SCS pages as *SCS-compatible* with a compatibility status that is
+   updated on a daily basis. SCS then tests the infrastructure on a daily basis.
+
+The precise rules that govern how certificates are issued or withdrawn are defined in the
+[SCS standard 0004](scs-0004-v1-achieving-certification.md).
+
+## Self-testing and technical adjustments
+
+In order for a cloud service offering to obtain a certificate, it has to
+conform to all standards of the respective scope, which will be tested at
```
garloff, line 36:
```suggestion
conform to all mandatory aspects of all standards of the respective scope, which will be tested at
```
Be more precise -- there can be *recommended* aspects and they are no prerequisite for passing.
We could go into more detail here and say that the recommendations should be taken seriously and providers should be prepared to answer customers why they don't fulfill these ...
But I preferred to keep it short here.

garloff:
I think this is good to be merged. Ideally with my suggestion merged that mentions the mandatory aspects of a standard. (Feel free to use better language, e.g. mandatory requirements.)
As quite some of the text stems from yours truly, I don't want to formally approve this.

fkr (approved):

mbuechse:
JUST NO!!!

PR #888, opened 2025-03-06
brush up mandatory-iaas-services.py

### LINKED ISSUE #862: [BUG] brush up mandatory-iaas-services.py
The [script](https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/mandatory-services/mandatory-iaas-services.py) has some issues still:

- it uses `assert`
- some code is commented out
- it uses print instead of logger
- it sometimes uses warning instead of error
- it enables OpenStack debug logging (this should at least be a different command-line switch)

mbuechse:
resolves #862

### DIFF
--- a/Tests/iaas/mandatory-services/mandatory-iaas-services.py
+++ b/Tests/iaas/mandatory-services/mandatory-iaas-services.py
@@ -77,8 +77,7 @@ def s3_conn(creds, conn=None):
         cacert = conn.config.config.get("cacert")
         # TODO: Handle self-signed certs (from ca_cert in openstack config)
         if cacert:
-            print("WARNING: Trust all Certificates in S3, "
-                  f"OpenStack uses {cacert}", file=sys.stderr)
+            logger.warning(f"Trust all Certificates in S3, OpenStack uses {cacert}")
             vrfy = False
     return boto3.resource('s3', aws_access_key_id=creds["AK"],
                           aws_secret_access_key=creds["SK"],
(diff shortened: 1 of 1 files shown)


Tests/iaas/mandatory-services/mandatory-iaas-services.py
```diff
@@ -179,20 +176,22 @@ def check_for_s3_and_swift(conn: openstack.connection.Connection, s3_credentials
...
 
     # If we got till here, s3 is working, now swift
     swift_containers = list_containers(conn)
     # if not swift_containers:
     #    swift_containers = create_container(conn, TESTCONTNAME)
     result = 0
     if Counter(s3_buckets) != Counter(swift_containers):
-        logger.warning("S3 buckets and Swift Containers differ:\n"
-                       f"S3: {sorted(s3_buckets)}\nSW: {sorted(swift_containers)}")
+        logger.error("S3 buckets and Swift Containers differ:\n"
```
garloff, line 193: Thanks for catching this. As we return 1, indicating an ERROR, we should also log an ERROR. Thanks for catching!

Tests/iaas/mandatory-services/mandatory-iaas-services.py
```diff
@@ -235,7 +234,7 @@ def main():
         format="%(levelname)s: %(message)s",
         level=logging.DEBUG if args.debug else logging.INFO,
     )
-    openstack.enable_logging(debug=args.debug)
+    openstack.enable_logging(debug=False)
```
garloff, line 237: Why?
If we enable debug with `--debug`, we should honor it, no?

garloff: Comparing the outputs: Without the change, we get some debugging from both openstacksdk and boto(s3), with it only from s3.

garloff (approved):
Good with me.

Tests/iaas/mandatory-services/mandatory-iaas-services.py
garloff: Yes, makes sense.

Issue #890, opened 2025-03-07
[BUG] mandatory-services test fails with access error on wavestack, pco1--4


garloff:
The testcase `mandatory-iaas-services.py`, included in IaaS v5[.1] scope fails on some clouds.
The error message indicates that the ec2 credential is not valid or fails to access the buckets/containers from swift.

garloff:
ec2 credentials belong to a user and grant access to the S3 storage belonging to the project it was requested for.
For wavestack and for PCO 1--4, we have EC2 creds that granted S3 access for a different project than the we one we use for testing, causing access failures.
Cleaning up the ec2 creds manually resolved the issue.

garloff:
We probably should always create a new ec2 credential and clean it up after usage rather than just reusing the old one, which could be stale in case something has changed ...
Let's do this in conjunction with #888

garloff:
Alternatively, when trying to reuse existing ec2 creds, we could filter for ones that match the correct project.

PR #892, opened 2025-03-13
Only use EC2 creds from our project. Clean up after use.

garloff:
When looking for existing EC2 credentials, we only accept those created for our own project.
If we need to create one, remember this and clean it up again.

### DIFF
--- a/Tests/iaas/mandatory-services/mandatory-iaas-services.py
+++ b/Tests/iaas/mandatory-services/mandatory-iaas-services.py
@@ -143,7 +143,7 @@ def s3_from_ostack(creds, conn, endpoint):
     if found_ec2:
         creds["AK"] = found_ec2["access"]
         creds["SK"] = found_ec2["secret"]
-        return None
+        return
     # Generate keyid and secret
     ak = uuid.uuid4().hex
     sk = uuid.uuid4().hex
@@ -152,13 +152,12 @@ def s3_from_ostack(creds, conn, endpoint):
         crd = conn.identity.create_credential(type="ec2", blob=blob,
                                               user_id=conn.current_user_id,
                                               project_id=conn.current_project_id)
-        creds["AK"] = ak
-        creds["SK"] = sk
-        return crd.id
-    except BaseException as excn:
-        logger.warning(f"ec2 creds creation failed: {excn!s}")
-        # pass
-    return None
+    except BaseException:
+        logger.warning(f"ec2 creds creation failed", exc_info=True)
+        return
+    creds["AK"] = ak
+    creds["SK"] = sk
+    return crd.id
 
 
 def check_for_s3_and_swift(conn: openstack.connection.Connection, s3_credentials=None):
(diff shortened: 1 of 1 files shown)


garloff:
This should avoid issue #890.

Tests/iaas/mandatory-services/mandatory-iaas-services.py
```diff
@@ -189,19 +195,24 @@ def check_for_s3_and_swift(conn: openstack.connection.Connection, s3_credentials
...
+    # No need to clean up swift container, as we did not create one
+    # (If swift and S3 agree, there will be a S3 bucket that we clean up with S3.)
     # if swift_containers == [TESTCONTNAME]:
     #    del_container(conn, TESTCONTNAME)
     # Cleanup created S3 bucket
     if s3_buckets == [TESTCONTNAME]:
         del_bucket(s3, TESTCONTNAME)
+    # Clean up ec2 cred IF we created one
+    if ec2_cred:
+        conn.identity.delete_credential(ec2_cred)
```
mbuechse, line 215: This may still leak the credential if we script is somehow interrupted. Would there be a safe way of removing credentials that can be inferred to be from this test, or do they expire at some point?

garloff: The damage is low though. If there is a matching credential when the script starts the next time, it will be used, so we don't up creating more and more credentials. The worst thing that could happen is that this one credential becomes stale at some point and the admin needs to remove it.

We can not differentiate between a credential that the test should use and that the admin created on purpose and one that got leaked by the script being interrupted prior to cleaning up the credential.

There's a few things that could be done, but I don't like either particularly well:
1. We could just clean up preexisting ec2 credentials, assuming that we have created them ourselves previously. At least if the S3_* variables are not set. This removes some flexibility, but probably the ability to set `S3_SECRET_ACCESS_KEY` and `S3_ACCESS_KEY_ID` is enough flexibility, so I would find this acceptable. But deleting things we have possibly not created is not normally good practice, so this needs some warning in the docs
2. We could set up a signal handler that does the deletion of the created ec2 credential. This will work in some cases (when the process properly gets signaled with SIGINT or SIGTERM) and not in others (SIGKILL or container just being removed ...). Implementing this is not particularly hard, so the usefulness vs. complexity is probably still OK.
3. We could add some magic to clean up a preexisting ec2 credential if and only if it fails to work, so we have some auto-healing capability in case there are stale credentials. Of course, intermittent failures would then also cause admin-provided credentials to be deleted, so we still have to warn users ...

Thinking about it a bit more, I think 2 is a reasonable approach. Not 100% failsafe, but reducing the chances.

garloff: OK, implemented in https://github.com/SovereignCloudStack/standards/pull/892/commits/30af575b715d237042aae9051587ddfb50a6582b

mbuechse (approved):
LGTM. I have changed a few bits; the commit messages should explain what I did.

PR #897, opened 2025-04-02
Fix/workaround secret leak

### LINKED ISSUE #896: [BUG] Secrets are not cleaned up in compliance test
If find many many secrets with the name `scs-member-role-test-secret` in the projects that are used to run compliance tests.

garloff:
This fixes #896.


garloff:
OK, my git tree was fscked up.

PR #898, opened 2025-04-02
Fix/workaround secret leak2

### LINKED ISSUE #896: [BUG] Secrets are not cleaned up in compliance test

garloff:
Fixes #896

### DIFF
--- a/Tests/iaas/key-manager/check-for-key-manager.py
+++ b/Tests/iaas/key-manager/check-for-key-manager.py
@@ -77,7 +77,16 @@ def check_key_manager_permissions(conn: openstack.connection.Connection) -> None
     try:
         existing_secrets = _find_secret(conn, secret_name)
         for secret in existing_secrets:
-            conn.key_manager.delete_secret(secret.id[secret.id.rfind('/')+1:])
+            # Workaround for SDK bugs:
+            # - The id field in reality is a href (containg the UUID at the end)
+            # - The delete_secret() function contrary to the documentation does
+            #   not accept openstack.key_managerv1.secret.Secret objects nor the
+            #   hrefs, just plain UUIDs.
+            uuid_part = secret.id.rfind('/') + 1
+            if uuid_part != 0:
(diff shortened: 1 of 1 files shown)


gtema (approved):
Right, I knew this is very awkward on the barbican side (not supporting I'd as such), but also on SDK side I never worked with the service

Tests/iaas/key-manager/check-for-key-manager.py
```diff
@@ -57,17 +57,42 @@ def check_presence_of_key_manager(conn: openstack.connection.Connection) -> None
...
+       an object or href, just silently does nothing.
+    The code here assumes that the SDK (when fixed) will continue to
+    accept UUIDs as argument for delete_secret() in the future.
+    Code is robust against those being passed directly in the .id attr
+    of the objects. (It would be even more robust to try to pass the
+    object first, then the href, then the UUID extracted from the href,
+    each time checking whether it was effective. But that's three delete
+    plus list calls and very ugly.)
+    """
+    uuid_part = secret.id.rfind('/') + 1
```
mbuechse, line 91: https://docs.python.org/3/library/stdtypes.html#str.rsplit

I think the case distinction can be dropped by using rsplit with maxsplit=1

```python
uuid = secret.id.rsplit('/', 1)[0]
```

Should work in both cases

garloff: Except that it should be `secret.id.rsplit('/')[-1]`.

garloff: ... which is the same as `secret.id.rsplit('/',1)[-1]`, except that the latter might be slightly cheaper as it creates a list with max 2 elements.

mbuechse: yeah, you're right, I wrote this in a hurry :see_no_evil:

PR #901, opened 2025-04-09
Docusaurus does not seem to like <links> in ref lists.

garloff:
github renders these well, but docusaurus breaks the build.

This breaks the docs page build, see https://github.com/SovereignCloudStack/docs/issues/307.

Patch replaces the links by classic `[Word](link)` style. There may be better workarounds, but we need the docs page building now to get R8 release notes published.


garloff:
Merging this for now to fix the build.
If this should be done in a better way, we need to do so in a followup patch.

garloff:
(And it works.)

PR #902, opened 2025-04-10
Increase timeout to 80s. Backoff poll frequency.

garloff:
Timeout was 60s, increase to 80s. This may help on heavily loaded storage environments.

We see ~60 debug messages for each timed out wait. This is more than needed and puts a bit more load on the control plane than needed. So change the approach: Start with 0.5s polling freq but increase wait time by 0.1s each time, slowly backing off. So we produce only 35 debug lines and API calls before running into the timeout.

Sidenote: Exponential backoff is a well-known approach to deal with congestion. We kept it simple (linear back-off), as our timeout is not huge.

### DIFF
--- a/Tests/iaas/volume-backup/volume-backup-tester.py
+++ b/Tests/iaas/volume-backup/volume-backup-tester.py
@@ -233,7 +233,7 @@ def main():
         "with the prefix specified via '--prefix' (or its default)"
     )
     args = parser.parse_args()
-    openstack.enable_logging(debug=args.debug)
+    openstack.enable_logging(debug=False)
     logging.basicConfig(
         format="%(levelname)s: %(message)s",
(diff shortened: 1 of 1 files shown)


garloff:
This is a minor improvement only.
Won't address the fails on cnds, maybe more luck on aov.

garloff:
Well, it's hard to predict the time it will take.
As the storage system can do things like copy-on-write for backups, this may well be done within 2s (which is basically the call overhead).
If on the other hand, the backup cluster is separated from the main storage, it will have to copy over data. For 1GB, this could reasonably take up to 10s (100MB/s effective copy rate). Now in a busy system, there may be congestion or internal retries, so we should not give up too early.
So we could do some waiting, 2+3+5+10+15+25+40 = 100s and then consider it failed.

mbuechse (approved):

PR #914, opened 2025-05-08
Bugfix: image-md-check.py too lenient, scs-0102-v1 too strict

### LINKED ISSUE #913: [BUG] image-md-check.py too lenient?
I see output like this all the time:

```text
WARNING: Image "Windows 2022 - 01/25": min_ram == 0 MiB < 64 MiB
WARNING: Image "Windows 2022 - 01/25": img size == 40960 MiB, but min_disk == 0 MiB
WARNING: Image "Ubuntu 24.04 (20240912)": empty value for property "hypervisor_type" not recommended
WARNING: Image "Ubuntu 24.04 (20240912)": empty value for property "hw_rng_model" not recommended
```

Problems:

- `min_ram` and `min_disk` are mandatory and should be set to something apart from 0?
- ~~`hypervisor_type` is mandatory as well and doesn't make much sense if empty?~~ (see comment below)
- `hw_rng_model` is mandatory, except this requirement is a bit strict and should be relaxed in the standard
- same for `hypervisor_type`

mbuechse:
resolves #913

### DIFF
--- a/Standards/scs-0102-v1-image-metadata.md
+++ b/Standards/scs-0102-v1-image-metadata.md
@@ -59,6 +59,8 @@ The following properties are considered mandatory:
 
 **Note**: Disk sizes tend to be measured in GB = 10^9 and not GiB = 2^30 in the disk industry, but OpenStack uses GiB.
 
+The value given vor `min_ram` MUST be sufficient for the VM to boot and survive for 10 s.
+
 The following property is recommended:
 
 - `hypervisor_type`
--- a/Standards/scs-0102-v1-image-metadata.md
+++ b/Standards/scs-0102-v1-image-metadata.md
@@ -59,7 +59,7 @@ The following properties are considered mandatory:
 
 **Note**: Disk sizes tend to be measured in GB = 10^9 and not GiB = 2^30 in the disk industry, but OpenStack uses GiB.
 
-The value given vor `min_ram` MUST be sufficient for the VM to boot and survive for 10 s.
+The value given vor `min_ram` MUST be sufficient for the VM to boot and survive (for at least 10 s).
 
 The following property is recommended:
 
--- a/Standards/scs-0102-v1-image-metadata.md
+++ b/Standards/scs-0102-v1-image-metadata.md
@@ -52,15 +52,20 @@ in the [OpenStack Image documentation](https://docs.openstack.org/glance/latest/
 
 The following properties are considered mandatory:
 
-- `architecture`, `hypervisor_type`
+- `architecture`
 - `min_disk` (in GiB), `min_ram` (in MiB)
 - `os_version`, `os_distro`
-- `hw_rng_model`, `hw_disk_bus` (`scsi` recommended, and then setting `hw_scsi_model` is also recommended)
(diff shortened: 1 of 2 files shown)


Tests/iaas/image-metadata/image-md-check.py
```diff
@@ -282,12 +282,16 @@ def validate_imageMD(img, outd_list):
     if img.min_ram < 64:
         print(f'WARNING: Image "{imgnm}": min_ram == {img.min_ram} MiB < 64 MiB', file=sys.stderr)
         warnings += 1
-        # errors += 1
+    if not img.min_ram:
+        print(f'ERROR: Image "{imgnm}": min_ram == 0', file=sys.stderr)
+        errors += 1
     if img.min_disk * GIB < img.size:
```
garloff, line 288: One could argue that we should throw an error if
* if min_disk is 0 OR
* if min_disk is smaller than the image_size
```python
    if not img.min_disk or img.min_disk * GIB < img.size:
        print(f'ERROR: Image "{imgnm}": Too small min_disk == {img.min_disk * GiB/MiB:.0f} MiB vs. img_size == (img.size / MiB:.0f} MiB', file=sys.stderr)
        errors += 1
```
This would also be a hard rejection of too small sizes that can't be true and thus catch operator errors.

mbuechse: I agree that this would be good, but I suppose we would have to really change the standard for that. In the meantime, the warning could catch operator errors as well (if less effectively).

garloff: The other comment: We could add a sentence that the specified min_ram must be sufficient for the VM to boot and survive for 10s.
(Why 10s? Because scheduling latency plus boot time is in the order of a few seconds which makes it unreasonable to spawn VMs that only survive a single-digit number of seconds. You could do that with containers. If we ever wanted to create a test case, we also know that we don't need to wait very long. As a generic (OS-agnostic) health check is hard to create, we may never test for this. Maybe responding to arp requests is a valid OS-agnostic test? But of course we can hold providers accountable in case someone reports failure ...)

garloff: > I agree that this would be good, but I suppose we would have to really change the standard for that. In the meantime, the warning could catch operator errors as well (if less effectively).

It goes without saying that the metadata provided by the operator must reflect the truth and work, no?
So rejecting things hard that can't be true seems legit to me.

Standards/scs-0102-v1-image-metadata.md
```diff
@@ -59,6 +59,8 @@ The following properties are considered mandatory:
 
 **Note**: Disk sizes tend to be measured in GB = 10^9 and not GiB = 2^30 in the disk industry, but OpenStack uses GiB.
 
+The value given vor `min_ram` MUST be sufficient for the VM to boot and survive for 10 s.
```
garloff, line 62:
```suggestion
The value given vor `min_ram` MUST be sufficient for the VM to boot and survive (for at least 10s).
```

garloff (approved):
Looks good, thanks for this improvement, @mbuechse !

Issue #889, opened 2025-03-06
[Feature Request] record aborted testcases


garloff:
Tests that don't finish don't have a link to a log file, which is bad.
    - Reason is that the script searches for the last result (PASS/FAIL), if test did not yield any results, it is not linked
    - Solution: I no data is found, link to last attempt

mbuechse:
I'm afraid the problem is more severe: the testcases can be missing in the report, and then, of course, they can't be found in the database.

A check script can output results for multiple testcases, and if its execution is cut short (by a badly handled exception for instance), then some testcases can be missing. (This holds true for the case that the number of testcases is 1.) We don't know prima facie that these testcases are missing, because we don't know that the script is supposed to output them; they could fall under the responsibility of some other script, or they could be missing intentionally because they are expensive to check and not checked every time.

I wouldn't want to make the specs more complicated by having them include which testcase is to be expected where and when. What I would do though:

- improve the quality of the check scripts throughout to make sure each testcase is reported on so long as it was intended to be checked
- add a third result besides "PASS" and "FAIL", namely "ABORT" (or similar)
- insert "ABORT" results into the database (recall that previously they would be ignored)

With these measures in place, there is a much higher chance that the report contains all relevant testcases, and we can distinguish "ABORT" from missing. This distinction is only relevant for debugging: both results have the same effect for the status of the subject under test, but the former one means that debug logs are present.

depressiveRobot:
Ah ok, sorry for that.

mbuechse:
With #1228, this issue is complete for now. The tests regarding the default storage class do need work (in multiple respects), but they aren't part of the scope yet.

PR #915, opened 2025-05-09
Improve docs on table/table_full, increase transparency

### LINKED ISSUE #904: compliance monitor: restructure table/table_full and permissions
The compliance monitor offers two similar views:

* https://compliance.sovereignit.cloud/page/table -- no HTTP auth, displays verified results, no access to reports
* https://compliance.sovereignit.cloud/page/table_full -- HTTP basic auth, displays all results (even unverified), access to reports

This is not well documented (see original bug report below) and less transparent than necessary. As decided in [SIG Std/Cert](https://github.com/SovereignCloudStack/minutes/blob/main/sig-standardization/20250508.md#compliance-monitor-table-vs-table_full):

* no auth should be required regardless of results are verified or not
* unverified results MUST be recognizable as such (with an admonition that they may contain false positives)
* reports should be made available without auth in a curated form (containing only those log lines that are explicitly linked to any testcase)
* the only gated view then would be the full report with the detailed logs

From my (mbuechse's) POV, this means:

1. make table view switchable between verified results and all results by including a corresponding link; I suppose we can use a simple query parameter for the `table` endpoint and replace `table_full` with an HTTP forward
2. as mentioned, make it very apparent at first glance what kind of results are displayed
3. include a link from the curated report to the full report, mentioning that the latter needs a login; this will most likely require introducing a new endpoint (because making auth optional doesn't work well with browsers)

---

**original bug report:**

Currently, the CNDS cloud fails volumes backups most of the time.
Without the volume backup test, it can still pass v4 of SCS-compatible IaaS, but not v5.1.
Currently, https://compliance.sovereignit.cloud/page/table shows CNDS as v5.1 compliant, where as https://compliance.sovereignit.cloud/page/table_full shows v4 and upon digging into the details, I can indeed see the volume backup test last night failed.

Questions/Remarks:
1. I would expect those two tables to come to the same conclusions. If that is not the case, some documentation that explains why not would be helpful. So, maybe I did not find the right documentation, or there is a documentation bug or a code bug.
2. We are hiding the logs from normal observers behind a long for the table_full. Do we really need to? Maybe better invest some extra care to ensure we can not possibly leak credentials, so we no longer need to hide logs? Or are there other valid reasons to hide them?

mbuechse:
resolves #904

### DIFF
--- a/compliance-monitor/monitor.py
+++ b/compliance-monitor/monitor.py
@@ -556,7 +556,8 @@ async def get_status(
 def _build_report_url(base_url, report, *args, **kwargs):
     if kwargs.get('download'):
         return f"{base_url}reports/{report}"
-    url = f"{base_url}page/report/{report}"
+    report_page = 'report_full' if kwargs.get('full') else 'report'
+    url = f"{base_url}page/{report_page}/{report}"
     if len(args) == 2:  # version, testcase_id --> add corresponding fragment specifier
         url += f"#{args[0]}_{args[1]}"
     return url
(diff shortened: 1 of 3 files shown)


mbuechse:
Screenshot from 2025-05-09 14-57-57
Screenshot from 2025-05-09 14-58-03
Screenshot from 2025-05-09 14-58-50
Screenshot from 2025-05-09 14-58-55

mbuechse:
Be advised that it will only really make sense when #889 is resolved as well.

fkr (approved):

PR #918, opened 2025-05-16
Factor out requirements for iaas/kaas

### LINKED ISSUE #864: [BUG] requirements for iaas/kaas
We do have separate requirements for kaas (such as kubernetes or junitparser), but we should also separate out those for iaas (such as boto3).

mbuechse:
resolves #864


mbuechse:
apparently, it doesn't work this way, because pip-compile computes pinned versions that can contradict each other in between files.

mbuechse:
it does work, but pip-compile --upgrade has to be run on all files if one file is upgraded

martinmo (approved):

fkr (approved):

PR #924, opened 2025-05-20
Bugfix: install sonobuoy executable in playbook

mbuechse:
Questions:

- Can we assume that `.local/bin` is on PATH? Otherwise, it would have to be added.
- Do you have a better idea?

Note that this method is what is officially suggested. I have found Ansible roles for installing Sonobuoy that use some obscure Debian repository, but most such roles were actually no longer available.

### DIFF
--- a/.zuul.d/config.yaml
+++ b/.zuul.d/config.yaml
@@ -35,6 +35,8 @@
       kaas: false
       do_provision: false
       do_cleanup: true
+      sonobuoy_tar_gz_url: >
+        https://github.com/vmware-tanzu/sonobuoy/releases/download/v0.57.3/sonobuoy_0.57.3_linux_amd64.tar.gz
     pre-run:
      - playbooks/pre.yaml
      - playbooks/pre_cloud.yaml
--- a/playbooks/pre_cloud.yaml
+++ b/playbooks/pre_cloud.yaml
@@ -19,7 +19,7 @@
 
     - name: Create local binary dir
       ansible.builtin.file:
-        path: "~/.local/bin"
+        path: "{{ ansible_user_dir}}/.local/bin"
         state: directory
         recurse: true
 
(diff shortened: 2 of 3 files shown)


playbooks/pre_cloud.yaml
```diff
@@ -17,6 +17,20 @@
...
 
+    - name: Create local binary dir
+      ansible.builtin.file:
+        path: "~/.local/bin"
+        state: directory
+        recurse: true
+
+    - name: Install Sonobuoy
+      ansible.builtin.shell:
+        curl -L https://github.com/vmware-tanzu/sonobuoy/releases/download/v0.57.3/sonobuoy_0.57.3_linux_amd64.tar.gz | tar xz sonobuoy
```
berendt, line 28: I think it makes sense to move the URL + the Version into vars. It could also make sense to make the used arch configurable.

mbuechse: something like

```suggestion
        curl -L {{ sonobuoy_tar_gz_url | quote }} | tar xz sonobuoy
```

??

.zuul.d/config.yaml
```diff
@@ -35,6 +35,8 @@
       kaas: false
       do_provision: false
       do_cleanup: true
+      sonobuoy_tar_gz_url: >
+        https://github.com/vmware-tanzu/sonobuoy/releases/download/v0.57.3/sonobuoy_0.57.3_linux_amd64.tar.gz
```
berendt, line 39: I do it this way:

Definition of sonobuoy_version : https://github.com/regiocloud/e2e-tests/blob/main/playbooks/gardener/pre.yml#L15

wget and removal: https://github.com/regiocloud/e2e-tests/blob/main/playbooks/gardener/pre.yml#L58-L60

But I think it makes sense to have a sonobuoy_url like https://github.com/vmware-tanzu/sonobuoy/releases/download/v{{ sonobuoy_version }}/sonobuoy_{{ sonobuoy_version }}_linux_amd64.tar.gz here. This way it's easier to use a local mirror.

janiskemper (approved):
Approving after communicating with @mbuechse that possible improvements will be part of future PRs, to see results quickly.

Issue #943, opened 2025-06-03
[Feature Request] combine compliance results for regions/k8s branches


mbuechse:
In the overview table, we currently show results for each iaas region separately, such as syseleven ham1 and dus2. Similarly for k8s branches.

We can aggregate these results per CSP (e.g., using some jinja2 filter or something): namely, compute the intersection of the sets of passed versions. But we also need to find a way to combine the details pages! Maybe we need some kind of tagging mechanism for the subjects, or we need to work with prefixes?

PR #949, opened 2025-06-04
combine compliance results for regions/k8s branches

mbuechse:
resolving #943


mbuechse:
Screenshot from 2025-06-04 12-03-27
Now we show only one badge, but the problem is that we still have to aggregate the details pages. This will be a bit more involved!

mbuechse:
Now we can combine the result details of multiple regions in one page:
Screenshot from 2025-06-04 16-39-23

Compare to single region (this works the same way as before this PR):
Screenshot from 2025-06-04 16-39-36

mbuechse:
Added links to individual subjects to the top:

Screenshot from 2025-06-04 19-19-27

fkr (approved):

PR #954, opened 2025-06-12
Involve Forum SCS-Standards for documents in the Global track

mbuechse:
Without this involvement, the Forum SCS-Standards would need a separate means of keeping their documents, which would make things confusing.


Standards/scs-0001-v1-sovereign-cloud-standards.md
```diff
@@ -184,7 +184,10 @@ where the group which has to form the consensus depends on the `track` of the do
 - KaaS: The team working on Kubernetes-as-a-service topics
 - Ops: The team working on operations topics
 - IAM: The team working on identity and access management topics
-- Global: The entire SCS community
+- Global: The entire SCS community as well as the Forum SCS-Standards
```
fkr, line 187:
```suggestion
- Global: The SCS community, as defined in SCS-0005, as well as the Forum SCS-Standards
```

PR #956, opened 2025-06-20
Bugfix: scs-0116-v1 is recommended and not mandatory

martinmo:
Thanks @josephineSei for pointing this out.


martinmo:
https://docs.scs.community/standards/scs-0116-v1-key-manager-standard#key-manager-standard

mbuechse:
The key manager API is recommended, and therefore not offering it will not lead to a failing of the test (but a warning).

PR #958, opened 2025-06-27
Suggested wording that the SIG Std/Cert can fill in for teams.

garloff:
We discussed and agreed on this in the SIG Std/Cert on 2025-06-26.


garloff:
In my mental model, the power to create technical standards is delegated from the SCS Community (represented by the Project Board) to the teams and when we decide that the SIG can fill in for inactive teams, the Project Board should probably agree and delegate to the SIG instead for that case. So I guess the Project Board should look at this quickly before we can merge it.
(Likewise, the power to create procedural standards and to stear the overall process originates at the Forum SCS-Standards and largely delegated (with the veto powers merged by #954) to the SIG Std/Cert.)

garloff:
The SCS Project Board is in favor of this as discussed today.

PR #961, opened 2025-07-02
Pin postgres:16 version, as major upgrades need intervention.

garloff:
We should be able to regularly do a docker image pull to get the latest versions. postgres[:latest] however would be dangerous, as postgres typically requires a few manual steps when upgrading between major versions. So let's pin the major version 16 for now and change this to 17 (or something even higher_ once we have done the manual migration.


mbuechse (approved):
Good thinking. Thanks!

PR #965, opened 2025-07-10
fix link to CIVS

depressiveRobot:
The old link leads to SSL error: no alternative certificate subject name matches target hostname 'civs.cs.cornell.edu'


fkr (approved):
thanks. good catch.

Issue #899, opened 2025-04-08
[Feature Request] Notification mechanism for compliance monitor


mbuechse:
See [minutes of SIG Std/Cert 2025-03-06](https://github.com/SovereignCloudStack/minutes/blob/main/sig-standardization/20250306.md#notification-mechanism):

- contact address for each test subject needed
- send mail to some appointee from Forum SCS-Standards as well as contact of test subject

garloff:
Makes sense to me.

mbuechse:
Decision by [SIG Std/Cert](https://github.com/SovereignCloudStack/minutes/blob/main/sig-standardization/20250508.md#notification-mechanism-for-compliance-monitor): store contact info for each test subject securely using a secret (Zuul or Vault).

PR #970, opened 2025-07-24
Add os_purpose field.

garloff:
Also add a link to OpenStack specs for os_distro and explain why we need it.

This is patch 1 from several. It adds this os_purpose field as *recommended*. As it does not break certification, this can be done in the existing v1 of the scs-0102 standard. I suggest we discuss this one first until we are aligned. In parallel, I'd like to prepare a blog article that serves as user guide how to select images in a portable way without relying on the name. We can reference it in tne implementation notes. We should also enhance the image-manager input to carry this field soon.

In a future patch, we create a v2 of scs-0102 which *requires* the os_purpose field. The naming in scs-0102 is already a recommendation only, so that can remain. scs-0104 will need a new version then, as it does use the recommended names as mandatory for the mandatory images.

### DIFF
--- a/Standards/scs-0102-v1-image-metadata.md
+++ b/Standards/scs-0102-v1-image-metadata.md
@@ -65,6 +65,39 @@ The following property is recommended:
 
 - `hypervisor_type`
 
+The values for `architecture` and `os_distro` and `hypervisor_type` (the latter only if specified) values
+must follow the [OpenStack specifications](https://docs.openstack.org/glance/2025.1/admin/useful-image-properties.html).
+The `os_version` string should be numeric if the distribution uses numbers, the pair `os_distro` `os_version` should
+for example be `ubuntu` `24.04` for Ubuntu Noble Numbat 24.04[.x] LTS.
+
+To allow the distinction between general purpose images (which should come from upstream with at most some
+targeted adjustments as required by the cloud such as e.g. drviers) and images that are purpose-built, we
+recommend an additional field:
+
+- `os_purpose`
+
+The following values are allowed
+
+| `os_purpose` value | Intention                                                 |
+|--------------------|-----------------------------------------------------------|
+| `generic`          | A general purpose image, (mostly) vanilla from upstream   |
+| `capinode`         | Node image built for cluster-API                          |
+| `gpu`              | Image with GPU drivers e.g. for HPC or AI                 |
+| `network`          | Image for a network appliance (router, loadbalancer, ...) |
+| `appliance`        | Other appliances (single-purpose solutions)               |
+| `custom`           | None of the above                                         |
+
+Note that no other values are currently allowed and `custom` (or `appliance`) should be used in case
+of doubt. Talk to the SCS standardization bodies if you'd like to see this list extended which is
+likely the case if you fall back to `custom`.
+
+The usage of standardized `os_distro`, `os_version` and `os_purpose` help cloud users to create
+automation that works across clouds without requiring image names to be standardized. Only
+one visible public image with `os_purpose` = `generic` and the same settings for `os_distro`
+and `os_version` should be active on any given cloud.
+The link to the OpenStack specs and the recommended `os_purpose` was added in 7/2025 to this
+standard; a new version of the standard that requires `os_purpose` will be created later.
+
 The following further properties are recommended (if the features are supported):
 
 - `hw_rng_model`
@@ -72,7 +105,8 @@ The following further properties are recommended (if the features are supported)
 - `hw_watchdog_action`, `hw_mem_encryption`, `hw_pmu`, `hw_video_ram`, `hw_vif_multiqueue_enabled`
 
 The `trait:XXX=required` property can be used to indicate that certain virtual hardware
-features `XXX` are required.
+features `XXX` are required which may be advertised in matching
+[flavor extra specs](https://docs.openstack.org/nova/latest/user/flavors.html#extra-specs).
 
 ## Image handling
 


Standards/scs-0102-v1-image-metadata.md
```diff
@@ -65,14 +66,47 @@ The following property is recommended:
...
+Note that no other values are currently allowed and `custom` should be used in case
+of doubt. Talk to the SCS standardization bodies if you'd like to see this list extended which is
+likely the case if you fall back to `custom`.
+
+The usage of standardized `os_distro`, `os_version`, `architecture`, and `os_purpose` help cloud users to create
+automation that works across clouds without requiring image names to be standardized. Only
+one visible public image with `os_purpose` = `generic` and the same settings for `os_distro`,
+`os_version` and `architecture` should be active on any given cloud.
+The link to the OpenStack specs and the recommended `os_purpose` was added to version 1.1 of
+this standard; a new version 2 of the standard that requires `os_purpose` will be created later.
```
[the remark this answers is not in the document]
garloff: Agreed. See 8cabaa5

Standards/scs-0102-v1-image-metadata.md
```diff
@@ -65,14 +66,47 @@ The following property is recommended:
...
+  | `custom`           | None of the above                                         |
+
+  Note that no other values are currently allowed and `custom` should be used in case
+  of doubt. Talk to the SCS standardization bodies if you'd like to see this list extended which is
+  likely the case if you fall back to `custom`.
+
+  The usage of standardized `os_distro`, `os_version`, `architecture`, and `os_purpose` help cloud users to create
+  automation that works across clouds without requiring image names to be standardized. Only
+  one visible public image with `os_purpose` = `generic` and the same settings for `os_distro`,
+  `os_version` and `architecture` should be active on any given cloud.
```
mbuechse, line 97: Comprehension question: what is meant by "be active" in this context? The text already speaks of visible public image, so I would presume it either exists or it doesn't?

Standards/scs-0102-v1-image-metadata.md
```diff
@@ -244,4 +278,13 @@ A boolean property that is not present is considered to be `false`.
 
 The script `image-md-check.py` retrieves the
 image list from a configured cloud and checks each image for the
-completeness and consistency of mandatory properties.
+completeness and consistency of mandatory properties. See the
+[Corresponding Implemetation and Testing Notes](https://docs.scs.community/standards/scs-0102-w1-image-metadata-implementation-testing)
```
mbuechse, line 282: I'm actually wondering whether the whole section (Conformance Tests) should be removed...

mbuechse: (or at least its entire contents replaced by a link to the notes)

Standards/scs-0102-v1-image-metadata.md
garloff: Feel free to word this "should exist" if you find "should be active" confusing. For me it's the same.
(There are several ways to make an image inactive. Setting `os_hidden`='True'`. Marking it as `visibility`=`community`. We don't prescribe any of these. As both basically affect visibility, the idea of deactivation and removing from visibility possibly being different is currently not true -- I would not exclude the possibility of a future mechanism to deactivate in ways that may not be well characterized by "hiding", although I don't expect it. I felt that the wording as chosen is a safe choice. But it has confused you, so it wasn't.)

mbuechse: I suppose when I set `os_hidden=True` then it's no longer visible, and when I set `visibility=community` then it's no longer public, so the word 'active' to me suggested that there would be a third way, but I didn't know what this might be. I'm afraid I'm a reviewer who is quite easily confused ;)

Standards/scs-0102-v1-image-metadata.md
```diff
@@ -65,14 +66,47 @@ The following property is recommended:
...
+  Note that no other values are currently allowed and `custom` should be used in case
+  of doubt. Talk to the SCS standardization bodies if you'd like to see this list extended which is
+  likely the case if you fall back to `custom`.
+
+  The usage of standardized `os_distro`, `os_version`, `architecture`, and `os_purpose` help cloud users to create
+  automation that works across clouds without requiring image names to be standardized. Only
+  one visible public image with `os_purpose` = `generic` and the same settings for `os_distro`,
+  `os_version` and `architecture` should be active on any given cloud.
+  The link to the OpenStack specs and the recommended `os_purpose` was added to version 1.1 of
+  this standard; a new version 2 of the standard that requires `os_purpose` will be created later.
```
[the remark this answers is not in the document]
mbuechse: The context is that this whole part is indented below the bullet point for precisely the field in question, no?

Standards/scs-0102-v1-image-metadata.md
mbuechse, line 97: For instance

```suggestion
  The usage of standardized `os_distro`, `os_version`, `architecture`, and `os_purpose` help cloud users to create
  automation that works across clouds without requiring image names to be standardized.

  _Uniqueness requirement_: whenever there are two images that have `os_hidden=False`, `visibility=public`,
  and that coincide in all three fields `os_distro`, `os_version`, and `architecture`, then only one of them may
  have `os_purpose=generic`. In other words, users who are search visible public images for a generic OS
  of a certain distro, version, and architecture will not get more than one result.
```

mbuechse:
@garloff Fine, we can add tests later. Have had a chance to look at the last remaining conversation?

garloff:
Help me: Anything open?

garloff:
Thanks for the pointer! 
You were right:
- The history here was redundant
- At this place, the context was clear
I merged your suggested change, thanks!

Standards/scs-0102-v1-image-metadata.md
```diff
@@ -65,14 +66,50 @@ The following property is recommended:
...
+  likely the case if you fall back to `custom`.
+
+  The usage of standardized `os_distro`, `os_version`, `architecture`, and `os_purpose` help cloud users to create
+  automation that works across clouds without requiring image names to be standardized.
+
+  _Uniqueness requirement_: whenever there are two images that have `os_hidden=False`, `visibility=public`,
+  and that coincide in all three fields `os_distro`, `os_version`, and `architecture`, then only one of them may
+  have `os_purpose=generic`. In other words, users who search visible public images for a generic OS
+  of a certain distro, version, and architecture will not get more than one result.
+  Note: Expect this field to become mandatory in v2 of this standard sooner rather than later.
```
mbuechse, line 102:
```suggestion

  Note: Expect this field to become mandatory in v2 of this standard sooner rather than later.
```

mbuechse (approved):
Great!

Minor suggestion: add another line break.

garloff:
OK, so we have a test now.
Required a bit more changes than I would have liked ... thus I tested quite a bit.
As collateral damage, we got rid of a warning (cirros 32MB) and report missing mandatory images correctly with ERROR prefix now and have a few more comments in the code.
A re-review by the members of the higher league of Python wizardry may seem appropriate.

mbuechse:
@garloff The main problem is that we have concurrent change of the file in #899 as I expressly told everyone in the SIG meetings, so now I have to sort that out, and I guess I will, but please don't merge yet.

mbuechse:
Converted back to draft to be extra sure that no accidental merge happens. I have to sort out the potential conflicts with #899 first.

garloff:
Oh well, I thought that the refactoring has landed meanwhile.
Apparently not.
Well, the changes to the test can be redone, so we could drop them from #970 as intended before.

mbuechse:
In all likelihood, the changes will be landing today, and then, I can rebase this.

PR #975, opened 2025-08-15
record aborted testcases, part 1

mbuechse:
closing #889

### DIFF
--- a/Tests/iaas/flavor-naming/flavor-names-openstack.py
+++ b/Tests/iaas/flavor-naming/flavor-names-openstack.py
@@ -16,13 +16,17 @@ provided by openstack, such as the number of vCPUs and memory.
 SPDX-License-Identifier: CC-BY-SA 4.0
 """
 
+import logging
 import os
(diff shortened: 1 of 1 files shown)


mbuechse:
@depressiveRobot the markdown-link-check fails, but this is because the link in question refers to the main branch; ironically, merging this branch will immediately repair this problem. Also, be advised that merging is entirely possible (the check is not a blocker).

depressiveRobot (commented):
Thanks for doing that. I welcome the approach of catching all exceptions outside the actual test method.

However, the origin for this change still seems to be not fixed: if a test didn't finish cause of an unhandled exception (now `ABORT`), it is still not linked in the compliance monitor:

Screenshot 2025-09-05 at 18 06 04

Here is the report with one aborted check which I used to test this: 
[scs2-report.yaml](https://github.com/user-attachments/files/22175730/scs2-report.yaml)

depressiveRobot (approved):

PR #980, opened 2025-09-03
stabilize w1 as the main is stable

fkr:


mbuechse:
I oppose stabilizing non-authoritative documents. Supplements are meant to be changed more easily than standards.

[scs-0001-v1](https://docs.scs.community/standards/scs-0001-v1-sovereign-cloud-standards#process):

> Supplements may be kept in Draft state, because they are not authoritative.

mbuechse:
But also, AFAICT the main scs-0124 is NOT stable, contrary to your statement.

fkr:
@mbuechse Should this sentence in 0001 "Supplements may be kept in Draft state, because they are not authoritative." be adjusted to a "shall be kept" ?

PR #981, opened 2025-09-03
Remove Status field from Supplement docs

mbuechse:
This field causes some confusion for Supplement docs, for the standard scs-0001-v1 merely states that these docs may be kept in Draft state, which provides no certainty as to what should be done. Here, I propose to go for maximum certainty by simply stating they are kept in Draft state in perpetuity, and therefore, the field is not even present in them.


mbuechse:
Note that the script `chk_adrs.py` is not changed here because that doesn't even check Supplement docs. (A separate matter for a separate PR, someday)

fkr:
wonderful. thanks!

PR #988, opened 2025-09-16
Add Standards/scs-0102-v2-image-metadata.md

mbuechse:


depressiveRobot (changes requested):
I think the `replaces` metadata field to the old document needs to be added.

depressiveRobot (approved):
Thanks for adding the missing field. Looks good now.

fkr:
> note to self: add testcase scs-0102-os_purpose-uniqueness

for clarification: are you going to add the testcase or is this something where someone else should deliver one?

PR #992, opened 2025-09-17
Add general sanity check to OpenStack tests

### LINKED ISSUE #991: Add general sanity check to openstack tests
Assume you have a typo in your OS_CLOUD.
Currently, each and every testcase will result in ABORT with very long and marginally helpful error messages.

Assume you run your tests with a privileged user.
Currently, this will only trip up the key-manager-check (because it checks that a normal user can do something that any admin can do trivially).

I propose to add a general sanity check at the beginning `openstack_test.py` that makes sure your OS_CLOUD works and the user doesn't have elevated privileges, as a safety measure, to not single out the poor key-manager-check. The actual testcases shouldn't even be attempted if the sanity check is failed.

RFC @fkr @garloff @depressiveRobot @toothstone

mbuechse:
closes #991

### DIFF
--- a/Tests/iaas/openstack_test.py
+++ b/Tests/iaas/openstack_test.py
@@ -291,7 +291,7 @@ def run_sanity_checks(container):
         raise
     if "member" not in ensure_unprivileged(conn, quiet=True):
         logger.critical("Please make sure that your OpenStack user has role member.")
-        raise RuntimeError("OpenStack user has elevated privileges.")
+        raise RuntimeError("OpenStack user is missing member role.")
 
 
 def main(argv):


Tests/iaas/openstack_test.py
```diff
@@ -280,6 +280,20 @@ def harness(name, *check_fns):
...
+    # the former would lead to each testcase aborting with a marginally useful message;
+    # the latter would lead to scs_0116_permissions aborting, which we don't want to single out
+    try:
+        conn = container.conn
+    except openstack.exceptions.ConfigException:
+        logger.critical("Please make sure that ~/.config/openstack/clouds.yaml exists and is correct!")
+        raise
+    if "member" not in ensure_unprivileged(conn, quiet=True):
+        logger.critical("Please make sure that your OpenStack user has role member.")
+        raise RuntimeError("OpenStack user has elevated privileges.")
```
toothstone, line 294: Shouldn't we raise something like "user is missing member role"?
"user privileges to high" is already covered in `ensure_unprivileged` line 21.

mbuechse: D'oh. You're right.
See 237541e

Issue #993, opened 2025-09-19
Hand compliance monitor ops over to Forum


mbuechse:
The compliance monitor runs on a VM within SCS2. This VM has been created by @garloff whrn he transferred my original deployment from the old environment gx-scs over to SCS2. According to Horizon, the associated keypair is `garloff`. For some reason, I can still access the VM (probably because @garloff transferred the volume).

At any rate, I think the operations should be handed over to the Forum, that is. @depressiveRobot. We have to determine who may access the VM and how this is implemented. I'm still puzzled that only one keypair can be configured in Horizion?

mbuechse:
I have added the keys from

- https://meta.sr.ht/~fkr.keys
- https://meta.sr.ht/~mbuechse.keys
- https://github.com/depressiveRobot.keys

PR #994, opened 2025-09-19
Add CNDS region RO to compliance monitor

mbuechse:


mbuechse:
When this is merged, we need to do

```shell
$ cd standards
$ git checkout main && git pull -p
$ cd compliance-monitor
$ docker compose build && docker compose down web && docker compose up web -d
```

on the VM of the compliance monitor. See #993.

PR #938, opened 2025-05-25
add larger memory flavors

scoopex:

PR #996, opened 2025-10-06
Add comment on `--limit-memory` flavor-manager parameter.

garloff:
This enhances #938 by a comment on the flavor-manager flag that limits recommended flavors by a maximum memory setting.


mbuechse (approved):
Thanks!

PR #997, opened 2025-10-06
Feat/limit new flavor recommendations

garloff:
As discussed in the SIG Std/Cert on 2025-10-02, a proposal to simplify the new recommended larger flavors:
- Drop the ones with disks (but add a comment how we would expect disk sizes)
- Drop flavors with 1:16 and 1:32 vCPU:GiB RAM ratios, as these are unusual (but add a comment how to systematically create flavors with more vCPUs).

Overall, trying to avoid recommendations that many providers would consider niche and might cause them to discount recommendations. Let's rather keep it short and describe once more how the list would be extended in a systematic way. This will reduce divergence and thus any pain in case we latter add larger recommended and mandatory flavors.


garloff:
Against wrong branch, fixed in #998.

PR #998, opened 2025-10-06
Remove 1:16, 1:32 and -disk flavors from recommended list.

garloff:
Instead create language that explains the way how to systematically extend the list to stay in line with what might later be standardized.

As discussed in the SIG Std/Cert on 2025-10-02, a proposal to simplify the new recommended larger flavors:

    Drop the ones with disks (but add a comment how we would expect disk sizes)
    Drop flavors with 1:16 and 1:32 vCPU:GiB RAM ratios, as these are unusual (but add a comment how to systematically create flavors with more vCPUs).

Overall, trying to avoid recommendations that many providers would consider niche and might cause them to discount recommendations. Let's rather keep it short and describe once more how the list would be extended in a systematic way. This will reduce divergence and thus any pain in case we latter add larger recommended and mandatory flavors.

This is intended to change PR #938.


mbuechse (approved):
I support reducing the number of recommended flavors. I want to make it clear that the long list is not due to me; rather it goes back to the original proposal. I'm wondering though whether @scoopex wants to chime in here!

PR #999, opened 2025-10-06
Feat/escape underscore

garloff:
I was under the impression that we need to escape underscores in markdown unless we have `literals` (marked by backticks). Both docusaurus and github render *single* underscores correctly, so maybe we don't need to care? Or maybe my memory failed on me ...


garloff:
Against wrong branch (main) again, fixed in #1000.

PR #1000, opened 2025-10-06
Escape underscore for markdown rendering.

garloff:
github seems to be able to do without, but maybe not all MD renderers are so tolerant.


mbuechse (approved):
It's a pity that we have to use the term `extra_specs` (with underscore, with or without typewriter font) in prose at all. Maybe we should just consistently use the spelling "extra specs"? Otherwise, this is fine by me.

PR #1002, opened 2025-10-26
Feat/curr cpu and gpu 202510

garloff:
As new generations of CPUs and GPUs get released and we have an optional ability for cloud operators to encode such details in the flavors names, we have to updates the available CPUs and GPUs. Do just that.
Two more details:
- I also addded Cloud Hypervisor (chy), as it may have relevance for higher security setups. (The Cyberus Hypervisor would also use this type ... as it's a variant)
- I fixed a confusion with RDNA and CDNA which I assumed to be more similar than they are. So RDNA gets a .1 number assigned now. As this is for AMD (which is not very popular as server GPU) and RDNA are consumer cards which are not supposed to be used in cloud servers, this should not affect anyone ... so we can do this change, avoiding future trouble.


Standards/scs-0100-w1-flavor-naming-implementation-testing.md
```diff
@@ -135,6 +135,12 @@ Cores and 64 Stream Processors per CU.
 | Inst MI300X | 1216     | 19456       | 304 | 192G HBM3  | `GA3-304-192h` |
 | Inst MI325X | 1216     | 19456       | 304 | 288G HBM3  | `GA3-304-288h` |
 
+Note that we previously assumed more similarity of consumer RDNA-x with
+server CDNA-x that actually is the case; the RDNA-x cards now use `x.1`
+(since v3.3 as of Oct 2025) to be able to differentiate them. We will
+tolerate potential rate cases of old installations calling RDNA-x as
```
mbuechse:
```suggestion
tolerate potential rare cases of old installations calling RDNA-x as
```

Standards/scs-0100-v3-flavor-naming.md
```diff
@@ -374,7 +375,7 @@ conjunction with a vendor. At present, these values are possible:
 | 3          | Ice Lake          | Zen-3 (Milan)  | A71x/NeoN2/V2(ARMv9) | TBD        |
 | 4          | Sapphire Rapids   | Zen-4 (Genoa)  | AmpereOne (ARMv8.6)  | TBD        |
 | 5          | Sierra Forest(E)  | Zen-5 (Turin)  | A72x/NeoN3/V3(Av9.2) | TBD        |
-| 6          | Granite Rapids(P) |                |                      | TBD        |
+| 6          | Granite Rapids(P) | Zen-6 (Morpheus) | OryonV3 (ARMv9.2)  | TBD        |
```
garloff:
```suggestion
| 6          | Granite Rapids(P) | Zen-6 (Venice)   |                            | TBD       |
```

mbuechse (approved):
I'm not too excited about the breaking change that forces us to "tolerate [the old scheme] for the time being". However, I understand this is about consumer-level chips that no one would use in their cloud installation anyway. I will have to trust your judgment about that.

garloff:
Yeah, compromises happens when you have to clean up old oversights.
But yes, I think this will not affect anyone.

PR #1008, opened 2025-10-30
Bugfix: Heisenbug when parsing console output in entropy check

### LINKED ISSUE #1007: [BUG] Heisenbug parsing console output in entropy check
Most of the time, the console output looks like this:

```text
[   20.645706] cloud-init[539]: _scs-test-entropy-avail
[   20.647956] cloud-init[539]: 256
[   20.650400] cloud-init[539]: _scs-test-fips-test
[   20.651530] cloud-init[539]: rngtest 5
```

This is generated with the following cloud-init user-data:

```yaml
runcmd:
  - echo '_scs-test-entropy-avail'; cat /proc/sys/kernel/random/entropy_avail
  - echo '_scs-test-fips-test'; cat /dev/random | rngtest -c 1000
```

However, often enough, the console output looks a little different:

```text
[   17.344168] cloud-init[520]: Processing triggers for man-db (2.11.2-2) ...

Debian GNU/Linux 12 scs-0101-server ttyS0

scs-0101-server login: [   18.627185] cloud-init[520]: _scs-test-entropy-avail
[   18.628961] cloud-init[520]: 256
```

You see, because of this login prompt, the line containing `_scs-test-entropy-avail` now has some nasty superfluous prefix. And this throws a veritable wrench in the parsing scheme, which looks for the keyword `_scs-test-entropy-avail` in order to learn how much of the line to chop off (because we want to get rid of `[   18.627185] cloud-init[520]: `).

So, thanks to the superfluous login prompt, the parser will chop off too much of the next line, including the number `256` that we are looking for.

mbuechse:
Resolves #1007.

The login prompt generated by the serial getty service could mess up the output and throw off our primitive parser.

Thanks @garloff for the crucial tip!

### DIFF
--- a/Tests/iaas/scs_0101_entropy/entropy_check.py
+++ b/Tests/iaas/scs_0101_entropy/entropy_check.py
@@ -32,11 +32,14 @@ FLAVOR_OPTIONAL = ("hw_rng:rate_bytes", "hw_rng:rate_period")
 
 TIMEOUT = 5 * 60  # timeout in seconds after which we no longer wait for the VM to complete the run
 MARKER = '_scs-test-'
+# NOTE we mask serial-getty@ttyS0.service because the login prompt messes up the console output that we want to parse
 SERVER_USERDATA_GENERIC = """
 #cloud-config
 # apt-placeholder
 packages:
   - rng-tools5
+bootcmd:
+  - systemctl mask serial-getty@ttyS0.service
 runcmd:
   - echo '_scs-test-entropy-avail'; cat /proc/sys/kernel/random/entropy_avail
   - echo '_scs-test-fips-test'; cat /dev/random | rngtest -c 1000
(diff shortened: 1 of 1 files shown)


mbuechse:
I have tested this code. I added debug output for the whole console output, and the login prompt no longer appears in multiple attempts. I am pretty confident that the problem is gone or at least greatly reduced...

fkr (approved):

garloff (approved):
I cannot judge for sure whether this provides a 100% guarantee to solve the issue or there is a possible race left where systemd has already invoked the getty process that is in the scheduling queue and that outputs the login prompt still intermingled with the debug output. Maybe it's impossible, maybe it's not. In the latter case, it's for sure very unlikely, and as our lives do not depend on it, I'd judge this good enough. Time will tell if we still have <0.1% fails left or not. If so, we can deal with it then. For sure, this is a huge improvement.

PR #1023, opened 2025-11-27
Avoid CPU:Mem which confuses some markdown parsers.

garloff:
Cosmetic ...


garloff:
> I would apply the same change to https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0103-v1-standard-flavors.md?plain=1#L135 I can do it (but I don't know when).

True, thanks for spotting! Done in @b105c8d.

PR #1029, opened 2025-12-01
Add SCS-0005-v2 project governance standard

berendt:
Add version 2 of the project governance standard. Compared to v1, this version adds the "Conflict of Interest and Organizational Diversity" section which ensures:

- No more than one board seat may be held by individuals from the same organization
- Rules for handling multiple top-ranked candidates from the same organization in elections
- Process for resolving conflicts when board members change their organizational affiliation during their term

### DIFF
--- a/Standards/scs-0005-v2-project-governance.md
+++ b/Standards/scs-0005-v2-project-governance.md
@@ -96,8 +96,8 @@ If two or more candidates from the same organization are among the top-ranked ch
 
 If an existing board member changes their organizational affiliation during their term, resulting in multiple board
 members from the same organization, one of the affected members must resign their seat. The affected members may mutually
-agree on who will resign. If no agreement can be reached within 30 days of the affiliation change, the SCS Project Board
-will decide by majority vote which member shall retain their seat. A replacement will be determined according to the most
+agree on who will resign. If no agreement can be reached within 30 days of the affiliation change, the member who received
+fewer votes in the most recent election will be required to resign. A replacement will be determined according to the most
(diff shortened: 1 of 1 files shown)


Standards/scs-0005-v2-project-governance.md
```diff
@@ -0,0 +1,162 @@
...
+
+If an existing board member changes their organizational affiliation during their term, resulting in multiple board
+members from the same organization, one of the affected members must resign their seat. The affected members may mutually
+agree on who will resign. If no agreement can be reached within 30 days of the affiliation change, the SCS Project Board
+will decide by majority vote which member shall retain their seat. A replacement will be determined according to the most
+recent election results, selecting the next eligible candidate from a different organization.
+
+#### Resignation
+
+Resignation can happen for several reasons:
```
mbuechse:
```suggestion
Resignation can happen for several reasons, such as:
```

mbuechse:
I noticed two points:

- If a change of affiliation happens and the affected members can not agree on who is to stay and who is to go, it seems like a major crisis to me, and I'm not sure if the board still has the necessary trust to proceed;
- According to the text, if one spot is vacated and cannot be filled, that seems fine, but if it happens a second time, an extraordinary election is triggered. I think it should depend more on the size of the board, because it can be fewer people to begin with, depending on the number of nominees.

mbuechse (approved):
Thanks Christian, I like the changes. They address my original concerns. I approve this, but my confidence here is rather low, so I hope others will approve as well.

jklare (approved):
LGTM

PR #1034, opened 2025-12-03
Include replaces field into chk_adrs.py

mbuechse:


depressiveRobot (commented):
Thank you for adding the check. Tested it, works for me too :)

PR #1052, opened 2026-01-20
Fix: stale ec2 credentials would be used

mbuechse:


mbuechse:
@garloff It's what I had time for today.

PR #1053, opened 2026-01-20
Migrate SCS-compatible IaaS v3 through v5.1 to fine-grained testcases

mbuechse:
This is possible because since the introduction of 'next', the fine-grained testcases have been reported -- even for the older versions.


garloff (approved):
1. Looks good
2. Tested successfully (against one cloud) for v3, v4, v5.1

PR #1054, opened 2026-01-21
Remove misleading reference to recommendation

mbuechse:
Whether a testcase is recommended or required depends on the context.


depressiveRobot (approved):
Looks good to me. Thanks for the quick fix.

However, to ensure that outsiders can understand the change (discussion took place on Matrix), we should state the reason for it too.
As I understand it, the adjustment is done to avoid confusion about the fact that properties can be recommended or required, depending on the version of the standard and/or the certification scope.

Issue #1055, opened 2026-01-21
[Feature Request] Restructure details page


mbuechse:
Currently, the page lists, one scope version after the other, all test cases, each one with result and description.

From my POV, this view has multiple clear issues:

- testcases are listed multiple times (description is repeated)
- most important info is not at the top (failed testcases)
- the page is just very long

My proposal:

1. Show a list of testcases, sorted/grouped by result
2. For each testcase, show the list of version/target pairs that refer to it

For instance:

| testcase | v3 | v4 | v5.1 | next | description |
| --- | --- | --- | --- | --- | --- |
| **FAILED:** | | | | | |
| scs-0102-prop-min_disk | main | main | main | main | Each image has a meaningful value for `min_disk`. |
| scs-0102-prop-os_purpose | | | | recommended | Each image has a meaningful value for `os_purpose`. |
| **MISSING:** | | | | | |
| ... | | | | | |

I guess we should also rename target `main` to `required`, to make it more accessible.

PR #1074, opened 2026-01-29
Make testcase docs more consistent, introduce testcase url

mbuechse:


depressiveRobot (approved):
Nice work! Thank you for implementing the idea of deep linking the corresponding standard for each test (see #1055) :)

Issue #1083, opened 2026-02-02
[BUG] flavor form field not accepting 3.5 GiB


mbuechse:
Steps to reproduce:

- open https://flavors.scs.community/cgi-bin/flavor-form.py?flavor=SCS-2V-3.5
- hit "generate"
- see error message (nearest possible values being 3 and 4)

Secondly, the form needs to be deployed again because it doesn't yet offer the "cloud hypervisor". See

- https://flavors.scs.community/cgi-bin/flavor-form.py?flavor=SCS-16V-48_chy

(But this is really just a matter of deploying the latest version.)

PR #1084, opened 2026-02-02
Pass step= to form to allow half GiB (if below 10).

garloff:
This addresses #1083.
A bit hacky, unfortunately, but workable.


mbuechse:
Something's rotten in the state of Github. The checks don't run, and I can't find this branch in my local copy even after a seemingly successful `git pull`.

mbuechse:
I think my local copy had some problems, and I managed to fix them. I tested the change, and it seemed fine.

mbuechse (approved):

PR #1094, opened 2026-02-17
Update README with instructions for sonobuoy

gerbsen:
Reword instructions for installing test dependencies and add note about Sonobuoy installation.


mbuechse (approved):
Important point! Thanks!

PR #1100, opened 2026-02-17
Install Sonobuoy in Dockerfile

gerbsen:
Add Sonobuoy installation to Dockerfile with architecture support.

### DIFF
--- a/Tests/Dockerfile
+++ b/Tests/Dockerfile
@@ -4,6 +4,15 @@ FROM python:3.10
 
 WORKDIR /scs-compliance
 
+ARG SONOBUOY_VERSION=0.57.3
+ARG TARGETOS
+ARG TARGETARCH
+
+RUN set -eux; \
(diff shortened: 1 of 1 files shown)


mbuechse (commented):
To be frank, I do hate this Dockerfile. It tries to install all dependencies, that is, for IaaS and KaaS alike, regardless of the eventual purpose. Would it be possible to supply sonobuoy via a mount? Otherwise, maybe the code can at least be shortened. With Ansible, we only need these lines (with curl):

```ansible
    - name: Install Sonobuoy
      ansible.builtin.shell:
        curl -L {{ sonobuoy_tar_gz_url | trim | quote }} | tar xz sonobuoy
      args:
        chdir: "{{ ansible_user_dir}}/.local/bin/"
        creates: sonobuoy
      when: kaas | bool
```

Apparently, when curl can be used, there is no need to have three separate commands for downloading, extracting and removing a temporary file, and even the chmod seems to be unnecessary? Not sure if curl is available on a python-3.10 image (or whatever is used here), but I think it's worth a try.

mbuechse (approved):
Approving under the assumption that it's been tested. My opinion about Docker hasn't changed, but I see myself outvoted, and that is fine.

PR #1105, opened 2026-02-18
improve logging for k8s_node_distribution_check

depressiveRobot:
Slightly improves logging to avoid confusion/be more explicit.


Tests/kaas/k8s-node-distribution/k8s_node_distribution_check.py
```diff
@@ -165,7 +165,8 @@ def compare_labels(node_list, node_type="control"):
     #
     # if node_type == "control":
     #     raise DistributionException("The distribution of nodes described in the standard couldn't be detected.")
-    logger.warning("No node distribution could be detected for the worker nodes. "
+    if node_type == "worker":
```
mbuechse: This change introduces two mistakes:

a) this warning applies to all kinds of nodes (the text in the warning is wrong)
b) you messed up the indentation (see flake8)

mbuechse (approved):

Issue #1004, opened 2025-10-28
[BUG] SCS-IaaS mandatory "s3" service type is ambiguous, untested and not aligned with the OS Service Type Authority


toothstone:
As discussed on [2025-09-18](https://github.com/SovereignCloudStack/minutes/blob/main/sig-standardization/20250918.md) and [on Matrix](https://matrix.to/#/!xlFjWCAoQenuOeIVQw:matrix.org/$m0sYKivZZtoaA6M5302Phmv8Rk9Qu7Tz3PllTs-8-_I?via=matrix.org&via=regio.chat&via=matrix.uhurutec.com), https://docs.scs.community/standards/scs-0123-v1-mandatory-and-supported-IaaS-services implies a Service type of "s3", but the test suite looks for "object-store", which is actually out of scope since the OpenStack API [listed by the Service Type Authority for that type](https://specs.openstack.org/openstack/service-types-authority/#service-data) is the Swift API.

The next version of SCS-IaaS should define the mandatory catalog in a more precise manner (e.g. in the format OS users are used to from `openstack catalog list`, since this is crucial for the (machine-to-machine) discovery of services.
Furthermore I suggest that whatever SCS defines should adhere with the guidelines of the OpenStack Service Type Authority.

garloff:
The service catalogue entry from OVH looks reasonable to me.
The standard could say something like:

An S3-compatible object-storage service MUST be available. The S3 endpoint MUST be discoverable through the OpenStack service catalogue. If the S3 service does *not* support the OpenStack Swift protocol, it MUST use the Name "s3" and Type "object-store-s3". Otherwise the Swift endpoint (with Name "swift" and Type "object-store") MAY be used to also imply the S3 endpoint and a separate "s3","object-store-s3" is not required. If only the Swift endpoint is advertised, it is MANDATORY that the objects and buckets/containers visible through both protocols are identical.

(The last sentence may go beyond what we had standardized in the past, so we could make this RECOMMENDED.)

berendt:
We deploy the S3 outside of OpenStack primarily as a backup target for Kubernetes on Pure or NetApp storage backends or on-top via . The S3 is not integrated with Keystone and is therefore not listed in the OpenStack catalog. We handle discoverability in the layer above it, since we use OpenStack solely as a foundation for Gardener or Managed VMs, and customers do not have direct access to the OpenStack API there.

fkr:
[...]

PR #1129, opened 2026-03-04
make test case for SCS-0123 "object-store" no longer mandatory

### LINKED ISSUE #1110: [BUG] Test for SCS-0123 is not correctly implemented
First: In the documented standard Swift is optional. But if not present, the test fails.

Second: If Swift is available beside S3 the test expects the same data backend for both APIs:

```
service-apis-check: PASS
DEBUG: ... conn
DEBUG: ** service-apis-check
DEBUG: ... service_apis_check
DEBUG: ... scs_0123_service_compute
DEBUG: ... services_lookup
DEBUG: ... scs_0123_service_identity
DEBUG: ... scs_0123_service_image
DEBUG: ... scs_0123_service_network
DEBUG: ... scs_0123_service_load_balancer
DEBUG: ... scs_0123_service_placement
DEBUG: ... scs_0123_service_object_store
DEBUG: ... scs_0123_storage_apis
DEBUG: ... scs_0123_swift_s3
ERROR: S3 buckets and Swift containers differ:
S3: ['scs-test-container']
SW: []
```

This is also not required regarding the standard documentation.

depressiveRobot:
Since several providers have complained that the test case for SCS-0123 "object-store" does not meet the standard (tested for being mandatory). ~~This will set it to recommended according to the standard.~~ This PR will make it no longer mandatory by removing it completely (as it is not even recommended).

"object-store" was mandatory before, as the test case for "S3" relies on “object-store” service API. As "S3" is currently skipped if no "object-store" is available, we can do this change to be in line with the standard. However, this should not prevent anyone from offering S3. This problem is already tackled in #1004 and #1093.

Fixes #1110

### DIFF
--- a/Tests/iaas/openstack_test.py
+++ b/Tests/iaas/openstack_test.py
@@ -149,7 +149,6 @@ def make_container(cloud):
     c.add_function('scs_0123_service_network', lambda c: compute_scs_0123_service_presence(c.services_lookup, 'network'))
     c.add_function('scs_0123_service_load_balancer', lambda c: compute_scs_0123_service_presence(c.services_lookup, 'load-balancer'))
     c.add_function('scs_0123_service_placement', lambda c: compute_scs_0123_service_presence(c.services_lookup, 'placement'))
-    c.add_function('scs_0123_service_object_store', lambda c: compute_scs_0123_service_presence(c.services_lookup, 'object-store'))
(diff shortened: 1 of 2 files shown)


fkr:
Likely I'm misunderstanding something: if object store is only in the supported list, why do we even test for it (and not just remove it from the test cases list). I understand that moving it to 'recommended' is not ideal, since it is not recommended (hence the question: why not remove it).

The hurdle that the assessment body in the certification has is that the companies running these checks complain to us that we test something that is not mandatory and it throws an error.

fkr:
Thanks @mbuechse for confirming.

PR #1141, opened 2026-04-01
Fix handling of self-signed certs in s3_conn

chess-knight:
Take insecure/verify/cacert parameter from clouds.yaml and pass it to boto3.resource.
If insecure is True or verify is False use verify=False also for boto3.
Else, if you provide cacert(via cacert or verify parameter) it is also used for boto3.
If nothing from above is met, use verify=None to keep default boto3 behaviour.

### DIFF
--- a/Tests/iaas/scs_0123_mandatory_services/mandatory_services.py
+++ b/Tests/iaas/scs_0123_mandatory_services/mandatory_services.py
@@ -28,12 +28,13 @@ def compute_scs_0123_service_presence(services_lookup, *names):
 
 def s3_conn(creds, conn):
     """Return an s3 client conn"""
+    insecure = conn.config.config.get("insecure")
+    verify = conn.config.config.get("verify")
     cacert = conn.config.config.get("cacert")
-    # TODO: Handle self-signed certs (from ca_cert in openstack config)
-    if cacert:
-        logger.warning(f"Trust all Certificates in S3, OpenStack uses {cacert}")
+    vrfy = False if (insecure is True or verify is False) else \
+        (cacert or (verify if isinstance(verify, str) else None))
     return boto3.resource(
-        's3', endpoint_url=creds["HOST"], verify=not cacert,
+        's3', endpoint_url=creds["HOST"], verify=vrfy,
         aws_access_key_id=creds["AK"], aws_secret_access_key=creds["SK"],
     )
 
--- a/Tests/iaas/scs_0123_mandatory_services/mandatory_services.py
+++ b/Tests/iaas/scs_0123_mandatory_services/mandatory_services.py
@@ -30,12 +30,15 @@ def s3_conn(creds, conn):
     """Return an s3 client conn"""
     cfg = conn.config.config
     # Take insecure/verify/cacert parameter from clouds.yaml and pass it to boto3.resource.
-    # Deliberately do un-Pythonic `is True` and `is False` here because of type mayhem:
-    # for instance, handle verify=False differently from verify=None (or not set) or verify='some.ca'.
-    if cfg.get("insecure") is True or cfg.get("verify") is False:
+    # If verify is True in clouds.yaml, fall back to cacert or None. In the latter case,
+    # the default boto3 behavior is applied (e.g., REQUESTS_CA_BUNDLE can still be used);
+    # but for that, verify must be `None` instead of boto3's default `True`.
+    # Note: cacert must be used to pass the certificate; don't use verify for that; cf.
+    # https://docs.openstack.org/openstacksdk/latest/user/config/configuration.html#ssl-settings
+    if cfg.get("insecure") or not cfg.get("verify", True):
         verify = False
     else:
-        verify = cfg.get("cacert") or cfg.get("verify")
+        verify = cfg.get("cacert")
     return boto3.resource(
         's3', endpoint_url=creds["HOST"], verify=verify,
         aws_access_key_id=creds["AK"], aws_secret_access_key=creds["SK"],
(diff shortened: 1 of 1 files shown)


mbuechse (approved):
Sorry for the long wait. I had to simplify the code a little because the double ternary, plus the unpythonic `is True` freaked me out a bit.

Tests/iaas/scs_0123_mandatory_services/mandatory_services.py
```diff
@@ -28,12 +28,16 @@ def compute_scs_0123_service_presence(services_lookup, *names):
...
-    if cacert:
-        logger.warning(f"Trust all Certificates in S3, OpenStack uses {cacert}")
+    cfg = conn.config.config
+    # Take insecure/verify/cacert parameter from clouds.yaml and pass it to boto3.resource.
+    # Deliberately do un-Pythonic `is True` and `is False` here because of type mayhem:
+    # for instance, handle verify=False differently from verify=None (or not set) or verify='some.ca'.
+    if cfg.get("insecure") is True or cfg.get("verify") is False:
+        verify = False
+    else:
+        verify = cfg.get("cacert") or cfg.get("verify")
```
chess-knight: AFAIK `verify` is by default `True`. Thats why I wrote `verify if isinstance(verify, str) else None` previously.

I think it make more sense to pass `verify=None` to boto3.resource when in openstack clouds.yaml `verify is True`, so default boto3 behavior is applied(e.g. `REQUESTS_CA_BUNDLE` can still be used when `verify is None` in boto3 - if this is not provided, boto3 still defaults to `True`).

mbuechse: Thanks for the explanation! I agree with you. We need to put some more effort into documenting this because this is quite intricate. One question: how important is the case that verify is a string anyway? It seems a bit redundant with cacert, and I think people should be (nudged towards) using that one?

chess-knight: > how important is the case that verify is a string anyway?

You are right, we should avoid that. Instead, the cacert should be used as documented in e.g. https://docs.openstack.org/openstacksdk/latest/user/config/configuration.html#ssl-settings. I consider this only as a "bug" in the openstacksdk. E.g., gophercloud is explicitly setting it is as a *bool*.

So maybe best would be following:
```python
    if cfg.get("insecure") is True or cfg.get("verify") is False:
        verify = False
    else:
        verify = cfg.get("cacert")
```

Tests/iaas/scs_0123_mandatory_services/mandatory_services.py
```diff
@@ -28,12 +28,19 @@ def compute_scs_0123_service_presence(services_lookup, *names):
...
     """Return an s3 client conn"""
-    cacert = conn.config.config.get("cacert")
-    # TODO: Handle self-signed certs (from ca_cert in openstack config)
-    if cacert:
-        logger.warning(f"Trust all Certificates in S3, OpenStack uses {cacert}")
+    cfg = conn.config.config
+    # Take insecure/verify/cacert parameter from clouds.yaml and pass it to boto3.resource.
+    # If verify is True in clouds.yaml, fall back to cacert or None. In the latter case,
+    # the default boto3 behavior is applied (e.g., REQUESTS_CA_BUNDLE can still be used);
+    # but for that, verify must be `None` instead of boto3's default `True`.
```
chess-knight:
```suggestion
    # but for that, verify must be `None`.
```
boto3's default is `None`

mbuechse: Oh, then I must have misunderstood you. Above you wrote

> if this is not provided, boto3 still defaults to True

chess-knight: I meant that if `REQUESTS_CA_BUNDLE` is not provided.

Basically boto3 has default `verify=None` and it does own logic, e.g. checking config file or env variables, if nothing is found, then it sets it to true.

mbuechse: Okay, sorry if was a bit slow. I will adapt the comment!

Tests/iaas/scs_0123_mandatory_services/mandatory_services.py
```diff
@@ -28,12 +28,19 @@ def compute_scs_0123_service_presence(services_lookup, *names):
...
     """Return an s3 client conn"""
-    cacert = conn.config.config.get("cacert")
-    # TODO: Handle self-signed certs (from ca_cert in openstack config)
-    if cacert:
-        logger.warning(f"Trust all Certificates in S3, OpenStack uses {cacert}")
+    cfg = conn.config.config
+    # Take insecure/verify/cacert parameter from clouds.yaml and pass it to boto3.resource.
+    # If verify is True in clouds.yaml, fall back to cacert or None. In the latter case,
+    # the default boto3 behavior is applied (where REQUESTS_CA_BUNDLE can still be used,
+    # and otherwise, boto3 defaults to verify=True).
```
chess-knight:
```suggestion
    # If insecure is False/None and verify is True/None in clouds.yaml, fall back to cacert or None. In the latter case,
    # the default boto3 behavior is applied (where config file or env variables can still be used,
    # and otherwise, boto3 defaults to verify=True).
```

Sorry, I hope this is the last one. There are multiple options for boto3 how to handle CAs, e.g. `ca_bundle` option in the aws config file(AWS_CONFIG_FILE), or env variables like(AWS_CA_BUNDLE, or REQUESTS_CA_BUNDLE)

mbuechse: No, sure, I do welcome the improvement. I for one really benefit from clear documentation.

PR #1153, opened 2026-04-10
Relax scs-0210-v2: extend update time for patch versions

### LINKED ISSUE #1098: [Feature Request] Reduce kubernetes version policy for minor versions
This came up during the RKE2 session at the hackathon. [SCS 0210 ](https://docs.scs.community/standards/scs-0210-v2-k8s-version-policy) states:

> The latest patch version MUST be provided no later than 2 weeks after release.

janeczku raised that there is no apparent reason for this to be a MUST, since CVEs are handled in a separate clause and for minor versions we allow four months.
Instead of choosing another (random) time frame, we discussed it might make sense to switch MUST for a SHOULD.

mbuechse:
This comes after comprehensive discussion in SIG Std/Cert (of 2026-04-09), as well as research by Syself into industry best practices.

### DIFF
--- a/Standards/scs-0210-v2-k8s-version-policy.md
+++ b/Standards/scs-0210-v2-k8s-version-policy.md
@@ -57,7 +57,7 @@ In order to keep up-to-date with the latest Kubernetes features, bug fixes and s
 the provided Kubernetes versions should be kept up-to-date with new upstream releases:
 
 - The latest minor version MUST be provided no later than 4 months after release.
-- The latest patch version MUST be provided no later than 2 weeks after release.
+- The latest patch version MUST be provided no later than 1 month after release.
 - This time period MUST be even shorter for patches that fix critical CVEs.
   In this context, a critical CVE is a CVE with a CVSS base score >= 8 according
   to the CVSS version used in the original CVE record (e.g., CVSSv3.1).
--- a/Standards/scs-0210-v2-k8s-version-policy.md
+++ b/Standards/scs-0210-v2-k8s-version-policy.md
@@ -58,10 +58,11 @@ the provided Kubernetes versions should be kept up-to-date with new upstream rel
 
 - The latest minor version MUST be provided no later than 4 months after release.
 - The latest patch version MUST be provided no later than 1 month after release.
-- This time period MUST be even shorter for patches that fix critical CVEs.
+- This time period (for providing the latest patch version) MUST not be longer than
+  two weeks if the patch addresses critical CVEs and it is RECOMMEDED to be provided
+  within 2 days of the release.
   In this context, a critical CVE is a CVE with a CVSS base score >= 8 according
   to the CVSS version used in the original CVE record (e.g., CVSSv3.1).
-  It is RECOMMENDED to provide a new patch version in a 2-day time period after their release.
 - New versions MUST be tested before being rolled out on productive infrastructure;
   at least the [CNCF E2E tests][cncf-conformance] should be passed beforehand.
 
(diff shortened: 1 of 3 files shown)


janiskemper (approved):
LGTM (of course let's wait for feedback from others)

depressiveRobot (approved):
Fixes #1098.

The requirement has been relaxed, so a new version of the standard is not necessary.

Looks good.

Standards/scs-0210-v2-k8s-version-policy.md
```diff
@@ -57,7 +57,7 @@ In order to keep up-to-date with the latest Kubernetes features, bug fixes and s
 the provided Kubernetes versions should be kept up-to-date with new upstream releases:
 
 - The latest minor version MUST be provided no later than 4 months after release.
-- The latest patch version MUST be provided no later than 2 weeks after release.
+- The latest patch version MUST be provided no later than 1 month after release.
 - This time period MUST be even shorter for patches that fix critical CVEs.
```
garloff: We are not precise here (how much shorter) here, as we just have a RECOMMENDATION to provide this within 2 days. This lack of precision is not new, it just got a bit more relevant by extending the time period. I suggest we address this in a followup discussion in the future.
Maybe it's <=1 months for normal patch versions, a REQUIREMENT of <= 2 weeks for CVSS>=8 issues and a RECOMMENDATION for <= 2 days for CVSS>8 then. (I feel like 1 week would be more appropriate as a hard requirement, but the 2 week would have the advantage of matching our previous hard requirement.)

garloff (approved):
Please consider my comment for a followup discussion and potential improvement.
This is not meant to block this agreed change, so approve.

Standards/scs-0210-v2-k8s-version-policy.md
mbuechse: If we want to implement something like this, we should do it now, because once the standard is relaxed, we would need to introduce a new version to make it more strict again.

janiskemper: Wouldn't it be anyway a relaxation for the general patch release, and a stricter version for CVE's @mbuechse ? 

In which way would we NOT have to add a new version of this based on what Kurt proposes?

mbuechse: Right now, any patch must be provided no later than 2 weeks. In the future, this holds only for those patches with CVSS => 8, whereas patches with CVSS < 8 have 1 month.

janiskemper: if I'm not mistaken, Kurt proposed to do CVE's in one week instead of two. The I guess we are on the same page that this would be stricter and would require a new version. 

If we do it with two weeks, then I agree that it wouldn't be stricter. 

@garloff WDYT?

mbuechse: Kurt proposed 2 weeks, and in parentheses he expressed a preference for 1 week, but he seemed to want to spare us the hassle of a new version. And the new version could still be done later anyway, if it finds a majority. Not sure if 1 week is at all realistic (we still have 2 days for CVSS > 8 as a recommendation)

garloff: OK, so my pragmatic suggestion is:
- One month *requirement* for any patch release
- Two weeks *requirement* for CVSS >= 8
- Two days *recommendation* for CVSS >= 8

That would be a relaxation (for non-CVSS>=8 patches) and thus not require a new version.

garloff: So, the language for this is in https://github.com/SovereignCloudStack/standards/pull/1153/commits/9bf23800741e654992fc5822d7f0a6654f388301. Feel fee to further wordsmith or reject.

garloff:
@mbuechse : Ready for review?

garloff:
My +1 vote for merging!

mbuechse:
I'm just realizing that the test so far has not been updated to reflect the new language...

PR #1154, opened 2026-04-15
add standard for CNCF Kubernetes conformance

### LINKED ISSUE #1022: Create standard for CNCF Conformance
This is analogous to #1010 -- if we introduce changes (such as #982), we should replace the plain link to the original CNCF location with a standard of our own (that merely explains the adjustments).

depressiveRobot:
Caused by #1022 

This PR introduces a draft for a new standard scs-0201-v1 and a corresponding supplement. It also adds versioning to the Sonobuoy configuration.


Standards/scs-0201-v1-cncf-conformance.md
```diff
@@ -0,0 +1,27 @@
...
+
+The SCS project always intended to reuse existing standards whenever possible. For this reason, the upstream tooling of the [Certified Kubernetes Conformance Program](https://github.com/cncf/k8s-conformance) is used for conformance testing.
+
+## Motivation
+
+As an operator as well as an user, I want to check the conformance of a Kubernetes environment in order to ensure the interoperability, consistency, and traceability of it.
+
+## Regulations
+
+The conformance testing is performed as part of the [SCS compliance check suite](https://github.com/SovereignCloudStack/standards/tree/main/Tests) executing the standard set of conformance tests defined by the `[Conformance]` tag in the [Kubernetes e2e suite](https://github.com/kubernetes/kubernetes/tree/master/test/e2e). All conformance tests MUST be passed successfully.
```
janiskemper, line 22: we also test something else (network tests), as seen in https://github.com/SovereignCloudStack/standards/blob/c1e9a1b761a0bb201e80d9207f4a1dd0a68ff57f/Tests/scs-compatible-kaas.yaml#L38.

From my understanding, those are actually the heavy tests.

mbuechse: For the network tests, we already have a standard -- scs-0219.

Both the network tests and cncf k8s conformance are quite heavy in terms of wall time: the former might take something like 3 hrs, the latter 2 hrs when parallel execution in Sonobuoy is not enabled (however, parallel execution has become the default for us).

janiskemper: IMO the network tests test a lot of things that are not related to our standard... Do you really wanna make this 1 on 1 relationship between Kubenernetes network tests and the standard 0219?

janiskemper: IMO they are a huge overkill if we want to just test 0219. I always considered them as part of the Kubernetes tests that we wanna run, so more in direction "Kubernetes conformance" rather than "SCS networking". 

To justify that we run these tests, I believe we need a better reason than our standard. A simple "apply one NetworkPolicy" would suffice to test the standard.

depressiveRobot: Those are some good points you've made, @janiskemper. We should take a closer look at this. However, this PR is about the CNCF conformance tests, which do not include the NetworkPolicy tests. Feel free to create another issue for that.

mbuechse: Well. I honestly don't know. When we introduced scs-0219, we needed tests that cover support of network policies (maybe not just that you can create and apply one, but also that is then being conformed to by the system in all the various ways). And the colleague who created the standard (I was not involved in the actual subject matter) apparently deemed the Sonobuoy tests from the k8s networking group appropriate and selected them specifically to test the standard (that much I know). If you think this needs a debate, then we should debate that in the SIG or in Team Container, but that's the status quo, and from the POV of this status quo, this PR here has nothing to do with the networking tests or any Sonobuoy tests at all, except for those that are to do with CNCF k8s conformance.

mbuechse: (Which is clear from this standard's title and description)

Standards/scs-0201-v1-cncf-conformance.md
```diff
@@ -0,0 +1,27 @@
...
+track: KaaS
+description: |
+  SCS-0201 describes conformance testing of Kubernetes environments using the upstream CNCF Kubernetes e2e suite.
+---
+
+## Introduction
+
+Interoperability, consistency, and traceability play a crucial role in the deployment and use of a Kubernetes environment. Testing ensures this conformance across different Kubernetes environments.
+
+The SCS project always intended to reuse existing standards whenever possible. For this reason, the upstream tooling of the [Certified Kubernetes Conformance Program](https://github.com/cncf/k8s-conformance) is used for conformance testing.
```
mbuechse, line 14: We reuse (or, rather, wrap and adapt) the rules and regulations regarding CNCF k8s conformance, see https://github.com/cncf/k8s-conformance/blob/master/terms-conditions/Certified_Kubernetes_Terms.md

So, it's much more than just the tooling that we use here.

Standards/scs-0201-w1-implementation-testing.md
```diff
@@ -0,0 +1,19 @@
...
+title: "CNCF Kubernetes conformance: Implementation and Testing Notes"
+type: Supplement
+track: KaaS
+supplements:
+  - scs-0201-v1-cncf-conformance.md
+---
+
+## Implementation notes
+
+The actual execution of the conformance tests is performed by the [SCS compliance check suite](https://github.com/SovereignCloudStack/standards/tree/main/Tests) using [Sonobuoy](https://sonobuoy.io). The rationale for this decision and further details regarding its implementation are documented in detail in [scs-0200-v1](https://docs.scs.community/standards/scs-0200-v1-using-sonobuoy-for-kaas-conformance-tests).
```
mbuechse: Not having looked at the decision record, I think it is NOT about using Sonobuoy for running official CNCF testcases, but rather about Sonobuoy for running novel, SCS-specific testcases (i.e., testcases that we implement ourselves).

Standards/scs-0201-v1-cncf-conformance.md
janiskemper: okay.. Until now I thought that we put it all together, especially because we mention the e2e test suite even in the standard: https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0219-v1-kaas-networking.md#conformance-tests

That made me believe that we make no different and just use extended Kubernetes e2e tests. 

If we say that we don't do it, that we just wanna have the conformance tests, then a 2 hours test run is for sure overkill to check whether a single custom resource can be applied. 

If you don't want to add the Kubernetes e2e networking tests "officially" to the standard here, then the proper way of handling it would be to replace it and use a simpler test. However, then someone has to implement it :sweat_smile:

Tests/kaas/sonobuoy_handler/run_sonobuoy.py
```diff
@@ -20,7 +20,7 @@ def cli():
 @click.option("-k", "--kubeconfig", "kubeconfig", required=True, type=click.Path(exists=True), help="path/to/kubeconfig_file.yaml",)
 @click.option("-r", "--result_dir_name", "result_dir_name", type=str, default="sonobuoy_results", help="directory name to store results at",)
 @click.option("-c", "--check", "check_name", type=str, default="sonobuoy_executor", help="this MUST be the same name as the id in 'scs-compatible-kaas.yaml'",)
-@click.option("--scs-sonobuoy-config", "scs_sonobuoy_config_yaml", type=click.Path(exists=True), default="kaas/sonobuoy-config.yaml", help="Configuration for Sonobuoy (yaml format)")
+@click.option("--scs-sonobuoy-config", "scs_sonobuoy_config_yaml", type=click.Path(exists=True), default="kaas/scs-sonobuoy-config-v1.yaml", help="Configuration for Sonobuoy (yaml format)")
```
mbuechse: While we are at it, I would actually prefer to remove the default value entirely. People should know what they're doing, and if you can run the tool without specifying the YAML yourself, then it's entirely possible that you do something wrong and get unintended results without even noticing.

Besides, once a v2 is added, the default no longer helps save time. Instead, I would always propose to not use this script directly, but `scs-test-runner.py` (which at some point might be extended so a regex or list of desired testcases could be specified).

Standards/scs-0201-v1-cncf-conformance.md
janiskemper: And until then we should specify the actual tests that we run for the standard 0219. Instead of "e2e test" we should maybe write "Kubernetes networking e2e tests" and specify the sonobuoy flag that is used.

Tests/kaas/sonobuoy_handler/sonobuoy_handler.py
```diff
@@ -42,7 +42,7 @@ def __init__(
         check_name="sonobuoy_handler",
         kubeconfig=None,
         result_dir_name="sonobuoy_results",
-        scs_sonobuoy_config_yaml="kaas/sonobuoy-config.yaml",
+        scs_sonobuoy_config_yaml="kaas/scs-sonobuoy-config-v1.yaml",
```
mbuechse: While we are at it, I would definitely remove default value here, as this is not user-facing.

Tests/scs-compatible-kaas.yaml
```diff
@@ -12,14 +12,14 @@ variables:
...
-  args: run -k {kubeconfig} --scs-sonobuoy-config kaas/scs-sonobuoy-config.yaml -r {subject_root}/sono-results-e2e -c 'cncf-k8s-conformance'  -a '--mode=certified-conformance' --execution-mode {execution_mode}
+  args: run -k {kubeconfig} --scs-sonobuoy-config kaas/scs-sonobuoy-config-v1.yaml -r {subject_root}/sono-results-e2e -c 'cncf-k8s-conformance'  -a '--mode=certified-conformance' --execution-mode {execution_mode}
   #~ args: -k {kubeconfig} -r {subject_root}/sono-results -c 'cncf-k8s-conformance'  -a '--plugin-env e2e.E2E_DRYRUN=true'
   testcases:
   - id: cncf-k8s-conformance
     lifetime: year
     section: heavy
-    description: Must fulfill all requirements of _CNCF Kubernetes conformance_.
-    url: https://github.com/cncf/k8s-conformance/tree/master
+    description: Must fulfill all requirements of scs-0201-v1.
```
mbuechse: I would keep the reference to CNCF k8s conformance (for instance, in parentheses after the standard).

mbuechse (commented):
Thanks for the initiative! I have a few remarks. I can imagine some more polishing (beyond my remarks), but for the first draft, I wouldn't be too pedantic.

Standards/scs-0201-v1-cncf-conformance.md
mbuechse: I'm pretty sure there is some misunderstanding here. For scs-0219, we don't run the entire e2e test suite, but only the test cases from the networking section. Those test cases go beyond CNCF conformance, but they are specific to networking. And again, we want to test that the network policies *work as intended*, not just that the custom resource can be applied.

fkr:
I have not the needed insights into kubernetes networking to be able to review this, so I removed myself from the list of reviewers.

Standards/scs-0201-v1-cncf-conformance.md
mbuechse: I mean, it specifically includes `'--e2e-focus "NetworkPolicy"'` -- NetworkPolicy! How much more specific can we get???

mbuechse:
> I have not the needed insights into kubernetes networking to be able to review this, so I removed myself from the list of reviewers.

Don't be confused by the discussion above. This standard is not about kubernetes networking. It's our wrapper around CNCF kubernetes conformance.

Standards/scs-0201-v1-cncf-conformance.md
janiskemper: I'm referring to the fact that in the standard we don't say that ;) We say "upstream e2e tests". So neither @depressiveRobot nor me understood that we refer to the networking e2e tests that we call in our test suite.

janiskemper: "work as intended" is very generic and I cannot really read that from https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0219-v1-kaas-networking.md#decision.

Just a general comment.

mbuechse: It does state (emphasis by me)

> CSPs MUST provide a network plugin that *fully supports* NetworkPolicy resources

Also,

> Required support for network policies will be tested using the upstream e2e tests via Sonobuoy

could indeed be read as "all existing upstream e2e tests", but that doesn't sound quite reasonable to me, as opposed to "all relevant upstream e2e tests". Indeed, the first thing you posted in this thread was indeed the command line from the YAML file, which expressly refers to `--e2e-focus "NetworkPolicy"`, so I have a hard time believing you really misread this.

At any rate, let's improve the language in scs-0219. I still don't understand what this has to do with this PR.

janiskemper: Well I only made the connection between this comment in 0219 and the e2e (networking) tests that we start in our script today after reading your comment from above :laughing: And that's why I can say that this sentence in the standard is indeed not understandable as "networking tests that we start with sonobuoy using `--e2e-focus "NetworkPolicy"`.

mbuechse: Got it. I will see to it that things will become more clear.

janiskemper: Great! I'll resolve this now to not litter this PR.

Tests/kaas/sonobuoy_handler/run_sonobuoy.py
depressiveRobot: Agree. I removed the default value and made the config option mandatory to avoid an exception later on.

Tests/kaas/sonobuoy_handler/sonobuoy_handler.py
depressiveRobot: Related to [this comment](https://github.com/SovereignCloudStack/standards/pull/1154#discussion_r3116907215). Fixed it.

Tests/scs-compatible-kaas.yaml
depressiveRobot: Added the reference.

Standards/scs-0201-w1-implementation-testing.md
depressiveRobot: I took another look and you're right. Removed the reference to the decision record.

Tests/kaas/sonobuoy_handler/run_sonobuoy.py
```diff
@@ -33,7 +33,7 @@ def sonobuoy_run(kubeconfig, result_dir_name, check_name, scs_sonobuoy_config_ya
 
 
 @cli.command("check-results")
-@click.option("--scs-sonobuoy-config", "scs_sonobuoy_config_yaml", type=click.Path(exists=True), default="kaas/scs-sonobuoy-config.yaml", help="Configuration for Sonobuoy (yaml format)")
+@click.option("--scs-sonobuoy-config", "scs_sonobuoy_config_yaml", type=click.Path(exists=True), default="kaas/scs-sonobuoy-config-v1.yaml", help="Configuration for Sonobuoy (yaml format)")
```
mbuechse: Sorry, I didn't mention this occurrence here. Could you remove the default here as well?

Standards/scs-0201-v1-cncf-conformance.md
mbuechse: @janiskemper see this PR: https://github.com/SovereignCloudStack/standards/pull/1158

Standards/scs-0201-v1-cncf-conformance.md
depressiveRobot: IMO, those are the terms and conditions for the Linux Foundation's certification program, and we do not adopt them. However, we do use the same tool to check for conformance. That's why I chose this formulation.

Tests/kaas/sonobuoy_handler/run_sonobuoy.py
depressiveRobot: Good catch. Done :)

Standards/scs-0201-v1-cncf-conformance.md
mbuechse: True. My point is that we not only use Sonobuoy, but the whole suite of testcases, which is independent of Sonobuoy (one could use hydrophone, for instance).

mbuechse: Apparently, the problem here is that for the CNCF, the test cases are (only) specified by means of the test code. From a certain point of view, one might say that this code is part of the tooling, but IMO this doesn't to it justice. Maybe we can just find an alternative for the word "tooling". Maybe "test suite" or something.

mbuechse: Oh, I would also change "intended" to "intends".

depressiveRobot: Thanks for catching the misspelling. I fixed it and changed the wording to "test suite" as suggested.

mbuechse (approved):
Looks good to me. However, it could be worth to consider whether we want to adopt the style of [scs-0219](https://docs.scs.community/standards/scs-0219-w1-kaas-networking#automated-tests), where we address the exceptions only in the supplement.

mbuechse:
Oh, the link in scs-0219-w1 would need to be updated (however, merge main or rebase to main first)

PR #1156, opened 2026-04-16
clarify meaning of subject in SCS-0004

depressiveRobot:
Be more explicit what subject means in the context of SCS-0004 certification.


Standards/scs-0004-v1-achieving-certification.md
```diff
@@ -20,7 +20,7 @@ As operator, I want to obtain a certificate with the scope SCS-compatible IaaS o
 
 0. Certificates are issued by the SCS certification assessment body. This is either the _Forum SCS-Standards_, part of the Open Source Business Alliance e.V. (OSBA), or an organization appointed by the _Forum SCS-Standards_ . An interested party has to apply for certification with this body, which in turn determines the rules that govern what parties are eligible for application (fees may apply).
 
-1. Each certificate issued pertains to a given combination of subject (i.e., cloud environment), scope (such as _SCS-compatible IaaS_), and version of that scope. The certificate is only valid for that combination and for the time frame that ends when the scope expires, or for six months if the expiration date for the scope is not yet fixed.
+1. Each certificate issued pertains to a given combination of subject (specific cloud environment including IaaS region or KaaS configuration), scope (such as _SCS-compatible IaaS_ or _SCS-compatible KaaS_), and version of that scope. The certificate is only valid for that combination and for the time frame that ends when the scope expires, or for six months if the expiration date for the scope is not yet fixed.
```
mbuechse: Interesting: a certificate subject is not the same thing as a test subject: the certificate subject can actually be a set of multiple test subjects (multiple regions in IaaS case or multiple configurations in KaaS case).

depressiveRobot: That's right. As you mentioned, multiple test objects can be certified collectively under a single subject. I updated it to the plural form right away to emphasize this even more.

mbuechse (approved):
Very minor change. It's okay, I guess. Not sure who else should be consulted.

PR #1162, opened 2026-04-28
Fix argument order in SonobuoyHandler call

mbuechse:
Fallout from https://github.com/SovereignCloudStack/standards/pull/1154


depressiveRobot (approved):
Oops 😅 Thanks for fixing. LGTM

PR #1166, opened 2026-04-30
Update SCS-0219 after retirement of ingress-nginx and add Gateway API

### LINKED ISSUE #1097: [Other] Update SCS-0219 after retirement of ingress-nginx and add Gateway API
Hi,

in the recent version of SCS-0219, ingress-nginx is still mentioned as example for the ingress-API despite that the project is retired starting April 2026 (https://www.kubernetes.dev/blog/2025/11/12/ingress-nginx-retirement/). 
Maybe update the examples to some other CNCF projects like Traefik or something similar.

Besides this, it would be beneficial if Gateway API, the successor, even if it is not supported from all applications, is at least mentioned together with some CNCF examples which support it. I know, ingress-api is not really deprecated at this point, but since the modern approach and successor is already out there, it should get an "RECOMMEND" or "SHOULD" approach if possible. Also because it is recommended by the kubernetes project itself.

lucasrattz:
This pull request fixes #1097

### DIFF
--- a/Standards/scs-0219-v1-kaas-networking.md
+++ b/Standards/scs-0219-v1-kaas-networking.md
@@ -10,7 +10,7 @@ track: KaaS
 
 Kubernetes defines a networking model that needs to be implemented by a separate CNI plugin.
 Beyond basic connectivity within the cluster, however, there are many networking features that are specified but optional.
-Some of these optional features provide vital functionality, such as the NetworkPolicy API and the Ingress API.
+Some of these optional features provide vital functionality, such as the NetworkPolicy API and the Gateway API.
 
 This standard specifies a minimal set of networking features that users can expect in clusters created by an SCS-compliant KaaS provider.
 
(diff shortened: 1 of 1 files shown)


berendt:
Don't we need a new version of this document here?

https://docs.scs.community/standards/scs-0001-v1-sovereign-cloud-standards#stabilized-phase-stable

"If a breaking change to an existing SCS document is deemed necessary, a new document with a new number shall be created and the old document SHOULD be deprecated."

jklare:
Since the Ingress API will continue to exist (https://gateway-api.sigs.k8s.io/faq/#will-gateway-api-replace-the-ingress-api) and the Gateway API is basically just an entirely new and independent parallel implementation, we can probably not just replace one with another in the standards (since this is not an update, but a switch towards another implementation).

mbuechse:
@jklare You're right. The point here is that we're only replacing one recommendation by another (if I'm not mistaken). This is not a breaking change, because every test subject that passes the old version will also pass the new one. The update is with respect to Traefik instead of nginx.

berendt:
Adding a new normative SHOULD ("CSPs SHOULD offer support for the Gateway API …") is, in my understanding, a substantive extension of the normative requirements for CSPs, which requires a v2 if the document has already been stabilized.

mbuechse:
@berendt If this matter is so controversial, we should discuss it more. I added it to the agenda of this week's SIG Std/Cert.

mbuechse:
@lucasrattz The SIG Std/Cert on May 7 has decided that a new major version is desired. I have implemented this in #1195. Please tell me if you're okay with me copying your text in this way. I suppose I can add a stanza (such as Co-authored-by) to the commit, or something similar. Please tell me what works for you.

lucasrattz:
> @lucasrattz The SIG Std/Cert on May 7 has decided that a new major version is desired. I have implemented this in #1195. Please tell me if you're okay with me copying your text in this way. I suppose I can add a stanza (such as Co-authored-by) to the commit, or something similar. Please tell me what works for you.

Yes, that's fine.

PR #1168, opened 2026-05-06
set SCS-0004 certificate validity period to one year

depressiveRobot:
Currently, a certificate according to SCS-0004 is only valid for six month, if no expiration date is given for the active scope (see [regulation 1](https://docs.scs.community/standards/scs-0004-v1-achieving-certification#regulations)). This might be too short with regard to the effort and possible fees. The current definition also creates uncertainty in planning if it is not clear how long the certificate will be valid.

Therefore, the PR sets the validity period to one year regardless of the expiry date of the active scope. This also aligns the validity period with that of the integrator certification (SCS-0007).


Standards/scs-0004-v1-achieving-certification.md
```diff
@@ -20,7 +20,7 @@ As operator, I want to obtain a certificate with the scope SCS-compatible IaaS o
...
 
-1. Each certificate issued pertains to a given combination of subject (specific cloud environment including IaaS regions or KaaS configurations), scope (such as _SCS-compatible IaaS_ or _SCS-compatible KaaS_), and version of that scope. The certificate is only valid for that combination and for the time frame that ends when the scope expires, or for six months if the expiration date for the scope is not yet fixed.
+1. Each certificate issued pertains to a given combination of subject (specific cloud environment including IaaS regions or KaaS configurations), scope (such as _SCS-compatible IaaS_ or _SCS-compatible KaaS_), and version of that scope. The certificate is only valid for that combination for the period of one year. The certificate remains valid for this period, even if the scope version expires for which it applies.
```
mbuechse:
```suggestion
1. Each certificate issued pertains to a given combination of subject (specific cloud environment including IaaS regions or KaaS configurations), scope (such as _SCS-compatible IaaS_ or _SCS-compatible KaaS_), and version of that scope. The certificate is only valid for that combination for the period of one year (regardless of the expiration date of the scope version).
```

mbuechse (approved):
Sorry, I seem to have forgotten about this.

PR #1177, opened 2026-05-20
Limit waiting times when provisioning/unprovisioning clusters

mbuechse:
otherwise this can wait indefinitely which can then stall the whole CI/CD job


Tests/kaas/plugin/plugin_clusterstacks.py
```diff
@@ -15,6 +15,7 @@
 
 
 TEMPLATE_KEYS = ('cluster', 'clusterstack', 'kubeconfig')
+TIMEOUTS = [30] * 30 + [0]  # wait at most 900 seconds (15 minutes); sentinel value at the end
```
janiskemper: are you sure that this is enough? Otherwise I propose maybe 20 min for starting

janiskemper (approved):

PR #1180, opened 2026-05-27
Fix db bootstrap logic

toothstone:
An empty db has a current version of 'None', so we have to make sure to check for that first and leave the loop before encountering the '>=' comparison to avoid a TypeError.


mbuechse (approved):
Thanks a lot! The mistake has been spotted independently (#1171). Your solution is quite elegant.

PR #1184, opened 2026-06-01
compliance-monitor: add column for KaaS technology

jklare:
* removed "based on Gardener" and "based on ClusterStacks" from Description column and move it to Technology column


jklare:
closed as stale

mbuechse:
The table is currently growing, so we might take this up again!

PR #1191, opened 2026-06-03
Let compliance monitor redirect from / to /page/table

mbuechse:


compliance-monitor/monitor.py
```diff
@@ -389,10 +389,7 @@ def check_role(account: Optional[tuple[str, str]], subject: str = None, roles: i
 
 @app.get("/")
 async def root():
-    # we might use the following redirect in the future:
-    # return RedirectResponse("/pages")
-    # but keep this silly message for the time being, so as not to expose the work in progress too much
-    return {"message": "Hello World"}
+    return RedirectResponse("/page/table")
```
gerbsen: I think this should be

`return RedirectResponse(f"{settings.base_url}page/table")`

so that it also redirects if non-/ environments are deployed.

compliance-monitor/monitor.py
```diff
@@ -777,14 +774,6 @@ async def get_scope(
     return render_view(VIEW_SCOPE, view_type, spec=spec, relevant=relevant, rows=rows, base_url=settings.base_url, title=spec['name'])
 
 
-@app.get("/pages")
```
gerbsen, line 780: I don't understand why this is being deleted. Does this have anything to do with the redirect?

compliance-monitor/monitor.py
mbuechse: So far, I've used the base url only to include the domain, which was necessary because the fragmets get embedded into the docs page, which lives on another domain. Are you sure it's possible to have another root? In my mind, the reverse proxy would have to remove that. Yes, I think, you're right. Will change that!

compliance-monitor/monitor.py
mbuechse: This is an incidental catch. The endpoint `/pages` is undocumented, I can't imagine anyone using it, and it serves no purpose whatsoever.

gerbsen (approved):

Issue #1196, opened 2026-06-05
[Cleanup] Folders Drafts, Operational-Docs


mbuechse:
These folders have been either closed for addition or been neglected for at least 3 years.

Each file should receive one of the following four treatments:

1. promote to a 'real' draft (in the ADR style) in the `Standards` folder,
2. move to the `docs` repo,
3. delete,
4. keep (if absolutely necessary).

If any files are kept at all, move those to a new folder `Archive` or `Attic`.

My hunch is the following.

1. promote these files to real decision records:
    - (done) https://github.com/SovereignCloudStack/standards/blob/main/Drafts/OSS-Health.md
    - (done) https://github.com/SovereignCloudStack/standards/blob/main/Drafts/flavor-naming-strategy.md
2. move these to docs repo:
    - https://github.com/SovereignCloudStack/standards/blob/main/Drafts/Release-Policies.md
3. delete all others.

mbuechse:
@garloff I added two PRs as suggestions for item 1 in this issue's description. Please have a look, and please evaluate items 2 and 3.

mbuechse:
@jklare I just saw that you were working on the release policies file, even introducing "must" and "should". Maybe then it should be promoted to a real standard as well.

PR #1201, opened 2026-06-06
Promote draft on flavor naming to decision record

mbuechse:
Contributes to #1196


garloff (approved):
Thanks, @mbuechse.
This is how it should be categorized, yes.

PR #1204, opened 2026-06-10
Defuse scs-0123-swift-s3

mbuechse:
This testcase had multiple problems:

- It is not mentioned in the normative part of the standard (only in the section on testing)
- What user demand it is supposed to satisfy is unclear
- After recent updates to Keystone this no longer works with restricted (normal) application credentials

Its benefit (if any) does not outweigh its cost.

Downgraded to only test for presence of S3, but nothing in the way of functionality or consistency with Swift.

### DIFF
--- a/Tests/scs-compatible-iaas.yaml
+++ b/Tests/scs-compatible-iaas.yaml
@@ -481,6 +481,7 @@ modules:
       - scs-0123-service-load-balancer
       - scs-0123-service-placement
       - scs-0123-storage-apis
+      recommended:
       - scs-0123-swift-s3
   - id: scs-0123-v2
     name: Mandatory and Supported IaaS Services
--- a/Tests/scs-compatible-iaas.yaml
+++ b/Tests/scs-compatible-iaas.yaml
@@ -481,7 +481,6 @@ modules:
       - scs-0123-service-load-balancer
       - scs-0123-service-placement
       - scs-0123-storage-apis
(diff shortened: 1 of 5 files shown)


Tests/scs-compatible-iaas.yaml
```diff
@@ -481,6 +481,7 @@ modules:
       - scs-0123-service-load-balancer
       - scs-0123-service-placement
       - scs-0123-storage-apis
+      recommended:
```
fzakfeld: The Keystone EC2/S3 integration is very limited in functionality and has a track record of bad security issues, so to me it feels wrong calling it `recommended`. I'd open the debate to completely remove it from the standard

mbuechse: Thanks for your input! That's precisely what I had hoped to get (not necessarily the specifics, but some push-back in general)

mbuechse: It would be good to test whether an S3 api is present at all, without doing any privileged operation (i.e., without having credentials). Then again, how many real users expect S3 at the same URL as Swift?

depressiveRobot: From the POV of certification, I advocate a simple test if the S3 API is present as you are suggesting.

depressiveRobot (approved):
Thanks for the simple s3 API check. Looks fine now.

PR #1205, opened 2026-06-12
Simplify compliance monitor views

mbuechse:
- Remove deprecated scope versions, because they are only achieved by accident (no one is actively working on supporting them), and they make the views (even) more complicated.
- Restructure details view a bit.

### DIFF
--- a/Tests/scs-compatible-iaas.yaml
+++ b/Tests/scs-compatible-iaas.yaml
@@ -501,6 +501,10 @@ modules:
       preview:
       - domain-manager-check
 timeline:
+  - date: 2026-06-11
+    versions:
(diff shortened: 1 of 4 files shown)


mbuechse:
I don't have a before-after for the part about the older scope versions. But I have a before-after for the details view:

Before:

Screenshot from 2026-06-12 14-17-11

After:

Screenshot from 2026-06-12 14-17-30

mbuechse:
After incorporating suggestions by @jklare:

Screenshot from 2026-06-12 15-40-00

jklare (approved):
LGTM, thanks!

PR #1208, opened 2026-06-14
Fix k8s_version_policy.py to log on stderr (instead of stdout)

mbuechse:


mbuechse:
Result of version policy check cannot be seen on
https://compliance.sovereignit.cloud/page/report/62965e1f-87d2-4228-988e-0d2d6b42a127#version-policy-check
because it erroneously gets logged on stdout.

(This will change once again with #1073 or so, but for now, this fix is good.)

depressiveRobot (approved):
Really a small change, so this is fine.

However, I still don't quite understand it. Do you mean that in the redacted state, the reason why the test fails cannot be seen?

mbuechse:
Yes, in the redacted state, only lines starting with ERROR or WARNING _on stderr_ are shown. Also, no one would expect them in stdout, even if they were shown.

PR #1210, opened 2026-06-14
Run test scripts in predefined order

mbuechse:
... instead of some nondeterministic order


depressiveRobot (approved):
Thanks for fixing this.

PR #1211, opened 2026-06-14
Increase number of workers in preset 'iaas' from 4 to 16

mbuechse:
These workers don't use a lot of resources except wall time


depressiveRobot (approved):
👍

PR #1213, opened 2026-06-16
Feat/more gpus

garloff:
Adding one letter `u` for Blackwell Ultra (GB300) and a lot more examples in the GPU table.
Added a disclaimer that providers need to use real values and not rely on data sheet fed table values.

### DIFF
--- a/Standards/scs-0100-v3-flavor-naming.md
+++ b/Standards/scs-0100-v3-flavor-naming.md
@@ -436,7 +436,8 @@ Note that the vendor letter X is mandatory, generation and processing units are 
 | `A`      | AMD    | compute units (CUs)             |
 | `I`      | Intel  | execution units (EUs)           |
 
-For nVidia, the generation N can be f=Fermi, k=Kepler, m=Maxwell, p=Pascal, v=Volta, t=turing, a=Ampere, l=Ada Lovelace, g=Grace Hopper, b=Blackwell, ...,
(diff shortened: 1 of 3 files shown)


mbuechse:
Good luck obtaining them 😂

mbuechse:
> Do tests and tooling need to be adjusted as well?

I think the "u" needs to be added here:

https://github.com/SovereignCloudStack/standards/blob/42773200673ea53e200126e1781bb3abf44ff033/Tests/iaas/scs_0100_flavor_naming/flavor_names.py#L236

garloff:
Apparently I wasn't awake ...
Thanks for noticing!

garloff:
GPU providers: Please have a look and let us know whether the updated table makes sense for you.
@fzakfeld @chrisschwa @Nils98Ar @horazont

garloff:
Feedback welcome, also from others of course ...

garloff:
I believe this should be a non-brainer: We evolve our implementation hints (and for the blackwell ultra also our standard) along the most recent hardware. This keeps it viable and consistent.

mbuechse:
@garloff I think this is undisputed. We wanted to give operators the chance to weigh in (and maybe dispute it for the first time), but (apparently) we can't pressure them to do so...

mbuechse (approved):
No one opted to weigh in. And I don't see a reason to block this any longer.

PR #1217, opened 2026-06-16
Make scs-0210-v2 a bit more precise

mbuechse:
All active k8s branches must be supported AND offered, which means not only do customers get support for existing clusters, but they must be able to create a cluster for each of those k8s brances.

Also remove redundant section on testing (which is covered by supplement).


garloff (approved):
That's what we discussed and agreed upon.
Thanks, @mbuechse !

PR #1190, opened 2026-06-03
Add task to debug Zuul encryption

mbuechse:

PR #1178, opened 2026-05-22
Add ScaleUp KaaS Offering to Compliance Monitor

fzakfeld:
Currently in preperation, please treat this as a draft for now

PR #1218, opened 2026-06-17
Refresh scaleup_api_server secret.

garloff:
It's not a secret secret, so this is uncritical.
We use it as a canary since #1190.
As the webhook was broken before, we would not run the check and post pipelines on it. We should now do so, so refresh the secret.


garloff:
A note on this one:
- zuul-merger does complex things when it calculates the future state
- Somehow the fact that the webhook was not working for a while made it work with an outdated state. This may be what broke #1178
- The change (which is just reencoding the same value) does touch the secure.yaml file so it gets in sync again
- I'd like to observe the check (scs-check-adr-syntax - already ran successfully) and the post (scs-check-scs2) pipelines to succeed. The latter should run after this PR has seen the needed approvals.
- Ideally, upon merging, the hourly pipeline (scs-check-scs2-main) starts working again

fzakfeld (approved):
Thanks for the fix, if other (actual) secrets are broken as well, let me know

garloff:
Well, your approval did not trigger the scs-check-scs2 pipeline, unfortunately. But that's a different thing to look at.
I'll merge and see if this fixes our hourly pipeline.

PR #1220, opened 2026-06-17
Add k8s branch 1.35 to tests

mbuechse:
This includes two improvements:

- test the k8s branches sequentially to avoid hitting quotas
- introduce alias 'autoVars/scaleup' for 'autoVars/noris' to reduce potential for confusion

### DIFF
--- a/.zuul.d/config.yaml
+++ b/.zuul.d/config.yaml
@@ -6,7 +6,9 @@
     periodic-daily:
       jobs:
         - scs-check-iaas
-        - scs-check-kaas
+        - scs-check-kaas1
+        - scs-check-kaas2
+        - scs-check-kaas3
     periodic-hourly:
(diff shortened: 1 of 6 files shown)


mbuechse:
@fzakfeld This will only work when you change your classification from 'preview' to 'supported'.

fzakfeld:
We plan to do this in the following days. I'll keep you updated

mbuechse:
@fzakfeld actually we can proceed here... worst case is that the cluster won't be created in the CI and, thus, not tested, which is not worse than the status quo!

fzakfeld:
@mbuechse yes, and we just set 1.35.4 to supported :-)

fzakfeld (approved):

mbuechse:
Sounds great! Thanks!

PR #1221, opened 2026-06-17
Scrap approval of test results, introduce color scale instead

### LINKED ISSUE #1207: Scrap approval of test results, introduce third state (yellow)
For fear of false positives, we invented a system where test results would have to be approved by an auditor.

This system is too labor-intensive, and has also been confusing many people and therefore hurting transparency.

Therefore it shall be scrapped. In its stead, we shall work with three states:

- fail
- indeterminate (or inconclusive, or partial, or ...)
- pass

We can still have two views: one that shows 'indeterminate' as green (because the cloud hasn't failed any required test; maybe show an additional asterisk), and one that shows 'indeterminate' as yellow (because the cloud hasn't passed all required tests).

The difference compared to today is as follows:

- our current views show different test results (one being restricted to approved results)
- the two future views would show exactly the same test results (but with slightly different colors).

To be fully precise, there shall actually be four states:

- fail (some required test has failed)
- miss (some required test is missing)
- indeterminate (some required test did not produce a result)
- pass (all required tests have passed)

For the end-user, the former two would be RED, the latter two would be GREEN (where indeterminate could get an additional asterisk or something).

To the engineer, the first one would be RED, the middle two would be YELLOW, and the final one would be GREEN.

mbuechse:
Show failed (some required testcase failed) as red, missing (some required testcase missing) as yellow, inconclusive (some required testcase aborted) as green with asterisk, passed (all required testcases passed) as green.

The reduces the probability of confusing people with false positives, while no longer requiring (unrealistically) labor-intensive approvals. Besides, the two views 'table_full' vs 'table' now only differ in that the former shows draft versions; similarly for 'details[_full]'.

Further work remains:

- remove approval column from database
- remove approval-related code from sql.py
- remove endpoint for POSTing /results
- simplify, consolidate, isolate business logic
- etc.

We're doing this incrementally... for now, let's see whether this works at all.

### DIFF
--- a/Tests/scs_cert_lib.py
+++ b/Tests/scs_cert_lib.py
@@ -84,7 +84,8 @@ def _resolve_spec(spec: dict):
     # step 4. resolve references
     # step 4a. resolve references to modules in includes
     # in this step, we also normalize the include form
-    for version in spec['versions'].values():
+    for idx, version in enumerate(spec['versions'].values()):
+        version['_idx'] = idx
         version['include'] = [
             {'module': module_lookup[inc], 'parameters': {}} if isinstance(inc, str) else
             {'module': module_lookup[inc['ref']], 'parameters': inc.get('parameters', {})}
@@ -205,8 +206,8 @@ def eval_buckets(results, testcase_ids) -> dict:
 
 def evaluate(results, testcase_ids) -> int:
     """returns overall result"""
-    return min([
-        # here, we treat None (MISSING) as 0 (ABORT)
-        results.get(testcase_id, {}).get('result') or 0
-        for testcase_id in testcase_ids
-    ], default=0)
+    buckets = eval_buckets(results, testcase_ids)
+    for value in (-1, None, 0):
+        if buckets[value]:
+            return value
+    return 1
--- a/compliance-monitor/monitor.py
+++ b/compliance-monitor/monitor.py
@@ -257,6 +257,7 @@ def _evaluate_version(version, scope_results):
         for tname, tc_ids in version['targets'].items()
     }
     return {
+        '_idx': version['_idx'],
         'result': target_results['main']['result'],
         'targets': target_results,
         'tc_target': version['tc_target'],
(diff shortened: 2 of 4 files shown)


mbuechse:
I took the liberty to deploy this to the compliance monitor (after I tested it locally, of course). I found a defect, that I resolved. Now it should be good. Have a look!

- https://compliance.sovereignit.cloud/page/table_full
- https://compliance.sovereignit.cloud/page/table

BTW the views use the same color scale; the only difference is that the former shows draft versions.

This is the color scale:

```python
COLOR_MAP = {
    -1: '🛑',  # fail
    None: '🟧',  # missing
    0: '✅*',  # inconclusive
    1: '✅',  # pass
}
```

I think the asterisk still needs to be explained, and it collides with the asterisk for draft versions. I'm sure we will get this ironed out.

mbuechse:
So, right now, with scs2, we have three testcases that couldn't be tested properly:
 
Screenshot from 2026-06-17 20-49-49

(the virtual machine didn't come up in time)

Which means: we don't know whether these testcases are actually satisfied or not.

Therefore, we now get the asterisk in the summary:

Screenshot from 2026-06-17 20-49-55

mbuechse:
Note: the asterisk for "draft" can probably just be dropped, because

a) we have the convention to call the draft "next", which is quite telling already,
b) the draft versions will only be shown in the "expert" view.

The double dagger for "deprecated" can probably be removed as well, because we have a convention to not show deprecated versions.

What remains is to put the legend of symbols at the top, and presto.

mbuechse:
After #1222, some subjects are missing the result for scs-0114-syntax-check, and so we see this:

Screenshot from 2026-06-18 19-41-51

Screenshot from 2026-06-18 19-42-04

mbuechse:
Now the Sonobuoy tests on Noris take so long that they literally time out. Luckily, the reports got sent anyway (apparently, Zuul is kind of gentle with killing the process; I don't know how this works). Notice the *unfinished* test cases in the tables and the corresponding asterisks in the summary at the top

Screenshot from 2026-06-20 20-30-14

jklare (approved):
lgtm

PR #1222, opened 2026-06-17
Fix missing testcase scs-0114-syntax-check

mbuechse:
The only requirement of the volume type standard was missing.


mbuechse:
@depressiveRobot  Yes, all subjects!

depressiveRobot (approved):
Thanks. Then let's merge it :)

PR #1224, opened 2026-06-18
Run at most one kaas job at one time, same for iaas

mbuechse:
We want to run the tests for the k8s branches in sequence to avoid hitting quotas and to conserve resources.

For iaas, we want to avoid that the hourly job interferes with the daily job or (more likely and observed already) the post-merge job.


depressiveRobot (approved):
LGTM :)

PR #1227, opened 2026-06-19
Improve robustness of SCS test suite

mbuechse:
- Save report file before uploading, so it won't be lost in case of a failure.
- If pre-flight checks fail, report testcases as ABORT (instead of not at all).


depressiveRobot (approved):
LGTM

PR #1228, opened 2026-06-20
Make version-policy-check more robust

mbuechse:
- Remove unused k8s-version-recency
- Move kaas/k8s-version-policy to kaas/scs_0210_version_policy (in line with iaas naming)
- Make version-policy-check report ABORT result


mbuechse:
I think after this, we can finally close #889.

depressiveRobot (approved):
Pair reviewed ✅

PR #1230, opened 2026-06-22
Improve Gardener setup

mbuechse:
- allow omitting machine_image_version
- allow specifying a volume via volume_size_gib
- account for state 'Pending' in plugin_gardener.py

### DIFF
--- a/playbooks/.config/scs/shoot.yaml
+++ b/playbooks/.config/scs/shoot.yaml
@@ -23,13 +23,19 @@ spec:
         machine:
           type: {{ machine_type }}
           image:
-            name: {{ machine_image_name }}
-            version: {{ machine_image_version }}
+            name: {{ machine_image_name }}{%
+if machine_image_version | default('') %}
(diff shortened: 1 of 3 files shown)


playbooks/.config/scs/shoot.yaml
```diff
@@ -23,13 +23,19 @@ spec:
         machine:
           type: {{ machine_type }}
           image:
-            name: {{ machine_image_name }}
-            version: {{ machine_image_version }}
+            name: {{ machine_image_name }}{%
+if machine_image_version | default('') %}
+            version: {{ machine_image_version }}{%
```
fzakfeld, line 28: Not sure if this works but if someone else can confirm I'm happy with it

fzakfeld (approved):

playbooks/.config/scs/shoot.yaml
mbuechse: I can confirm that it works. Cluster will be created and tests can be run (they are still running though)

PR #1232, opened 2026-06-25
Add noris nsc-iaas scs-compliance check

dhoessl:
* adds subject noris-iaas with api-keys and keys to compliance-monitor/bootstrap.yaml
* adds noris-iaas to compliance-monitor/templates/overview.md.j2


mbuechse:
@dhoessl Could you please rebase on current main so the pipelines can run and eventual merge go through more easily?

PR #1199, opened 2026-06-06
Replace yaml scopes by regular standards

mbuechse:
Resolves #1198 

This is meant as a preview of a quite radical change. It's possible that multiple PRs would be needed to be in line with extant regulations.

Issue #1081, opened 2026-01-30
[Feature Request] Introduce testcase-specific reporting


mbuechse:
Currently, our report assigns to every testcase a result (PASS/FAIL/ABORT). Everything else is buried in blanket stdout/stderr capture, which has its own internal structure, but that is only apparent if you see all the intervening debug messages, which are redacted on the compliance monitor. It's not great UX.

Also stderr should not be relevant to anyone except test developers! That need not even be part of the UI (except for the full report view).

- [ ] change all scripts so that relevant output is done on stdout, prefixed by testcase-id
- [ ] make sure stderr only contains debug info and unexpected operational errors
- [ ] hide stderr from CM UI (except in full report view)
- [ ] make testcase-specific stdout messages available in CM UI (no longer calling it stdout)

mbuechse:
**I instrumented the logger** in `Tests/iaas/openstack_test.py` (which contains all iaas testcases) to automatically output for each line the testcase (if the line can be connected to a single testcase). In the following, lines starting with `DEBUG:`, `INFO:` etc. originate from logging, whereas the others are printed on stdout.

**Exhibit A:**

```text
$ ./iaas/openstack_test.py -c artcodix-ro scs-0100-semantics-check scs-0100-syntax-check
DEBUG: ** scs-0100-semantics-check
INFO: scs-0100-semantics-check: Non-standard disk size for 'SCS-2V-4-2s': 2 not in (5, 10, 20, 50, 100, 200, ...)
INFO: scs-0100-semantics-check: Non-standard disk size for 'SCS-1V-2-1s': 1 not in (5, 10, 20, 50, 100, 200, ...)
scs-0100-semantics-check: PASS
DEBUG: ** scs-0100-syntax-check
scs-0100-syntax-check: error parsing SCS-16V-48-4GPU: Extra characters: GPU
scs-0100-syntax-check: FAIL
```

Here one can now see that the two INFO lines are in fact related to the testcase mentioned in the DEBUG line above (the instrumented logger tells us as much by prepending the name of the testcase to the log message).

**Exhibit B:**

```text
$ ./iaas/openstack_test.py -c pco-prod1 scs-0100-syntax-check scs-0100-semantics-check  
DEBUG: ** scs-0100-syntax-check
WARNING: [scs_flavors] Name is merely tolerated v1: SCS-4V:8:100
WARNING: [scs_flavors] Name is merely tolerated v1: SCS-4V:32:100
[...]

Issue #1069, opened 2026-01-27
[Feature Request] Specify and implement redacted reports


mbuechse:
Applicants for SCS-0004 integrator certification have to prove that they have set up / have been operating compliant cloud environments. These environments can be quite sensitive, and their owners could have reasonable reservations regarding the sharing of our report files, because they might contain sensitive information.

- [ ] (loosely) specify what MUST, SHOULD, MAY, MUST NOT to be contained in redacted report vs. a full report
- [ ] similarly for Tempest!
- [ ] add a command-line switch to `scs-compliance-test.py` that requests for the report to be redacted
- [ ] add a similar option to `scs-test-runner.py` (probably as a subject-specific option in `config.toml`)
- [ ] provide an official means of redacting tempest logs/output

mbuechse:
I think if a testcase needs to add messages to the report besides the result (PASS/FAIL/ABORT), then these messages belong in `stdout` (not `stderr`). It is part of the normal operation, the normal output of the testcase.

The main difficulty here is programming-wise: we have many testcases that are implemented by the same function with some parameter (e.g., test presence of some image, test presence of some flavor, test presence of some service). The function doesn't know the name of the testcase. So either this name has to be supplied to the function as a parameter, or the function cannot directly print to stdout. Either way, the general type of a testcase function becomes more complicated...

mbuechse:
@fkr When one of the image-metadata testcases fails, it produces a list of all public images that need correcting for the testcase to pass (for instance: "property X missing in images Y, Z"). Would this list (Y, Z) already be sensitive information? I suppose it could be so construed?

PR #1237, opened 2026-07-01
Overhaul test scripts in a major way

mbuechse:
This is a follow-up and accompaniment to #1199.

Streamline spec yaml, report yaml, test scripts, ...

Vastly improve report view in compliance monitor.


mbuechse:
This could also address #1069 and #1081.

PR #1239, opened 2026-07-03
Hide syseleven from overview table for the time being

mbuechse:


mbuechse:
@depressiveRobot tested locally -- works

berendt:
Is there a reason why this is done?

mbuechse:
@berendt 
Table entries that are permanently red are quite irritating and cause increased mental load for Marvin and me. We have decided to ask the operators in question whether they object to having their entries removed. Some of them replied that they now want to push for becoming green again, others replied that they want to become green again, but it would take them a lot of time, and therefore they agreed with the removal. Mind you: the entries in question are for testing purposes anyway and not related to any certificates.

mbuechse:
Not going ahead with this after further internal discussion.

PR #1240, opened 2026-07-06
Update hints for certification fees

depressiveRobot:
During conversations, it repeatedly comes up that interested parties are unaware that certification fees may apply.

This PR therefore adds a hint in the standards SCS-0004 and SCS-0007 regarding fees that may apply. The respective supplements provide more details on the fees. If there are any changes to the fees, this will make it easier for us by just updating the supplements.

Please (auto)merge if everything is fine.

### DIFF
--- a/Standards/scs-0004-v1-achieving-certification.md
+++ b/Standards/scs-0004-v1-achieving-certification.md
@@ -18,7 +18,7 @@ As operator, I want to obtain a certificate with the scope SCS-compatible IaaS o
 
 ## Regulations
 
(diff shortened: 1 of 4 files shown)


garloff (approved):
Sounds good to me and states the facts more clearly than before.

One observation:
- In two places, we clearly say that the fees are paid either by membership (of the right level) or they are paid without becoming a member.
- In two other places we just mention that fees may apply.

This may be intentional, as the level of precision may be more appropriate at one place than on another one ... Suggesting the author to double-check that this was intentional. (Sorry, I did not invest the time to look at enough context to develop a good feeling for this.)

PR #1241, opened 2026-07-07
Certification status in compliance overview

### LINKED ISSUE #1059: [Feature Request] Restructure CM overview page
We're talking these two (very similar) pages:

- verfied overview: https://compliance.sovereignit.cloud/page/table
- unverified overview: https://compliance.sovereignit.cloud/page/table_full

Instead of one big table, make three sections (or three tables):

1. officially certified subjects
2. subjects not officially certified but expected to be by-and-large compliant
3. experimental subjects where compliance is expected to be a bit shaky

We need to find well-sounding headings for these categories... The third section could be hidden in the verified overview.

depressiveRobot:
This PR adds a new column "Certified" with a short explanation to the [compliant cloud environments overview](https://docs.scs.community/standards/certification/overview#compliant-cloud-environments) (aka Compliance Monitor). If an environment is certified, it is also linked to the respective certification status page.

This was discussed in the last Forum SCS-Standards call.

Please (auto)merge if successfully reviewed.

### DIFF
--- a/compliance-monitor/templates/overview.md.j2
+++ b/compliance-monitor/templates/overview.md.j2
@@ -3,66 +3,81 @@ we could of course iterate over results etc., but hardcode the table (except the
 for the time being to have the highest degree of control
 -#}
 
-The following environments are regularly checked for SCS compliance.
+The following tables show environments that are regularly checked for SCS compliance. A distinction is made between certified and non-certified environments. Certified environments are officially certified by the SCS certification assessment body and must maintain compliance throughout the validity period of the certification. Non-certified environments are listed on a voluntary basis to demonstrate their compliance for potential future certification or internal purposes.
 
-Officially certified environments are marked with an X in the “Certified” column and are linked to the respective certification status page.
+The columns "SCS-compatible IaaS"/"SCS-compatible KaaS" show the last compliance status of the environment for the tested scope version. The scope version is suffixed by a symbol depending on the state of the scope: * for _draft_, † for _warn_ (soon to be deprecated), and †† for _deprecated_. Clicking on the compliance status directs to the detailed testing report for the environment.
 
-The scope version numbers in the columns "SCS-compatible IaaS"/"SCS-compatible KaaS" are suffixed by a symbol depending on the state of the scope: * for _draft_, † for _warn_ (soon to be deprecated), and †† for _deprecated_.
+### Certified environments
 
-### SCS-compatible IaaS
+These environments are officially certified by the SCS certification assessment body according to the standard [SCS-0004](https://docs.scs.community/standards/global/scs-0004). Clicking on the name in the first column directs to the official certification status page.
 
-This is a list of IaaS clouds that we test on a nightly basis against the certificate scope [SCS-compatible IaaS](https://docs.scs.community/standards/scs-compatible-iaas/).
+#### Certified SCS-compatible IaaS
+
+This is a list of officially certified IaaS environments that are tested regularly against the certificate scope [SCS-compatible IaaS](https://docs.scs.community/standards/scs-compatible-iaas/).
 
 {% set iaas = 'scs-compatible-iaas' -%}
-| Name  | Description  | Operator  | Certified  | SCS-compatible IaaS  | HealthMon  |
-|-------|--------------|-----------|------------|----------------------|:----------:|
-| [scs2](https://docs.scs.community/community/cloud-resources/plusserver-gx-scs) | Dev/Test/Demo environment (2nd gen) provided for SCS & GAIA-X context (1 SCS region) | plusserver GmbH | |
-{#- #} [{{ results | pick(iaas, 'scs2') | summary }}]({{ detail_url('scs2', iaas) }}) {# -#}
-| [HM](https://health.prod1.plusserver.sovereignit.cloud:3000/d/9ltTEmlnk/openstack-health-monitor2?orgId=1&refresh=5m&var-mycmd=All&var-mymethod=All&var-mywait=All&var-mybench=All&var-mycloud=gx-scs2) |
-| [aov.cloud](https://www.aov.de/) | Community cloud for customers (1 SCS region) | aov IT.Services GmbH | |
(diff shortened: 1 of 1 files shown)


compliance-monitor/templates/overview.md.j2
```diff
@@ -3,43 +3,47 @@ we could of course iterate over results etc., but hardcode the table (except the
 for the time being to have the highest degree of control
 -#}
 
-Version numbers are suffixed by a symbol depending on state: * for _draft_, † for _warn_ (soon to be deprecated), and †† for _deprecated_.
+The following environments are regularly checked for SCS compliance.
+
+Officially certified environments are marked with an X in the “Certified” column and are linked to the respective certification status page.
```
mbuechse: I'd very much prefer to see them in their own table, or at least to group the rows.

fzakfeld: Not sure about this, we already have 2 tables on there.

depressiveRobot: This was my thought too. A further separation would mean four different tables. And with more certifications (e.g. Paas), it would be six or more tables. That's why I decided on a column.

mbuechse: From my POV, there is a very clear qualitative difference between those offerings that are certified and those that we merely watch for the fun of it.

There are so many options here:

- keep each table mixed, but at least group the rows (as I already proposed)
- move non-certified entries to an entirely separate page
- put all layers into one table (again with rows grouped), and then make separate tables for certified vs. fun

To me, each of these options appears more favorable than what is currently in this PR.

janiskemper: I agree here with @mbuechse with one aspect: I think we don't yet make it clear enough what the difference is. What is even the use case of being in the compliance monitor, and why is that not the same thing as being certified? That's something we have to explain for people having no knowledge of SCS.

Different tables can support this by giving us the chance to introduce the respective next table with 1-2 sentences. 

We can potentially also achieve that in a different way, so I'm not gonna veto the idea of keeping 2 tables. 

I just vote for making the difference between certification and no certification clearer.

compliance-monitor/templates/overview.md.j2
```diff
@@ -51,14 +55,14 @@ This is a list of IaaS clouds that we test on a nightly basis against the certif
...
 {#- #} [{{ results | pick(kaas, 'group-noris') | summary }}]({{ detail_url('group-noris', kaas) }}) {# -#}
 |
-| [ScaleUp Open Cloud](https://www.scaleuptech.com/cloud-hosting/) | Public KaaS cloud based on Gardener | ScaleUp Technologies GmbH & Co. KG | {# #}
+| [ScaleUp Open Cloud](https://www.scaleuptech.com/cloud-hosting/) | Public KaaS cloud based on Gardener (1 SCS configuration) | ScaleUp Technologies GmbH & Co. KG | [X](https://sovereigncloudstack.org/en/certified-solution/scaleup-open-cloud/)
 {#- #} [{{ results | pick(kaas, 'group-scaleup') | summary }}]({{ detail_url('group-scaleup', kaas) }}) {# -#}
 |
-| [Syself Autopilot](https://syself.com/platform) | KaaS offering based on ClusterStacks | Syself GmbH | {# #}
+| [Syself Autopilot](https://syself.com/platform) | KaaS offering based on ClusterStacks (1 SCS configuration) | Syself GmbH | [X](https://sovereigncloudstack.org/en/certified-solution/syself-gmbh/)
```
mbuechse: This is a redirect to the Syself page, whereas the corresponding URL for Scaleup is not a redirect, but a page on sovereigncloudstack.org. I find this a bit confusing.

I think the 'touch target' is a bit small for a link that is just labelled "X". I would prefer to put this link on the name in the first column.

compliance-monitor/templates/overview.md.j2
```diff
@@ -3,43 +3,47 @@ we could of course iterate over results etc., but hardcode the table (except the
 for the time being to have the highest degree of control
 -#}
 
-Version numbers are suffixed by a symbol depending on state: * for _draft_, † for _warn_ (soon to be deprecated), and †† for _deprecated_.
+The following environments are regularly checked for SCS compliance.
```
mbuechse: This is most probably true. What's stated further below, under each subheading, is most probably false: "This is a list of [...] clouds that **we test on a nightly basis**".

compliance-monitor/templates/overview.md.j2
mbuechse: Putting the non-certified environments on a separate page would be logical, because this page is embedded into the docs page under the heading "Certification" (and the subheading "Compliant cloud environments" that is put to shame by the red entries underneath)

fzakfeld: I think with sub-headings it would be a good compromise. E.g. certified first, "not certified but still checked" second

depressiveRobot: I have implemented multiple tables now with more explanations, see e51d043

compliance-monitor/templates/overview.md.j2
depressiveRobot: I "fixed" the touch target by using the "name" column to link to the official certification status pages.

For Syself there is currently no official certification status page. That's why it's redirecting. However, the link is valid. The same applies for REGIO.cloud. The status pages will be fixed on sovereigncloudstack.org soon.

compliance-monitor/templates/overview.md.j2
depressiveRobot: I have changed the explanations accordingly.

mbuechse:
> This is how it looks now:
> Screenshot 2026-07-14 at 17-34-32 SCS compliance overview

Looking good! Two things are bothering me though:

1. to my eye, the texts are too verbose and redundant. If we need so much prose, then it's still too complicated to begin with.
2. in the KaaS section, the columns are off (this can bee seen in the screenshot): the result column is empty, and the result is stuck to the end of the operator column.

The latter should be easy to fix, and the former, well, I can propose a less verbose and redundant version, or we find ways of simplifying the whole thing more generally.

mbuechse (approved):

mbuechse:
This is what it looks like now:

Screenshot from 2026-07-21 21-27-08

PR #1244, opened 2026-07-10
feat: add t8s kaas offering

cwrau:


cwrau:
I hope I implemented this correctly, but at least my preliminary tests were successful.

mbuechse:
Thanks @cwrau. At first glance, it looks promising! Will have to read carefully and write up a review, but due to vacation, it could take until July 22nd or so...

mbuechse:
Question: you wrote that it's based on Gardener, yet you don't seem to use the Gardener API, but some Helm stuff (I still need to read more carefully). Is Gardener not exposed to your customers? If that's true, we should not mention Gardener.

cwrau:
> Question: you wrote that it's based on Gardener, yet you don't seem to use the Gardener API, but some Helm stuff (I still need to read more carefully). Is Gardener not exposed to your customers? If that's true, we should not mention Gardener.

Däh, where did you read that? 😅
I looked at `git log -p` again and I don't think I mentioned that anywhere, as it's not true anyways 😅

compliance-monitor/templates/overview.md.j2
```diff
@@ -62,3 +62,6 @@ This is a list of KaaS clouds that we test on a nightly basis against the certif
 | [Syself Autopilot](https://syself.com/platform) | KaaS offering based on ClusterStacks | Syself GmbH | {# #}
 {#- #} [{{ results | pick(kaas, 'group-syself') | summary }}]({{ detail_url('group-syself', kaas) }}) {# -#}
 |
+| [t8s](https://teuto.net/produkte/t8s/) | Public KaaS cloud based on Gardener | teuto.net Netzdienste GmbH | {# #}
```
mbuechse: Here it says Gardener

cwrau: Oh, apparently my search was case-sensitive...

Will fix that!

berendt: A Freudian slip ;)

cwrau:
Ehlo, just wanted to ask if I need to adjust anything?

mbuechse:
@cwrau 

No, I'm afraid you just have to wait

> Will have to read carefully and write up a review, but due to vacation, it could take until July 22nd or so...

playbooks/clusters.yaml.j2
```diff
@@ -209,6 +209,24 @@ clusters:
         num_worker_nodes: 3
       secrets:
         token: "{{ clouds_conf.syself_token }}"
+  teuto-1.35:
+    kind: t8s
+    config:
+      kubernetesVersion: '1.35'
+      version_patch: 2
```
mbuechse, line 216: Ideally, the latest patch version should be selected automatically. For instance, is there a way to enumerate available patch versions?

Tests/kaas/plugin/plugin_t8s.py
```diff
@@ -0,0 +1,196 @@
...
+    def _build_helmrelease(self) -> dict[str, Any]:
+        return {
+            "apiVersion": f"{HR_GROUP}/{HR_VERSION}",
+            "kind": "HelmRelease",
+            "metadata": {"name": HR_NAME, "namespace": HR_NAMESPACE},
+            "spec": {
+                "chartRef": HR_CHART_REF,
+                "driftDetection": {"mode": "enabled"},
+                "interval": "1m",
+                "values": {**HR_VALUES_FIXED, "version": self.k8s_version},
```
mbuechse: Minor remark: Since this method only uses this one instance variable, I'm wondering whether it's better turned into a function that takes `k8s_version` as a parameter.

Tests/kaas/plugin/plugin_t8s.py
```diff
@@ -0,0 +1,196 @@
...
+        logger.debug(f"writing {path}")
+        with open(path, "wb") as f:
+            f.write(data)
+
+    def create_cluster(self) -> None:
+        with ApiClient(self.client_config) as api_client:
+            core_api = CoreV1Api(api_client)
+            co_api = CustomObjectsApi(api_client)
+            self._apply_helmrelease(api_client)
+            self._write_kubeconfig(self._wait_for_kubeconfig_secret(core_api))
```
mbuechse: I would prefer to see `self._wait_for_kubeconfig_secret` on a dedicated line, because this is where most of the time will be spent:


```suggestion
            kubeconfig = self._wait_for_kubeconfig_secret(core_api)
            self._write_kubeconfig(kubeconfig)
```

Tests/kaas/plugin/plugin_t8s.py
```diff
@@ -0,0 +1,196 @@
...
+import os.path
+import time
+from typing import Any, cast
+
+from jinja2 import Environment, Template
+from kubernetes.client import ApiClient, ApiException, Configuration, CoreV1Api, CustomObjectsApi, V1Secret
+import yaml
+
+from interface import KubernetesClusterPlugin
+import cs_helper as _csh
```
mbuechse: This is imported for `setup_client_config` only (note that `_csh.Configuration` is identical to `Configuration` from a few lines further up).

This function, `setup_client_config`, is currently duplicated in `cs_helper` and `gardener_helper`, and I think it should be moved into a separate module, say, `k8s_helper`. Either that, or it should be duplicated once more in a t8s-specific helper, which would be a bit crazy :)

Tests/kaas/plugin/plugin_t8s.py
```diff
@@ -0,0 +1,196 @@
...
+        )
+        logger.debug(f"HelmRelease {HR_NAME} applied")
+
+    def _get_kubeconfig_from_secret(self, core_api: CoreV1Api) -> bytes:
+        secret = cast(V1Secret, core_api.read_namespaced_secret(HR_KUBECONFIG_SECRET, HR_NAMESPACE))
+        data: dict[str, str] = secret.data or {}
+        if "value" in data:
+            return base64.standard_b64decode(data["value"].encode())
+        raise RuntimeError(
+            f"kubeconfig secret {HR_KUBECONFIG_SECRET} has unexpected keys: {list(data.keys())}"
```
mbuechse: First (a nitpick), this is un-Pythonic; instead of `list(data.keys())`, one would simply say `list(data)`.

Second, it seems a bit illogical to me: this Exception is thrown if and only if the key `value` is missing, but the error messages is suggesting that there are extraneous keys.

mbuechse (commented):
Looking quite good! Thank you for contributing this!

I am missing k8s 1.34, and I have a few remarks, but small fry only.

Personally, I'm not into the excessive typing and unit-testing, where the former makes the code harder to read and the latter increases the codebase for marginal gain (because, in my experience so far with these plugins, they won't be changed once they work, and the changes we do would not be covered by the tests). However, this is mainly my personal opinion, and if you prefer the typing and unit-testing, then so be it.

playbooks/clusters.yaml.j2
cwrau: Sadly no, to do that we'd have to look at our TeutonetesClouds status and there are internal settings/infos on that object.

mbuechse: We may be able to live with this restriction for a while, but in the long run, it won't be fun bumping the patch version by hand every time the testcase `version-policy-check` is failed...

Tests/kaas/plugin/plugin_t8s.py
cwrau: This would "break" consistency to the other _build_* and _apply_* methods. But I can change it if you want

mbuechse: You're right. But now that is even more concerning: can we not provision multiple clusters at the same time? At the very least two, one for each branch?

cwrau:
> Looking quite good! Thank you for contributing this!
> 
> I am missing k8s 1.34, and I have a few remarks, but small fry only.

Ah, we don't have anyone on 1.34 anymore, so I didn't think about that, is that necessary? 

> Personally, I'm not into the excessive typing and unit-testing, where the former makes the code harder to read and the latter increases the codebase for marginal gain (because, in my experience so far with these plugins, they won't be changed once they work, and the changes we do would not be covered by the tests). However, this is mainly my personal opinion, and if you prefer the typing and unit-testing, then so be it.

Typing is definitely for my own sanity and maintainability, but if that isn't this repos style I can drop it. 

Same goes for the tests, if those testable changes aren't really happening then OK 😁

Tests/kaas/plugin/plugin_t8s.py
cwrau: That wasn't planned, no. 

K8s RBAC isn't that flexible, but if we have some form of deterministic order we could allow 3; each with an indexed suffix (`-1`, `-2` and `-3`)

playbooks/clusters.yaml.j2
cwrau: Oh, I didn't know the patch version was part of the certification, I assumed only the minor was required. 

But the patch version will be removed when our customer portal allows machine access, then we'd have no choice of the patch version (or any version for that matter) in any case. (we'd have update schedules just for SCS that have the old versions 😅)

Tests/kaas/plugin/plugin_t8s.py
mbuechse: We can live with one, I suppose. The CI pipeline has been equipped with a semaphore so only one k8s branch will be tested at a time, so we should be safe.

From a programming standpoint, I think the class doesn't make a lot of sense because many of its methods don't even use `self`. So these methods could be turned into simple functions or `staticmethod` or `classmethod`. However, I would rather suggest we pretend that certain quantities are not actually constants and introduce instance variables for them, even if we only ever instantiate them with the same value. Like you just said, even if these values need to be hard-coded/prepared on the server side, several of them could exist, and then this would make sense.

mbuechse:
Like I said, we can keep the typing and the tests. I don't want to impose my style upon contributors, and it's a welcome chance to prove me wrong. :smile: 

So long as 1.34 is officially supported (this would be until Oct 27, 2026), we would need to have it, but Teutonet could also declare that you don't want to be certified before that date :wink:

playbooks/clusters.yaml.j2
mbuechse: For SCS, we would like to be able to select the k8s branch (so major.minor), and the patch version can just be whatever is most recent. From what you're writing it doesn't sound so far off.

cwrau:
> So long as 1.34 is officially supported (this would be until Oct 27, 2026), we would need to have it, but Teutonet could also declare that you don't want to be certified before that date 😉

Probably, we don't have 34 anymore 😅

Tests/kaas/plugin/plugin_t8s.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def __init__(self, config: dict[str, Any], basepath: str = ".", cwd: str = ".") -> None:
+        self.basepath = basepath
+        self.cwd = cwd
+        self.env = Environment()
+
+        # All of these are mandated by the ValidatingAdmissionPolicy on the management cluster
+        # and cannot be changed without updating both the VAP and the RBAC. They are instance
+        # (rather than module-level) attributes because, in principle, different certification
+        # targets could be configured with different values here.
```
mbuechse, line 64: It would be good to pull the relevant values from the config dict as well.

compliance-monitor/templates/overview.md.j2
```diff
@@ -63,3 +63,13 @@ These environments are not officially certified. Operators voluntarily list them
 | [syseleven](https://www.syseleven.de/en/products-services/openstack-cloud/) | Public OpenStack Cloud (2 SCS regions) | SysEleven GmbH |
 {#- #} [{{ results | pick(iaas, 'group-syseleven') | summary }}]({{ detail_url('group-syseleven', iaas) }}) {# -#}
 | (soon) |
+
+### SCS-compatible KaaS
+
+This is a list of KaaS clouds that we test on a nightly basis against the certificate scope [SCS-compatible KaaS](https://docs.scs.community/standards/scs-compatible-kaas/).
```
mbuechse:
```suggestion
```

We have three sections now:

- certified IaaS
- certified KaaS
- un-certified environments (IaaS, KaaS)

mbuechse:
The latest commit is missing the sign-off, but this can be added with `git rebase --signoff HEAD~1`. Note that due to the rewritten history you then have to `git push -f`.

mbuechse (approved):
Let's fetz

PR #1245, opened 2026-07-16
SCS-0128: Remove testcase that needs instance creation with admin password

toothstone:
This is not available on all clouds due to reasonable security concerns.

AFAIK it is not a mandatory feature to have this enabled, so we shouldn't test for it.


toothstone:
@depressiveRobot I assume you've already greenlit certification requests with 
```text
{0} tempest.api.compute.servers.test_servers.ServersTestJSON.test_create_server_with_admin_password ... SKIPPED: Instance password not available.
```
in their Tempest logs?

depressiveRobot (approved):
Thank you, LGTM

PR #1247, opened 2026-07-16
SCS-0128: Drop user account lockout test

toothstone:
It's not standardized how CSPs are expected to handle this, so we shouldn't test for it.

### DIFF
--- a/Standards/scs-0128-v1-e2e-testing.md
+++ b/Standards/scs-0128-v1-e2e-testing.md
@@ -28,3 +28,18 @@ The end-to-end testing is performed using [Tempest](https://docs.openstack.org/t
 
 The required tests are listed in
 [Tests/iaas/scs_0128_e2e_testing/tempest-tests-non-admin.lst](https://raw.githubusercontent.com/SovereignCloudStack/standards/refs/heads/main/Tests/iaas/scs_0128_e2e_testing/tempest-tests-non-admin.lst).
+
+Tempest MUST NOT report any _failed_ test cases.
+
+## Rationale
+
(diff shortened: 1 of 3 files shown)


toothstone:
@mbuechse AFAIK this found consensus to remove in the last SIG, would you give your approval and merge please?

mbuechse:
Can we add a comment to the standard text about this?

toothstone:
@mbuechse I've added a note explaining the removal to the implementation notes, is this what you had in mind?

mbuechse:
@toothstone I don't quite understand why this was put in the testing notes... The list is normative, and we should motivate this list in the standard, I would say.

toothstone:
> @toothstone I don't quite understand why this was put in the testing notes... The list is normative, and we should motivate this list in the standard, I would say.

IMHO this kind of history/change log should be kept outside the standard text, since it is irrelevant to everyone using or implementing it. I've not seen this in any other SCS standard, either.

mbuechse:
Other standards do have that. I can look up examples later. But more importantly, other standards motivate their regulation, be it interlaced within the normative text or in a motivation section. We cannot just toss a naked list of testcasea over to people.

toothstone:
> We cannot just toss a naked list of testcasea over to people.

This PR is not what adds that list, though. It feels quite lopsided that removing a single test case now needs more explanation and documentation than all of OPC/SCS-0128 seems to have ever needed...

Feel free to drop a suggestion where to move the explanation, I'm happy to accept it.

mbuechse:
We don't need to compensate for everything that was neglected previously. The only thing I want is that we document why we removed the test case, and that we do it in the main document. The main document is a draft for a reason: this needs to be expanded upon in a big way. For the time being, we just don't want to forget this change.

mbuechse (approved):

PR #1248, opened 2026-07-16
SCS-0128: Drop broken tests

### LINKED ISSUE #990: Modify OpenStack Powered Compute to make it work
Apparently, some test cases are not suitable in their present state. This issue is meant as a place for collecting the findings.

- `test_default_domain_exists` seems to need admin access (either that or the test must be configured in another way?)
- `test_rebuild_server` has issues with a changed SSH cert (which is to be expected, however)

Big thanks to @fzakfeld for finding these things!

Details:

## `test_default_domain_exists`

- required since June 2022
- "the test is wrong from any possible perspective. It claims to test "default" domain existence which is in public cloud  a) not guaranteed to exist b) no user is going to be able to list other domains. I really believe it was a mistake to add it and very soon the whole interop died so nobody noticed"
- you can, however, change the `domain_id` that you test through config (`identity.default_domain_id`) to "show" details of the domain you use for testing

Maybe the latter item is a solution?

## `test_rebuild_server`

- "[this test] will rebuild the server and then complain that the SSH host key is not the same as before. This is to be expected with pretty much any cloud image, as they generate the keys on first boot. I think this is due to paramiko adding the key at the first time - see https://opendev.org/openstack/tempest/src/commit/d12ddc31cb403ba1c7534e43bfa5650b6c131f4b/tempest/lib/common/ssh.py#L119"
- patching tempest to use `paramiko.WarningPolicy()` instead will make the test work

toothstone:
Resolves #990 - remove tests:
- default domain exists - out of scope
- rebuild server - expects persistent SSH keys, which is not how modern cloud images work


toothstone:
SIG decided that faking the `default_domain_id` is the appropriate way to pass the first test case, the second one was most likely patched to work upstream in the meantime.

PR #1249, opened 2026-07-16
SCS-0128: Append test case qualifiers

### LINKED ISSUE #1165: [BUG] Incorrect Tempest test identifiers in test lists
## Summary

Some Tempest tests currently defined in [our test lists](https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/scs_0128_e2e_testing/tempest-tests-non-admin.lst) are not being discovered correctly. This is due to missing the required `[,negative]` suffix in their identifiers. Probably there are some more incorrect identifiers.

## Problem Description

When defining specific Tempest tests, simply using the class path is often insufficient for negative tests. If the `[,negative]` suffix (including the test ID) is missing, the tempest run command fails to match the test, resulting in it being ignored during execution.

### Example
For instance, the test `test_update_non_existent_security_group` is currently not correctly identified unless the full string is provided.

Incorrect/Incomplete reference: 
```
tempest.api.compute.security_groups.test_security_groups_negative.SecurityGroupsNegativeTestJSON.test_update_non_existent_security_group
```

Correct reference:
 ```
tempest.api.compute.security_groups.test_security_groups_negative.SecurityGroupsNegativeTestJSON.test_update_non_existent_security_group[id-27edee9c-873d-4da6-a68a-3c256efebe8f,negative]
```

The correct identifiers can be verified by running the following command within the tempest container/environment:

```bash
tempest run --list-tests | grep test_update_non_existent_security_group
```

toothstone:
Fixes #1165 - I sorted the [updated test list provided by @depressiveRobot](https://gist.githubusercontent.com/depressiveRobot/1d38cd338a52712d12a67954acd8c626/raw/169311ac1e10c0c694dd0811796b67170e915f13/tempest-tests-non-admin-fixed.lst) and checked it in.

The diff shows no ID changes, but now all qualifiers should be present.


depressiveRobot (approved):
Looks good and thank you for doing this.

PR #1250, opened 2026-07-17
Revert "SCS-0128: Remove testcase that needs instance creation with admin password"

depressiveRobot:
Reverts SovereignCloudStack/standards#1245

I mistakenly merged this too quickly. Most of the certified providers successfully run this check. This should first be discussed in the SIG Std/Cert.


toothstone:
Have you re-created my original PR somewhere? We need to track this somewhere.

mbuechse:
@toothstone What do you mean by 'recreated the PR'? It was never gone, and it's referenced in this PR's description.

toothstone (approved):
Approving to comply with you wish to discuss this in the SIG, I stand by my assessment that this test case should be dropped ;)

I've opened #1253 to re-apply my original PR in case this one here gets merged. In the future I would appreciate it if that wouldn't fall onto me.

toothstone:
Out of curiosity, why does this commit not need to be signed-off?

PR #1251, opened 2026-07-20
Fix delegation problem due to removed wavestack

mbuechse:
Amending a13247bdb2e2764b75b942e52fc4ae9557bfd22f

### DIFF
--- a/Tests/config.toml
+++ b/Tests/config.toml
@@ -30,7 +30,6 @@ subjects = [
     "scaleup-occ2",
     "syseleven-dus2",
     "syseleven-ham1",
-    "wavestack",
 ]
 workers = 16
 


gerbsen (approved):
LGTM, but in my opinion, it shouldn't be the case that an incorrect entry affects the test results of the other entries. I'll open an issue for this.

mbuechse:
> LGTM, but in my opinion, it shouldn't be the case that an incorrect entry affects the test results of the other entries. I'll open an issue for this.

The HTTP request has to report failure, and it shouldn't succeed "half way" (partially). It's an atomic operation. However, we shouldn't send all reports with one HTTP request. We should expend one request per report. This is probably going to be addressed with a major overhaul that's currently in development anyway.

PR #1253, opened 2026-07-20
Re-apply "SCS-0128: Remove admin pass instance creation test"

toothstone:
Recreation of #1245 after it was erroneously merged and had to be reverted in #1250 

Only applicable once #1250 is merged.


toothstone:
There was an agreement with @mbuechse that SCS-0128 will be amended to only require not failing any tests, so skipping tests is allowed.
This means I can just set the tempest feature flag controlling this as `enable_instance_password = false` and always skip the problematic test case, so if the revert goes through it's not a problem.

PR #1255, opened 2026-07-22
Fix broken links reported by the link checker

jklare:
The daily link-validator workflow was failing on a number of broken links across the Standards and Drafts. This PR fixes them, split into commits by category so each change is easy to review.

## Moved / relocated targets (`docs: Update links whose targets have moved`)
Resources that still exist but have a new location:

- **OSS-Health**: `Drafts/OSS-Health.md` was promoted to the decision record `Standards/scs-0008-v1-oss-health.md` (referenced from scs-0212, scs-0218, scs-0401).
- **k8s_version_policy.py**: moved to `Tests/kaas/scs_0210_version_policy/` (scs-0210-w1).
- **Mandatory flavors spec**: renamed to `Tests/iaas/SCS-Spec.MandatoryFlavors.verbose.yaml` (scs-0110).
- **Container registry standard**: Harbor draft referenced the old singular filename; corrected to `scs-0212-v1-requirements-for-container-registries.md`.
- **STIG Viewer** finding V-242387: `/stig/` → `/stigs/` (scs-0217).
- **dNation Thanos SVG**: moved into `docs/images/` (scs-0403).

## Dead links (`docs: Replace dead links with live equivalents`)
Resources gone with no redirect, repointed to the closest live equivalent:

- **cState demo** `cstate.netlify.app` (404) → upstream repo `github.com/cstate/cstate` (scs-0400).
- **Patrole "RBAC Overview"**: Patrole is a retired OpenStack project; cite the current Keystone identity-API-protection docs instead (scs-0302).

## False positive (`docs: Fix footnote so link checker parses RBAC link`)
The Neutron RBAC link in scs-0126 was valid, but a footnote placing the link directly after the `[^rbac]:` label confused the checker's parser (reported as a mangled `%5BRBAC%5D(...)` target). Added leading text so it parses; the URL is unchanged.

## Config (`ci: Ignore medium.com links in link checker`)
Medium serves HTTP 403 to automated clients even though the cited article (scs-0008) is reachable in a browser. Added a `medium.com` `ignorePatterns` entry to `mlc_config.json`.

## Not changed
`scs.community` / `docs.scs.community` links were left as-is: they resolve fine and were verified live; they are not part of the failures.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Standards/scs-0110-v1-ssd-flavors.md
```diff
@@ -260,7 +260,7 @@ not be performed.
 ## Conformance Tests
 
 The list of mandatory flavors that needs to be present should be added to the
-[SCS-Spec.MandatoryFlavors.yaml](https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/SCS-Spec.MandatoryFlavors.yaml)
+[SCS-Spec.MandatoryFlavors.verbose.yaml](https://github.com/SovereignCloudStack/standards/blob/main/Tests/iaas/SCS-Spec.MandatoryFlavors.verbose.yaml)
```
mbuechse, line 263: Actually, this decision record was implemented by a new version of scs-0100 (v3 IIRC).

Some of these documents have not been aging so well 😭 

Just a remark. No need to address this here though.

Standards/scs-0400-v1-status-page-create-decision.md
```diff
@@ -156,4 +156,4 @@ With those requirements in mind the projects that initially were found, were eva
 | project healthy | ❌ last commit 17 months | ❌ last commit 3 years | ❌ last commit 5 months | ✅ last commit 2 months | ✅ recent activities | ✅ recent activities | ❌ archived and abandoned by the owner |
 | documentation | ✅ API ❌ User Documentation | ❌ | ❌ | ❌ | ✅ | ⁇u | ❌ not reachable anymore |
 | git based | ❌ | ✅ | ❌ | ✅ | ❌ | ✅ | ⁇ a netlify based installation is able to communicate with github |
-| project page | [project page](https://cachethq.io/) | [project page](https://github.com/weeblrpress/clearstatus) | [project page](https://www.brotandgames.com/ciao/) | [project page](https://cstate.netlify.app/) | [project page](https://gatus.io/) | [project page](https://github.com/tadhglewis/issue-status) | [project page](https://marquez.co/statusfy) |
+| project page | [project page](https://cachethq.io/) | [project page](https://github.com/weeblrpress/clearstatus) | [project page](https://www.brotandgames.com/ciao/) | [project page](https://github.com/cstate/cstate) | [project page](https://gatus.io/) | [project page](https://github.com/tadhglewis/issue-status) | [project page](https://marquez.co/statusfy) |
```
mbuechse, line 159: This decision record is yet another document that is bound to be outdated at some point. Not sure if we should invest into keeping URLs up to date when multiple years later, the whole landscape may have changed. Alternatively, we could point to the Internet Archive, the way Wikipedia often does.

mbuechse (approved):
LGTM. Thanks a ton! Two remarks that need not be addressed here, but should be discussed.

Standards/scs-0400-v1-status-page-create-decision.md
jklare: I thought about using the internet archive here, but since the link to the repo here should still deliver the same relevant link towards the cstate upstream project, I opted for the replacement.

Standards/scs-0110-v1-ssd-flavors.md
jklare: As far as I understood the context here, the link should lead to the normative list that is also tested. The standard here https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0100-v3-flavor-naming.md defines plenty of examples, but no complete list of mandatory flavors. Since this file here is also the one that is directly consumed via URL by the openstack-flavor-manager I think it is the most faithful replacement.

mbuechse: Your replacement is faithful, and I did not object it!

The statement in the decision record is not factually correct, though.

The standard scs-0100-v3 did at first (in v3.0) contain the lists of mandatory and recommended flavors, until at some point these lists got split off into scs-0103-v1, and, unfortunately, we changed scs-0100-v3 in place, creating v3.1.

Standards/scs-0400-v1-status-page-create-decision.md
mbuechse: The change is fine in isolation.

However, the value of the table is becoming ever more questionable. We may update the URLs, but we don't update the feature matrix, and (naturally) we don't add new projects that may have come about in the meantime. This seems to me quite inconsistent. The table will become wrong either way, but why not let it be wrong in a consistent manner?

Standards/scs-0110-v1-ssd-flavors.md
jklare: Mh, that sounds like something that needs to be fixed by adding a fitting note or editing the content of the file. I would not want to change that here, since this is purely focussed on repairing broken links.

Standards/scs-0400-v1-status-page-create-decision.md
jklare: I think we should still fix broken links, or explicitly remove them, to avoid confusing new contributors (like me) with failing pipelines for things they did not touch.

jklare: And i would also argue the linter needs to be reconfigured to always check all links to avoid this in the future.

mbuechse: Now I'm puzzled. If the link checker does not check all files, how can it confuse a new contributor? On the other hand, if it does check all files, we definitely should use the internet archive or remove the links, because I don't think we have the capacity to maintain these old decision records. The status-page implementation repository has been archived, mind you.

jklare: Which links are checked is based on which files the PR modified: https://github.com/SovereignCloudStack/standards/blob/main/.github/workflows/link-validator.yml#L24 . I would recommend to always check all links and either explicitly disable/remove them once they break, or fix them as early as possible when they break to avoid piling them up and confusing the poor person that fixes a typo in a file with this big pile of broken links in the CI.

mbuechse: Okay. Let's discuss the topic of archiving documents we no longer want to maintain outside this PR.

PR #1260, opened 2026-07-29
Specify what presence of an image means

mbuechse:


mbuechse:
Note: This has to be debated in SIG Std/Cert, particularly the case of `visibility=community`.
I presume @berendt will want this, but I'm not sure this finds consensus.

toothstone:
@mbuechse Could you please elaborate/confirm that this change only aligns the canonical texts with the existing compliance checks. Since I see no diff in `Tests` I assume so, but I think it would be good to have it explicitly spelled out.

berendt:
Visibility of community < public. Fine for me the add this. I don't like the public visibility and prefer community over public.

mbuechse:
@toothstone I can confirm that: https://github.com/SovereignCloudStack/standards/blob/514377b421470602152816269240c6a641587a8f/Tests/iaas/openstack_test.py#L66

However, in my mind, it's more important that the standard makes sense. This is v2, which is still a draft, so we can and should discuss in the SIG what makes most sense.

garloff (approved):
From my perspective, this is good to go.
- Operators can provide images as public (preferrable) or as community images.
- Only those are considered by the standard.
- This means that creating a customer-specific image (private) or one that is shared from one customer to another are neither subject to these rules nor can they be used to fulfill the mandatory list ...

There is a separate discussion on filtering out snapshots.

toothstone (approved):
Aligns the text with the de-facto tests, so I'm very glad to see this happen!

toothstone:
As a follow up @mbuechse , don't we need a similar clarification on [SCS-0102](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0102-v2-image-metadata.md#technical-requirements-and-features)? I just managed to fail the compliance check due to a user-supplied image with "community" visibility, which of course lacks some mandatory metadata. The (non-binding) "motivation" section implies that SCS-0102 is intended to only apply to CSP-provided images, but the tech specs don't explain how those are recognized - the test code does `img.visibility in ('public', 'community')`.

mbuechse:
@toothstone Interesting question. The compliance tests would usually be run on a more-or-less 'pristine' project, so not 'tainted' by such things as user-supplied images. However, we should still answer the question whether it's acceptable that users can so easily interfere with these guarantees -- not necessarily breaking any automation, but who knows?

toothstone:
I think as a first step it would be good to clarify in SCS-0102 that "provided" images are defined as having `img.visibility in ('public', 'community')`, and maybe add some note to the implementation supplement that this is known to cause false negatives for user-provided `community` images.

PR #1261, opened 2026-07-30
fix: avoid t8s HelmRelease name colliding with run_plugin's cluster_id

cwrau:
## Summary
- `run_plugin.py` sets `config['name']` to the cluster_id (e.g. `teuto-1.36`) for every plugin kind, for use by other plugins.
- `PluginT8s` (merged in #1244) reused that same `name` key for the VAP/RBAC-mandated HelmRelease name, so its `"scs-kaas-certification"` default never took effect — the HelmRelease was created as `teuto-1.36`/`teuto-1.35` instead, which the ValidatingAdmissionPolicy on the management cluster rejects, breaking CI.
- Renamed the config key to `hrName` to avoid the collision. `teuto-1.35`/`teuto-1.36` entries in `clusters.yaml.j2` don't set it, so they correctly fall back to `scs-kaas-certification`.

## Test plan
- [x] Added `test_hr_name_ignores_run_plugin_name`, simulating `run_plugin.py`'s `setdefault('name', cluster_id)` behavior, to prevent regression.
- [x] `python -m pytest Tests/kaas/plugin` — 29/29 passed.
- [x] `flake8 --config=.flake8` — clean except pre-existing unrelated warnings.

### DIFF
--- a/Tests/kaas/plugin/plugin_t8s.py
+++ b/Tests/kaas/plugin/plugin_t8s.py
@@ -29,7 +29,11 @@ class PluginT8s(KubernetesClusterPlugin):
     via a Flux HelmRelease.
 
     Creates a HelmRelease on mgmt-bfe2-prod and waits for the resulting cluster's
-    kubeconfig to appear in the secret '<name>-kubeconfig'.
+    kubeconfig to appear in the secret '<hrName>-kubeconfig'.
+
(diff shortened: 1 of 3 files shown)


mbuechse (approved):
Looking great. Thanks a ton for the quick action!

PR #1262, opened 2026-08-05
Include actual requirement into scs-0128-v1

mbuechse:
(the way it's been handled heretofore)


mbuechse:
has been handled with #1247

PR #1265, opened 2026-08-21
New standard for node software recency.

garloff:
This was discussed in the SIG Std/Cert.
One key motivation behind the k8s version policy is enabling users to keep their k8s clusters secure by having the latest k8s patch versions available without undue delay.
We need more: The other software running on the nodes must include the latest security fixes as well. Require this in a separate standard scs-0220-v1 (Draft).

### DIFF
--- a/Standards/scs-0220-v1-node-image-recency.md
+++ b/Standards/scs-0220-v1-node-image-recency.md
@@ -8,7 +8,7 @@ enhances: scs-0210-v2-k8s-version-policy.md
 
 ## Introduction
 
-In [scs-0210](scs-0210-v2-k8s-version-policy),
+In [scs-0210](https://docs.scs.community/standards/scs-0210-v2-k8s-version-policy),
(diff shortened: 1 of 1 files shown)


Standards/scs-0220-v1-node-image-recency.md
```diff
@@ -0,0 +1,84 @@
...
+and avoiding including packages that are unmaintained. In general, it
+is a good idea to avoid the inclusion of unnecessary software and provide
+minimal and hardened node images.
+
+If node images are registered as public or community images on an
+SCS-compatible IaaS, the `build_date` property indicates that (at least)
+all (security) fixes up to that date are included, see
+[scs-0102](https://docs.scs.community/standards/scs-0102-v2-image-metadata).
+
+Note that the qualifier "relevant" was used when requiring all released
```
garloff, line 69: This could be moved into the implementation and test notes.

In there, I would also add some thoughts on testing and how building images that use the latest maintenance updates do trivially fulfill this requirement.

PR #1268, opened 2026-08-27
Add ScaleUp OCC3 Region

fzakfeld:
Second region. SCS-compatible IaaS v5.1 passes locally on my machine


playbooks/clouds.yaml.j2
```diff
@@ -100,6 +100,15 @@ clouds:
     region_name: "RegionOne"
     interface: "public"
     identity_api_version: 3
+  scaleup-occ3:
+    auth_type: v3applicationcredential
+    auth:
+      auth_url: https://keystone.occ3.scaleup.cloud
+      application_credential_id: "{{ clouds_conf.scaleup_occ3_ac_id }}"
+      application_credential_secret: "{{ clouds_conf.scaleup_occ3_ac_secret }}"
+    region_name: "RegionOne"
```
depressiveRobot, line 109: The region name is the same as for scaleup-occ2. Should it be "RegionTwo" instead?

Besides that, I am curious why scaleup-occ2 is named "RegionOne". The number 2 suggests "RegionTwo" and therefore scaleup-occ3 suggests "RegionThree". Why do the regions start with 2 instead of 1?

fzakfeld: The Keystone region is just internal to Keystone (defaults to RegionOne in most deployments). Within our design (like many others do), we don't run a single Keystone deployment across our cloud environments, thus we only ever have one Keystone region per environments, named the same.

We historically had other cloud environments (OCC1, even older ones...) without SCS compatability / EoL

fzakfeld:
I'd be fine waiting until the official certification, as the local tests are running fine (v5.1 for now)

playbooks/clouds.yaml.j2
depressiveRobot: Thanks for your explanation.

However, I still think the region names must be distinct as we use this also as some kind of identifier. @mbuechse probably knows more on this.

fzakfeld: If this is the case, then it must be changed. There is nothing to guarantee that region names are unique across environments / providers

mbuechse: No, I don't think we're using this `region_name` anywhere. I'm fine with using `RegionOne` everywhere (although I would prefer `dummy` or something to make clear that it's not really being used in a meaningful way).

depressiveRobot: Sorry, I mixed up something. `RegionOne` is totally fine. Actually, leave it that way to find the correct endpoints in the Keystone service catalog.

fzakfeld: @mbuechse to be clear, the region_name does serve the purpose of referencing the (single) region within Keystone, altering it would not work

mbuechse: I realize that. I was not being entirely serious. However, I think the default value in Keystone should not be RegionOne if most deployments only ever have one region in Keystone. My proposal was not directed at you specifically.

PR #1273, opened 2026-09-07
ScaleUp KaaS: check 1.36, drop checks for 1.33

fzakfeld:


playbooks/clusters.yaml.j2
```diff
@@ -83,11 +83,11 @@ clusters:
         garden_api_server: "{{ clouds_conf.noris_api_server }}"
         garden_ca_data: "{{ clouds_conf.noris_ca_data }}"
         garden_token: "{{ clouds_conf.noris_token }}"
-  scaleup-1.33:
+  scaleup-1.34:
```
mbuechse: I would've preferred if this was changed to 1.36, so 1.34 and 1.35 could stay the same.

Tests/config.toml
```diff
@@ -41,8 +41,8 @@ scopes = [
 subjects = [
     "noris-1.34",
     "noris-1.35",
-    "scaleup-1.34",
```
mbuechse, line 44: Why is 1.34 removed?

Tests/config.toml
```diff
@@ -67,7 +68,6 @@ scopes = [
 ]
 subjects = [
     "noris-1.34",
-    "scaleup-1.34",
```
mbuechse, line 70: ditto

playbooks/clusters.yaml.j2
fzakfeld: This was just to keep the order, github displays the changes like this because nothing else other than versions changed

Tests/config.toml
fzakfeld: Should we test all 3? I'd be fine with it, thought the rule of thumb was to test the latest 2 versions

playbooks/clusters.yaml.j2
mbuechse: think of it as slots:

- `minor % 3 == 0`
- `minor % 3 == 1`
- `minor % 3 == 2`

Tests/config.toml
mbuechse: We test 2 if 2 are active, and we test 3 if 3 are active.

mbuechse (approved):

PR #1274, opened 2026-09-07
Add subject syself-1.36 to KaaS checks

mbuechse:
replacing outdated syself-1.33. Also: add missing scaleup-1.36 to bootstrap.yaml


janiskemper (approved):
looks reasonable. To me it looks though as if you do both the same things for Syself and ScaleUp. You can maybe update the title ;)

mbuechse:
@janiskemper it says

> Also: add missing scaleup-1.36 to bootstrap.yaml

The other parts are only for syself-1.36

PR #1275, opened 2026-09-08
Introduce k8s_test.py in analogy to openstack_test.py

mbuechse:
Handle kaas test the same way we handle iaas tests: with preflight checks.


Tests/kaas/k8s_test.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def run_sono(config, testcase, *args):
+    return SonobuoyHandler(
+        SCS_SONOBUOY_CONFIG_PATH, testcase, config.kubeconfig_path,
+        args=config.compute_sono_args(*args),
+    ).run()
+
+
+TESTCASES = {
```
depressiveRobot: Let's move this to the top as is it a global variable and to be more in line with `openstack_test.py` where the test case initialization is also at the beginning.

Tests/kaas/k8s_test.py
```diff
@@ -0,0 +1,159 @@
...
+import os
+import os.path
+import sys
+
+
+from scs_0210_version_policy.k8s_version_policy import version_policy_check
+from sonobuoy_handler.sonobuoy_handler import SonobuoyHandler
+
+
+HERE = os.path.dirname(__file__)
```
depressiveRobot: Let's use a more understandable name here like `FILE_PATH` or `FILE_DIR`.

depressiveRobot (commented):
This is a nice improvement 💯 Thanks for doing this.

Tests/kaas/k8s_test.py
mbuechse: Yup, I agree.

Tests/kaas/k8s_test.py
mbuechse: Hm. We do use HERE already in other files. (Maybe they should be changed then.) Not sure however if this is kind of a common thing in the Python community. Will do a quick round of research.

mbuechse: `grep` results:

```text
./iaas/scs_0100_flavor_naming/flavor_names.py:HERE = Path(__file__).parent
./kaas/scs_0214_node_distribution_RETIRED/check_nodes_test.py:HERE = Path(__file__).parent
./kaas/scs_0214_node_distribution_RETIRED/check_nodes_test.py:    with open(Path(HERE, "testdata", "scenarios.yaml")) as stream:
./kaas/k8s_test.py:HERE = os.path.dirname(__file__)
./kaas/k8s_test.py:SCS_SONOBUOY_CONFIG_PATH = os.path.join(HERE, 'scs-sonobuoy-config-v1.yaml')
./kaas/scs_0210_version_policy/k8s_version_policy_test.py:HERE = Path(__file__).parent
./kaas/scs_0210_version_policy/k8s_version_policy_test.py:    with open(Path(HERE, "testdata", "releases.json")) as stream:
./kaas/scs_0210_version_policy/k8s_version_policy.py:HERE = Path(__file__).parent
./kaas/scs_0210_version_policy/k8s_version_policy.py:EOLDATA_PATH = Path(HERE, "k8s-eol-data.yml")
```

Tests/kaas/k8s_test.py
mbuechse: ad4a991

Tests/kaas/k8s_test.py
mbuechse: 71ac525

mbuechse: I decided to rename in spite of the other occurrences, not least because the others are different in the sense that they use pathlib, which I didn't want to do because it introduces further concepts for very little gain.

depressiveRobot (approved):
Thanks for incorporating the suggestions. Looks fine, let's merge :)

PR #1276, opened 2026-09-10
Add testcases for scs-0104-v2

### LINKED ISSUE #933: Extend standard images standard to move away from fixed names
The currently stable version [scs-0104-v1](https://docs.scs.community/standards/scs-0104-v1-standard-images/) enforces a certain name (scheme) for each mandatory and recommended image, along with a certain source URL.

This fixed name is

- undesireable, except maybe for the special case of the capi image, because CSPs already have their own naming schemes in place, and duplicating images just for the sake of getting the name right confuses customers.
- unnecessary, because Glance (the OpenStack service responsible for image managing) has viable filtering capabilities (although this needs to be confirmed).

Therefore, in a new v2 or an updated v1 of the standard, we want to be able to express the following scenario:

- an image with
  - either os_distro=X, os_version=Y, and os_original=true (used for stock OS images such as Ubuntu 24.04 or Debian 12)
  - or name matches regex R (used for special case of capi)
- MAY, SHOULD or MUST be present (corresponding to status=optional, recommended, or mandatory, respectively),
- and if present, it MUST have image_source starting with one of XYZ.

From my POV, we can justify updating v1 for the following reason:

- the added capability will not be activated by existing yaml files, so their meaning will not change,
- changing the yaml file in a way that breaks compliance is not possible for a stable version of a certificate scope.

In other words, whatever test subject complies with current v1 will also comply with the updated v1.

I think we may even extend the yaml file, because no CSP is using `os_original` yet, so no breaking of compliance is to be expected.

## Definition of Done:

- [x] Confirm that selecting images using the os_X properties is readily possible with Glance.
- [x] Write a ~~blogpost~~ docs-page article explaining how this image selection works for end users.
- [x] Extend the standard.
- [x] Also extend standard to make sure future versions of operating systems are already captured (so long as they follow the current schema, i.e., Ubuntu 26.04, Ubuntu 28.04, Debian 13, Debian 14, etc.).
- [x] Adapt the check script.
- [x] Change relevant YAML file so names are no longer enforced for stock OS images.

mbuechse:
well, at least the required testcases


mbuechse:
compare this to https://github.com/SovereignCloudStack/standards/issues/933#issuecomment-3807622323
(so far, I only implemented the required ones, not the recommended ones)

depressiveRobot (approved):
Thank you for writing the tests. Tested it locally against `scs2`. Everything else looks fine too. Let's merge :)

PR #1279, opened 2026-09-17
Fix missing documentation for os_distro

### LINKED ISSUE #1278: [BUG] SCS-0102-V2 os_purpose uniqueness not possible with Almalinux images
SCS-0102-V2 requires uniqueness for any combination of os_distro, os_version, architecture.

For `os_distro`, allowed values of https://docs.openstack.org/glance/2025.1/admin/useful-image-properties.html must be followed.

This makes it impossible to offer Almalinux and Centos Stream images of the same version at the same time, as both are considered `centos`.

Possible fixes include:

- allow a non-standard `os_distro=almalinux`
- allow a max. of 2 images for any combination with `os_distro=centos`

mbuechse:
Fixes #1278


mbuechse:
@fzakfeld The whole thing turned out to be more complex than I thought. The OpenStack docs do refer to "the vocabulary" of libosinfo, but I'm not sure what is meant. OpenStack does omit any `linux` suffix. So my suggestion would be to use `alma` instead of `almalinux`.

